PR #9, opened 2018-05-07
fix: nodejs 6 async_hook issue

fabiogomessilva:
Fix for the issue found on PR/6

### DIFF
--- a/packages/opencensus-core/src/trace/model/tracer.ts
+++ b/packages/opencensus-core/src/trace/model/tracer.ts
@@ -169,6 +169,7 @@ export class Tracer implements types.Tracer {
     }
   }
 
+  // TODO: reveiw clearCurrentTrace when using continuation-local-storage
   /** Clears the current root span. */
   clearCurrentTrace() {
     this.currentRootSpan = null;
--- a/packages/opencensus-core/src/trace/model/tracer.ts
+++ b/packages/opencensus-core/src/trace/model/tracer.ts
@@ -52,7 +52,6 @@ export class Tracer implements types.Tracer {
   constructor() {
     this.activeLocal = false;
     this.contextManager = cls.createNamespace();
-    this.clearCurrentTrace();
   }
 
   /** Gets the current root span. */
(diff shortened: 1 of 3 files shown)


packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -52,7 +52,6 @@ export class Tracer implements types.Tracer {
   constructor() {
     this.activeLocal = false;
     this.contextManager = cls.createNamespace();
-    this.clearCurrentTrace();
```
kjin: What is the purpose of this deletion?

packages/opencensus-core/test/test-tracer.ts
```diff
@@ -52,17 +52,13 @@ describe('Tracer', () => {
 
   /** Should get/set the current RootSpan from tracer instance */
   describe('get/set currentRootSpan()', () => {
-    let tracer, rootSpan;
-    before(() => {
-      tracer = new Tracer();
-      rootSpan = new RootSpan(tracer);
-      tracer.currentRootSpan = rootSpan;
-    });
+    const tracer = new Tracer().start(defaultConfig);
```
kjin: This change looks unrelated to the Node 6 change. Was this intended?

packages/opencensus-core/src/trace/model/tracer.ts
fabiogomessilva: Some unit tests were failing. Both cases related to context issues.

packages/opencensus-core/test/test-tracer.ts
fabiogomessilva: Some unit tests were failing. Related to context issues.

packages/opencensus-core/src/trace/model/tracer.ts
kjin: `clearCurrentTrace` is part of the public API but it will throw when using `continuation-local-storage`. Could you add a TODO to address this in a follow-up PR?

kjin:
Current CircleCI failure is unrelated (Circle seems to default to Node 4 for some reason.)

PR #12, opened 2018-05-11
Opencensus propagation stackdriver

ofrobots:
Introduces a package for Stackdriver propagation. Right now only v1 format is implemented.

TODO:
- [x] Add tests for generate.

### DIFF
--- a/packages/opencensus-propagation-stackdriver/src/index.ts
+++ b/packages/opencensus-propagation-stackdriver/src/index.ts
@@ -28,24 +28,31 @@ export interface SpanContext {
 /**
  * An transport and environment neutral API for getting request headers.
  */
-export interface Getter {
+export interface HeaderGetter {
   getHeader(name: string): string|string[]|undefined;
 }
 
 /**
  * A transport and environment neutral API for setting headers.
  */
-export interface Setter {
+export interface HeaderSetter {
   setHeader(name: string, value: string): void;
 }
 
+
 export interface Propagation {
-  extract(getter: Getter): SpanContext|null;
-  inject(setter: Setter, spanContext: SpanContext): void;
+  extract(getter: HeaderGetter): SpanContext|null;
+  inject(setter: HeaderSetter, spanContext: SpanContext): void;
   generate(): SpanContext;
 }
 
-export const v1: Propagation = v1API;
+export const v1: Propagation = {
+  extract: v1API.extract,
+  inject: v1API.inject,
+  generate: v1API.generate
+};
 
 // Also export the v1 API as the default API.
-export * from './v1';
+export const extract = v1.extract;
+export const inject = v1.inject;
+export const generate = v1.generate;
--- a/packages/opencensus-propagation-stackdriver/src/index.ts
+++ b/packages/opencensus-propagation-stackdriver/src/index.ts
@@ -1,0 +1,21 @@
+
+import {IncomingMessage, ServerResponse} from 'http';
+import * as v1API from './v1';
+
+export interface SpanContext {
+  /** Trace ID */
+  traceId: string;
+  /** Span ID */
+  spanId: string;
+  /** Options */
+  options?: number;
+};
+
+export interface API {
+  extract(req: IncomingMessage): SpanContext|null;
+  inject(res: ServerResponse, spanContext: SpanContext): void;
+  generateNewSpanContext(): SpanContext;
+};
+
+export const v1: API = v1API;
+
--- a/packages/opencensus-propagation-stackdriver/src/index.ts
+++ b/packages/opencensus-propagation-stackdriver/src/index.ts
@@ -1,5 +1,19 @@
+/**
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
 
-import {IncomingMessage, ServerResponse} from 'http';
 import * as v1API from './v1';
 
 export interface SpanContext {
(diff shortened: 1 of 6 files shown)


packages/opencensus-propagation-stackdriver/src/index.ts
```diff
@@ -0,0 +1,51 @@
...
+  /** Span ID */
+  spanId: string;
+  /** Options */
+  options?: number;
+}
+
+/**
+ * An transport and environment neutral API for getting request headers.
+ */
+export interface Getter {
```
kjin: Maybe we should prefer names that have to do more with headers or propagation. `Getter` seems too generic of an interface name.

packages/opencensus-propagation-stackdriver/src/v1.ts
```diff
@@ -0,0 +1,84 @@
+/**
+ * Copyright 2015 Google Inc. All Rights Reserved.
```
kjin: Does it make sense to update the copyright?

packages/opencensus-propagation-stackdriver/test/test-v1.ts
```diff
@@ -0,0 +1,119 @@
...
+describe('extract', () => {
+  describe('valid inputs', () => {
+    it('should return expected values: cafef00d/667;o=1', () => {
+      const getter = {
+        getHeader() {
+          return 'cafef00d/667;o=1';
+        }
+      };
+      const result = notNull(extract(getter));
+      assert.strictEqual(result.traceId, 'cafef00d');
```
kjin: nice

packages/opencensus-propagation-stackdriver/test/test-v1.ts
```diff
@@ -0,0 +1,119 @@
...
+      input: {traceId: '123456', spanId: '667', options: undefined},
+      output: '123456/667'
+    },
+    {input: {traceId: '123456', spanId: '667'}, output: '123456/667'}
+  ];
+
+  testData.forEach(({input, output}) => {
+    it(`returns well-formatted trace context for ${inspect(input)}`, () => {
+      const setter = {
+        setHeader(name: string, value: string) {
```
kjin: Maybe assert that this function will be called?

packages/opencensus-propagation-stackdriver/src/v1.ts
ofrobots: I don't think so. This is derived from the source file in the Stackdriver trace agent.

packages/opencensus-propagation-stackdriver/test/test-v1.ts
ofrobots: Changed the tests substantially. This is no longer an issue.

packages/opencensus-propagation-stackdriver/test/test-v1.ts
```diff
@@ -108,12 +91,47 @@ describe('inject', () => {
...
-      };
-      inject(setter, input);
+      const header = serializeSpanContext(input);
+      assert.deepEqual(header, output);
     });
   });
 });
+
+describe('extract', () => {
+  it('should return prased header from getter.getHeader', () => {
```
kjin: `parsed`

packages/opencensus-propagation-stackdriver/test/test-v1.ts
```diff
@@ -108,12 +91,47 @@ describe('inject', () => {
...
+        done();
+      }
+    };
+    inject(setter, spanContext);
+  });
+});
+
+describe('generate', () => {
+  it('should generate a valid span context', () => {
+    const TIMES = 20;
```
kjin: Should we be checking uniqueness as well?

packages/opencensus-propagation-stackdriver/src/index.ts
fabiogomessilva: There is a SpanContext already defined in Core package. Do you any restriction to use it here?

packages/opencensus-propagation-stackdriver/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as v1API from './v1';
+
+export interface SpanContext {
+  /** Trace ID */
```
fabiogomessilva: There is a SpanContext already defined in Core package. Do you have any restriction to use it here?

packages/opencensus-propagation-stackdriver/src/index.ts
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * A transport and environment neutral API for setting headers.
+ */
+export interface HeaderSetter {
+  setHeader(name: string, value: string): void;
+}
+
+
+export interface Propagation {
+  extract(getter: HeaderGetter): SpanContext|null;
```
fabiogomessilva: We were planning to add a Propagation interface to Core package in the next few weeks. What do you think about to add this interface to Core package? It should be at :
```
opencensus-core/src/trace/propagation/types.ts
```
or maybe in a new PR.

ofrobots: Here's the list of dependencies that `core` has:

```
  "dependencies": {
    "continuation-local-storage": "^3.2.1",
    "debug": "^3.1.0",
    "intercept-stdout": "^0.1.2",
    "log-driver": "^1.2.7",
    "semver": "^5.5.0",
    "shimmer": "^1.2.0",
    "uuid": "^3.2.1"
  }
```

I think the `propagation` modules should have close to zero dependencies. In my specific use case, I want to use it in a logging context, and I don't specifically do not want to bring in `continuation-local-storage` or other similar functionality. The only things I need from `core` are two abstract types.

I think this discussion is suggesting to me that `core` is perhaps not "core-enough". /cc @kjin 

> in the next few weeks. 

I need this propagation module right now. If there is refactoring possible – I'm +1 but let's remove the code from when you're ready in a few weeks. For now this type can live here.

kjin: > I think this discussion is suggesting to me that `core` is perhaps not "core-enough".
This makes sense to me. We should discuss the possibility of an interfaces-only module.

fabiogomessilva: > ..  If there is refactoring possible – I'm +1 but let's remove the code from when you're ready in a few weeks. For now this type can live here.

Ok

Do you prefer to open yourself the PR to include those definitions to ``opencensus-core/src/trace/propagation/types.ts`` or do you mind if we do so? Of course, we'll gonna keep you as the author of the commit.

packages/opencensus-propagation-stackdriver/test/test-v1.ts
ofrobots: Done.

PR #14, opened 2018-05-16
feat: add packages intrumentation-http/https and propagation-b3

fabiogomessilva:
This PR adds packages intrumentation-http/ intstrumentation-https and propagation-b3 to opencensus-node


packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+        if (!plugin.tracer.currentRootSpan) {
+          plugin.logger.debug('outgoingRequest starting a root span');
+          return plugin.tracer.startRootSpan(
+              traceOptions,
+              plugin.makeRequestTrace(request, arguments, plugin));
+        } else {
+          plugin.logger.debug('outgoingRequest starting a child span');
+          const span: types.Span = plugin.tracer.startChildSpan(
+              traceOptions.name, traceOptions.type);
+          return (span: types.Span) =>
```
kjin: The return type here is incorrect. This function should return `ClientRequest` but this is of type `(span: types.Span) => ClientRequest`.

kjin: This is causing the `http` tests to fail, but not `https`. I believe the reason for this is because `http.get` and `https.get` are implemented differently; the way you are patching `https.get` in the `https` plugin, it is very likely that `req.end` is called _before_ this line, in `original.apply(this, arguments)` (line 170), because in Node 6 its implementation might simply call through the `http` implementation.

fabiogomessilva: Thanks @kjin.

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+   * @param spanContext
+   */
+  inject(setter: types.HeaderSetter, spanContext: types.SpanContext): void {
+    if (setter) {
+      setter.setHeader(
+          X_B3_TRACE_ID, spanContext && spanContext.traceId || 'undefined');
+      setter.setHeader(
+          X_B3_SPAN_ID, spanContext && spanContext.spanId || 'undefined');
+      if (spanContext && (spanContext.options & SAMPLED_VALUE) !== 0) {
+        setter.setHeader(X_B3_SAMPLED, SAMPLED_VALUE.toString());
```
kjin: Nit: `${SAMPLED_VALUE}` -- or just make `SAMPLED_VALUE` itself a string.

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+export class B3Format implements types.Propagation {
+  /**
+   * Gets the trace context from a request headers. If there is no trace context
+   * in the headers, null is returned.
+   * @param getter
+   */
+
+  extract(getter: types.HeaderGetter): types.SpanContext {
+    // tslint:disable-next-line
+    if (getter) {
```
kjin: `getter` should be guaranteed to exist because of type checking. Same with `setter`.

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,69 @@
+/**
```
ofrobots: style nit: can we use snake-cake filename format across the repo?

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Copyright 2018 Google Inc. All Rights Reserved.
```
ofrobots: Should be `Google LLC.`.

packages/opencensus-propagation-b3/AUTHORS
```diff
@@ -0,0 +1 @@
+Google Inc.
```
ofrobots: Should be `Google LLC.`.  Should CESAR be included here?

packages/opencensus-propagation-b3/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@opencensus/opencensus-propagation-b3",
+  "version": "0.0.1",
+  "description": "OpenCensus is a toolkit for collecting application performance and behavior data.",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.js",
+  "repository": "census-instrumentation/opencensus-node",
+  "scripts": {
+    "build": "node_modules/.bin/tsc --declaration",
+    "test": "npm run script run-unit-tests",
```
ofrobots: Q: is there a reason we are not using `gts check` to enforce the style guide at this point?

packages/opencensus-propagation-b3/tsconfig.full.json
```diff
@@ -0,0 +1,17 @@
+{
```
ofrobots: Do both these tsconfig files need to exist?

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,69 @@
...
+const SAMPLED_VALUE = '1';
+
+/** Propagates span context through B3 Format propagation. */
+export class B3Format {
+  /**
+   * Gets the trace context from a request headers. If there is no trace context
+   * in the headers, null is returned.
+   * @param headers
+   */
+  static extractFromHeader(headers: object): TraceContext {
```
ofrobots: The API is diverging from https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-propagation-stackdriver/src/index.ts#L43. I take it that it would be reconciled in a follow-on?

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,69 @@
...
+    if (headers && (headers[X_B3_TRACE_ID] || headers[X_B3_SAMPLED])) {
+      const traceContext = {
+        traceId: headers[X_B3_TRACE_ID],
```
ofrobots: This may not be the right thing to do if `headers[X_B3_TRACE_ID]` returns a `string[]`. This applies to the other headers below as well.

packages/opencensus-propagation-b3/src/b3Format.ts
ofrobots: Whoops, I see this has already been addressed. I was looking at a subset of the commits.

packages/opencensus-propagation-b3/AUTHORS
ofrobots: Whoops, I see this has already been addressed. I was looking at a subset of the commits.

packages/opencensus-propagation-b3/src/b3Format.ts
ofrobots: Whoops, I see this has already been addressed. I was looking at a subset of the commits.

packages/opencensus-instrumentation-https/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "@types/node": "^9.4.7",
+    "@types/semver": "^5.5.0",
+    "@types/shimmer": "^1.0.1",
+    "@types/uuid": "^3.4.3",
+    "gts": "^0.5.1",
+    "mocha": "^5.0.4",
+    "ncp": "^2.0.0",
+    "nock": "^9.2.6",
+    "nyc": "^11.7.1",
+    "ts-node": "^4.0.0",
```
ofrobots: missing `rimraf` here?

packages/opencensus-instrumentation-https/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "typescript": "^2.7.2"
```
ofrobots: Use `~` as `typescript` doesn't follow semver. A minor change will cause build breaks.

packages/opencensus-instrumentation-https/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "keywords": [
+    "opencensus",
+    "nodejs",
+    "tracing",
+    "profiling"
+  ],
+  "author": "Google Inc.",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=6.00"
```
ofrobots: nit `>=6.0.0`, or just `>=6`.

packages/opencensus-propagation-b3/package.json
```diff
@@ -0,0 +1,55 @@
...
+  },
+  "devDependencies": {
+    "@types/mocha": "^2.2.48",
+    "@types/node": "^9.4.7",
+    "@types/uuid": "^3.4.3",
+    "gts": "^0.5.1",
+    "mocha": "^5.0.4",
+    "ncp": "^2.0.0",
+    "nyc": "^11.7.1",
+    "ts-node": "^4.0.0",
```
ofrobots: `rimraf` missing?

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Gets the trace context from a request headers. If there is no trace context
+   * in the headers, null is returned.
+   * @param getter
+   */
+
+  extract(getter: types.HeaderGetter): types.SpanContext {
+    // tslint:disable-next-line
+    if (getter) {
+      const opt = getter.getHeader(X_B3_SAMPLED);
```
ofrobots: Question: are this logic valid if any of the `getHeader` operations returns a `string[]`.

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+      const spanContext = {
+        traceId: getter.getHeader(X_B3_TRACE_ID),
+        spanId: getter.getHeader(X_B3_SPAN_ID),
+        options: isNaN(Number(opt)) ? undefined : Number(opt)
+      } as types.SpanContext;
```
ofrobots: Why is this cast needed? Is is due to the fact that we are potentially assigning `string[]` to `traceId` etc.?

packages/opencensus-instrumentation-http/package.json
```diff
@@ -0,0 +1,64 @@
+{
+  "name": "@opencensus/instrumentation-http",
+  "version": "0.0.1",
+  "description": "OpenCensus is a toolkit for collecting application performance and behavior data.",
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/package.json
```diff
@@ -0,0 +1,64 @@
...
+    "mocha": "^5.0.4",
+    "ncp": "^2.0.0",
+    "nock": "^9.2.6",
+    "nyc": "^11.7.1",
+    "ts-node": "^4.0.0",
+    "typescript": "^2.7.2"
+  },
+  "dependencies": {
+    "@opencensus/opencensus-core": "^0.0.1",
+    "@opencensus/propagation-b3": "^0.0.1",
```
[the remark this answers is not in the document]
fabiogomessilva: It was before the refactoring, not anymore. Removed!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+import {types} from '@opencensus/opencensus-core';
+import {classes} from '@opencensus/opencensus-core';
+import {logger} from '@opencensus/opencensus-core';
+import * as semver from 'semver';
+import * as shimmer from 'shimmer';
+import * as url from 'url';
+import * as uuid from 'uuid';
+
+// import * as Debug from 'debug';
+// const debug = Debug('opencensus');
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  /**
+   * Patches HTTP incoming and outcoming request functions.
+   * @param moduleExporters The HTTP package.
+   * @param tracer A tracer instance to create spans on.
+   * @param version The package version.
+   */
+  // tslint:disable:no-any
+  applyPatch(moduleExporters: any, tracer: types.Tracer, version: string) {
```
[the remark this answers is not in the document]
fabiogomessilva: > moduleExports rather than moduleExporters. 
Indeed I already know about this typo. But is a typo from BasePlugin from Core that has a property named moduleExporters, so we should keep this name until correct the typo from BasePlugin in a specific PR - or break this rule and rename the BasePlugin from Core right now. What do you think? I will add a TODO even so. 
  
> Also, the type here is typeof http (after import * as http from 'http'), is that correct?
Yes, it should be, but we were not importing http. Maybe we should. I will analyze that.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+    super(moduleName);
+  }
+
+  /**
+   * Patches HTTP incoming and outcoming request functions.
+   * @param moduleExporters The HTTP package.
+   * @param tracer A tracer instance to create spans on.
+   * @param version The package version.
+   */
+  // tslint:disable:no-any
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+
+    shimmer.wrap(moduleExporters, 'request', this.patchOutgoingRequest());
+
+    // In Node 8, http.get calls a private request method, therefore we patch it
+    // here too.
+    if (semver.satisfies(version, '>=8.0.0')) {
+      shimmer.wrap(moduleExporters, 'get', this.patchOutgoingRequest());
+    }
+
+    shimmer.wrap(
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+          type: 'CLIENT',
+        };
+
+
+        // Checks if this outgoing request is part of an operation by checking
+        // if there is a current root span, if so, we create a child span. In
+        // case there is no root span, this means that the outgoing request is
+        // the first operation, therefore we create a root span.
+        if (!plugin.tracer.currentRootSpan) {
+          plugin.logger.debug('outgoingRequest starting a root span');
```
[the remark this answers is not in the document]
fabiogomessilva: It does make sense when the node app is working as a CLIENT. For example, running asynchronous jobs - not directly related to a method response.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+          return plugin.tracer.startRootSpan(
+              traceOptions,
+              plugin.makeRequestTrace(request, arguments, plugin));
+        } else {
+          plugin.logger.debug('outgoingRequest starting a child span');
+          const span: types.Span = plugin.tracer.startChildSpan(
```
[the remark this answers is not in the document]
fabiogomessilva: Removed.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+  }
+
+  /**
+   * Injects span's context to header for distributed tracing and finshes the
+   * span when the response is finished.
+   * @param original The original patched function.
+   * @param args The arguments to the original function.
+   */
+  // tslint:disable:no-any
+  makeRequestTrace(request: any, args: any, plugin: HttpPlugin): any {
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+        propagation.inject(setter, span.spanContext);
+      }
+
+      if (!span) {
+        plugin.logger.debug('makeRequestTrace span is null');
+        return request;
+      }
+
+      // tslint:disable:no-any
+      const req = request.on('response', (response: any) => {
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+        plugin.tracer.wrapEmitter(response);
+        plugin.logger.debug('outgoingRequest on response()');
+
+        response.on('end', () => {
+          plugin.logger.debug('outgoingRequest on end()');
+          span.addAttribute('http.host', args[0].host);
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+        'emit');
+  }
+
+  /**
+   * Creates spans for incoming requests, restoring spans' context if applied.
+   * @param plugin The HTTP plugin itself.
+   */
+  patchIncomingRequest() {
+    const plugin: HttpPlugin = this;
+    return (original: Function): Function => {
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+        this.moduleExporters && this.moduleExporters.Server &&
+            this.moduleExporters.Server.prototype,
+        'emit');
+  }
+
+  /**
+   * Creates spans for incoming requests, restoring spans' context if applied.
+   * @param plugin The HTTP plugin itself.
+   */
+  patchIncomingRequest() {
```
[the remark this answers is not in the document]
fabiogomessilva: Before defining the func (original) required by shimmer we are setting  ``const plugin = this``;

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+    const plugin: HttpPlugin = this;
+    return (original: Function): Function => {
+      // tslint:disable:no-any
+      return function incomingRequest(
+          event: string, request: any, response: any) {
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+        // Only traces request events
+        if (event !== 'request') {
+          return original.apply(this, arguments);
+        }
+
+        plugin.logger.debug('%s plugin incomingRequest', plugin.moduleName);
+        const propagation = plugin.tracer.propagation;
+        const headers = arguments[1].headers;
+        const getter = {
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+import * as http from 'http';
+import * as mocha from 'mocha';
+import * as nock from 'nock';
+import * as shimmer from 'shimmer';
+
+import {plugin} from '../src/';
+import {HttpPlugin} from '../src/';
+
+
+function doNoke(
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+    url: string, path: string, httpCode: number, respBody: string,
+    times?: number) {
+  let i = times;
```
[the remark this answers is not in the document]
fabiogomessilva: Done !

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+  if (!times) {
+    i = 1;
+  }
+  nock(url).get(path).times(i).reply(httpCode, respBody);
+}
+
+
+const httpRequest = {
+  get: (options: {}|string) => {
+    return (new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
fabiogomessilva: Done !

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+   * @param plugin The HTTP plugin itself.
+   */
+  patchOutgoingRequest() {
+    const plugin: HttpPlugin = this;
+    // tslint:disable:no-any
+    return (original: any): any => {
+      return function outgoingRequest() {
+        // Makes sure the url is an url object
+        if (typeof (arguments[0]) === 'string') {
+          arguments[0] = url.parse(arguments[0]);
```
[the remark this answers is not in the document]
fabiogomessilva: Refactor done!

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -0,0 +1,303 @@
...
+          response.end = function(this: any) {
+            response.end = originalEnd;
+            const returned = response.end.apply(this, arguments);
+
+            rootSpan.addAttribute('http.host', request.url.host);
+            rootSpan.addAttribute('http.method', request.method);
+            rootSpan.addAttribute('http.path', request.url.pathname);
+            rootSpan.addAttribute('http.route', request.url.path);
+            rootSpan.addAttribute(
+                'http.user_agent', 'HTTPClient/' + request.httpVersion);
```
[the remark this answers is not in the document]
fabiogomessilva: Refactor Done!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+
+  onEndSpan(root: types.RootSpan) {
+    this.endedRootSpans.push(root);
+  }
+}
+
+const server = http.createServer((request, response) => {
+                     response.end('Test Server Response');
+                   })
+                   .listen(3000);
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+
+
+// mocha this.timeout() fails when using ES6's arrow functions
+// (Issue link: https://github.com/mochajs/mocha/issues/2018)
+// tslint:disable-next-line
+describe('HttpPlugin', function() {
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+
+class RootSpanVerifier implements types.OnEndSpanEventListener {
+  endedRootSpans: types.RootSpan[] = [];
+
+  onEndSpan(root: types.RootSpan) {
+    this.endedRootSpans.push(root);
+  }
+}
+
+const server = http.createServer((request, response) => {
```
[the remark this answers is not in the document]
fabiogomessilva: Done !

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+        headers: {'x-opencensus-outgoing-request': 1}
+      };
+
+      assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 0);
+      await httpRequest.get(options).then((result) => {
+        assert.equal(result, 'Ok');
+        assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 0);
+      });
+    });
+  });  // describe pathcOutgoing
```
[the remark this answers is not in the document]
fabiogomessilva: Removed!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+  const tracer = new classes.Tracer();
+  const rootSpanVerifier = new RootSpanVerifier();
+  tracer.start({samplingRate: 1, logger: log});
+
+  it('should return a plugin', () => {
+    assert.ok(plugin instanceof HttpPlugin);
+  });
+
+  before(() => {
+    plugin.applyPatch(http, tracer, VERSION);
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+    rootSpanVerifier.endedRootSpans = [];
+    nock.cleanAll();
+  });
+
+  after(() => {
+    server.close();
+  });
+
+  /** Should intercept outgoing requests */
+  describe('patchOutgoingRequest()', () => {
```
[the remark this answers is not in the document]
fabiogomessilva: Done for Attributes !

packages/opencensus-instrumentation-http/tsconfig.json
```diff
@@ -0,0 +1,22 @@
...
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "noImplicitThis": false,
+    "pretty": true,
+    "module": "commonjs",
+    "target": "es6",
+    "strictNullChecks": false,
+    "sourceMap": false
```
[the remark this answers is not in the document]
fabiogomessilva: Removed from http, https and propagation-b3. It should be removed from opencensus-nodejs also, but It will be in another PR.

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+      await httpRequest.get(options).then((result) => {
+        assert.ok(
+            rootSpanVerifier.endedRootSpans[0].name.indexOf(testPath) >= 0);
+        assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 1);
+      });
+    });
+  });
+
+  /** Should not intercept incoming and outgoing requests */
+  describe('applyUnpatch()', () => {
```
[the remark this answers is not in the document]
fabiogomessilva: > This tests relies on a specific order in which tests are executed. Is it possible to make this test more isolated?

Could you clarify on that? Do you mean specific order related to others tests? Anyway, a TODO was added.

packages/opencensus-instrumentation-https/README.md
```diff
@@ -0,0 +1,19 @@
+# OpenCensus HTTPS Instrumentation for Node.js
```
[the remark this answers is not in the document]
fabiogomessilva: All READMEs are TODOs.

packages/opencensus-instrumentation-https/encryption/cert.pem
```diff
@@ -0,0 +1,24 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-https/src/https.ts
```diff
@@ -0,0 +1,62 @@
...
+  applyPatch(moduleExporters: any, tracer: types.Tracer, version: string) {
+    this.setPluginContext(moduleExporters, tracer, version);
+    this.logger = tracer.logger || logger.logger('debug');
+
+    shimmer.wrap(
+        moduleExporters && moduleExporters.Server &&
+            moduleExporters.Server.prototype,
+        'emit', this.patchIncomingRequest());
+
+    shimmer.wrap(moduleExporters, 'get', this.patchOutgoingRequest());
```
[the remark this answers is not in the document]
fabiogomessilva: > As noted above, there is a possibility that this calls through to http.get. (Same thing with request). Not sure if there are implications here -- just a heads up.

Ok

> Also, where is the patch for https.request?

It was not needed. A TODO was added to review this approach.

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+
+
+/** Propagates span context through B3 Format propagation. */
+export class B3Format implements types.Propagation {
+  /**
+   * Gets the trace context from a request headers. If there is no trace context
+   * in the headers, null is returned.
+   * @param getter
+   */
+
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+  extract(getter: types.HeaderGetter): types.SpanContext {
+    // tslint:disable-next-line
```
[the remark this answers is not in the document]
fabiogomessilva: Removed.

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+   */
+  generate(): types.SpanContext {
+    return {
+      traceId: uuid.v4().split('-').join(''),
+      // tslint:disable-next-line:ban Needed to parse hexadecimal.
+      spanId: parseInt(crypto.randomBytes(6).toString('hex'), 16).toString(),
+      options: SAMPLED_VALUE
+    } as types.SpanContext;
+  }
+}
```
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+      if (spanContext && (spanContext.options & SAMPLED_VALUE) !== 0) {
+        setter.setHeader(X_B3_SAMPLED, SAMPLED_VALUE.toString());
+      } else {
+        setter.setHeader(X_B3_SAMPLED, NOT_SAMPLED_VALUE.toString());
+      }
+    }
+  }
+
+  /**
+   * Generate SpanContexts for testing
```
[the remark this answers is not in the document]
fabiogomessilva: > This function should be used in the http/https instrumentation if it isn't already.

If there is no propagation the logic to generate traceId, spanId are inside span classes.

packages/opencensus-propagation-b3/src/b3Format.ts
fabiogomessilva: Done !

packages/opencensus-propagation-b3/src/b3Format.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * Adds a trace context in a request headers.
+   * @param setter
+   * @param spanContext
+   */
+  inject(setter: types.HeaderSetter, spanContext: types.SpanContext): void {
+    if (setter) {
+      setter.setHeader(
+          X_B3_TRACE_ID, spanContext && spanContext.traceId || 'undefined');
```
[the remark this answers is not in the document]
fabiogomessilva: > Using types ....

I agree, but we also have to suppose users that do not use types, right?

packages/opencensus-propagation-b3/src/b3Format.ts
fabiogomessilva: Done. It was not for this file. Thanks.

packages/opencensus-instrumentation-https/package.json
fabiogomessilva: Done

packages/opencensus-instrumentation-https/package.json
fabiogomessilva: Done!

packages/opencensus-instrumentation-https/package.json
fabiogomessilva: Done !

packages/opencensus-propagation-b3/package.json
fabiogomessilva: Done !

packages/opencensus-propagation-b3/src/b3Format.ts
fabiogomessilva: In this case for options as a number.

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -0,0 +1,206 @@
...
+  });  // describe pathcOutgoing
+
+
+  /** Should intercept incoming requests */
+  describe('patchIncomingRequest()', () => {
+    it('should create a root span for incoming requests', async () => {
+      const testPath = '/incoming/rootSpan/';
+
+      const options = {host: 'localhost', path: testPath, port: 3000};
+      shimmer.unwrap(http, 'get');
```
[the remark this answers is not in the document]
fabiogomessilva: I think we can keep only one class, maybe a property to setup as a client or server.  I will add a TODO.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -17,88 +17,121 @@
...
 import * as shimmer from 'shimmer';
 import * as url from 'url';
 import * as uuid from 'uuid';
 
-// import * as Debug from 'debug';
-// const debug = Debug('opencensus');
 
+// TODO: maybe we should have a setup as a Client or as Server.
+
+type HttpModuleType = typeof http;
```
kjin: Nit: `HttpModule` is fine, it's already a type. Also, call the module itself `httpModule` rather than `http`.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -17,88 +17,121 @@
...
+type RequestFunction = typeof http.request;
+
+/** Http instrumentation plugin for Opencensus */
 export class HttpPlugin extends classes.BasePlugin {
+  /**
+   * Atributes Names according to Opencensus HTTP Specs
+   * https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md
+   */
+  readonly ATRIBUTE_HTTP_HOST = 'http.host';
```
kjin: Spelling: `ATTRIBUTE` (this should be changed in the docs as well)

kjin: Also, these should be `static`. They're not properties of the plugin instance.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -17,88 +17,121 @@
...
+   * @param moduleExports The http module exports
    * @param tracer A tracer instance to create spans on.
    * @param version The package version.
    */
-  // tslint:disable:no-any
-  applyPatch(moduleExporters: any, tracer: types.Tracer, version: string) {
-    this.setPluginContext(moduleExporters, tracer, version);
+  // disable-next-line to disable no-any check for moduleExporters parameter
+  // tslint:disable-next-line
+  applyPatch(moduleExports: any, tracer: types.Tracer, version: string) {
```
kjin: `moduleExports: HttpModule`

Then you can also remove the disable tslint lines.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -17,88 +17,121 @@
...
-            moduleExporters.Server.prototype,
-        'emit', this.patchIncomingRequest());
+    if (moduleExports && moduleExports.Server &&
+        moduleExports.Server.prototype) {
+      shimmer.wrap(
+          moduleExports && moduleExports.Server &&
+              moduleExports.Server.prototype,
+          'emit', this.patchIncomingRequest());
+    } else {
+      this.logger.error(
```
kjin: 👍

packages/opencensus-instrumentation-http/src/http.ts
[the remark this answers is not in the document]
kjin: You can still set that in the first line of the function that is currently being returned:

```ts
return (original: RequestFunction): types.Func<http.ClientRequest> => {
  const plugin = this;
  // ...
}
```

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -17,88 +17,121 @@
...
   /**
    * Creates spans for incoming requests, restoring spans' context if applied.
    * @param plugin The HTTP plugin itself.
    */
   patchIncomingRequest() {
-    const plugin: HttpPlugin = this;
-    return (original: Function): Function => {
-      // tslint:disable:no-any
+    const plugin = this;
+    return (original: RequestFunction): types.Func<http.ClientRequest> => {
```
kjin: The return type should be `RequestFunction` as well, right?

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -173,8 +220,8 @@ export class HttpPlugin extends classes.BasePlugin {
 
         plugin.logger.debug('%s plugin outgoingRequest', plugin.moduleName);
         const traceOptions = {
-          name: `${request.method ? request.method : 'GET'}  ${
-              arguments[0].pathname}`,
+          name:
+              `${request.method ? request.method : 'GET'}  ${options.pathname}`,
```
kjin: Nit: Is the extra space between the two expressions needed?

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -204,13 +251,15 @@ export class HttpPlugin extends classes.BasePlugin {
    * @param original The original patched function.
    * @param args The arguments to the original function.
    */
-  // tslint:disable:no-any
-  makeRequestTrace(request: any, args: any, plugin: HttpPlugin): any {
-    return (span: types.Span) => {
+  // disable-next-line to disable no-any check for args parameter
+  // tslint:disable-next-line
```
kjin: `tslint:disable-next-line:no-any`

kjin: Though, it looks like you never need to use anything other than `args[0]`. Why not just pass the first argument in?

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -226,66 +275,75 @@ export class HttpPlugin extends classes.BasePlugin {
...
-          span.addAttribute('http.method', request.method);
-          span.addAttribute('http.path', args[0].pathname);
-          span.addAttribute('http.route', args[0].path);
+          const method = response.method ? response.method : 'GET';
+          const reqUrl = url.parse(args[0]);
+          span.addAttribute(plugin.ATRIBUTE_HTTP_HOST, reqUrl.hostname);
+          span.addAttribute(plugin.ATRIBUTE_HTTP_METHOD, method);
+          span.addAttribute(plugin.ATRIBUTE_HTTP_PATH, reqUrl.pathname);
+          span.addAttribute(plugin.ATRIBUTE_HTTP_ROUTE, reqUrl.path);
+          span.addAttribute(plugin.ATRIBUTE_HTTP_USER_AGENT, null);
```
kjin: Looks like the user agent disappeared here? Is this intentional?

packages/opencensus-instrumentation-http/test/test-http.ts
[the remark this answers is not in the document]
kjin: If we run only this test (using `describe.only` or similar), it will fail. Or at least that's what it looks like... I might be wrong?

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -102,40 +107,143 @@ describe('HttpPlugin', function() {
...
+           const options = {name: 'TestRootSpan'};
+           return tracer.startRootSpan(options, async root => {
+             await httpRequest.get(`${urlHost}${testPath}`).then((result) => {
+               assert.ok(root.name.indexOf('TestRootSpan') >= 0);
+               assert.strictEqual(root.spans.length, 1);
+               assert.ok(root.spans[0].name.indexOf(testPath) >= 0);
+               assert.strictEqual(root.traceId, root.spans[0].traceId);
+
+               const span = root.spans[0];
+               assert.strictEqual(
```
kjin: Suggestion: A lot of this code looks the same. For maintainability it would be a good idea to see what can be factored out into a common function.

packages/opencensus-propagation-b3/src/b3Format.ts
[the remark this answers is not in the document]
kjin: It makes sense to check pre-conditions -- but consider that the TypeScript definitions are the public API of the module, and also the source from which public documentation is generated, and therefore the source of truth. If someone is using plain JavaScript and doesn't specify `setter`, then they're not using the API correctly, and the onus is on them to correct it.

packages/opencensus-propagation-b3/src/b3Format.ts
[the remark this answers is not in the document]
kjin: Our original thinking was that generating span context should be an implementation detail of `Propagation` rather than `Span`. But actually, I think it makes more sense the way you put it -- the way trace context is generated should not depend on how it is propagated. @ofrobots what do you think here?

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Done by ``gts fix``.
 ``gts check`` fails otherwise.

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Agree.

packages/opencensus-instrumentation-http/test/test-http.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Some side effects on the compilation. A TODO was added.

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Some side effects on the compilation. A TODO was added.

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: A refactore was made.

packages/opencensus-propagation-b3/package.json
fabiogomessilva: Yes, we are using ``gts check`` for all the packages. I think it was an outdated version.

packages/opencensus-propagation-b3/tsconfig.full.json
fabiogomessilva: I think it was an outdated version.

packages/opencensus-instrumentation-http/src/http.ts
kjin: Hmm... I'm wondering about the two spaces in the string, which I don't think has to do with code formatting:
```js
`${request.method ? request.method : 'GET'}  ${options.pathname}`
                                           ^^
```

packages/opencensus-instrumentation-http/src/http.ts
kjin: Nit: The type of `options` should now be known (isn't it `RequestOptions | string | URL`)

fabiogomessilva:
I'm unable to reply on coments like:
#14 ([comment](https://github.com/census-instrumentation/opencensus-node/pull/14#discussion_r189407130))
#14 ([comment](https://github.com/census-instrumentation/opencensus-node/pull/14#discussion_r189410516))

packages/opencensus-propagation-b3/src/b3Format.ts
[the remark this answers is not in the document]
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Should be only one space. Done !

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Some compiler and/or tests issues. A TODO was added.

PR #18, opened 2018-05-23
feat: add package exporter-trace-stackdriver to opencensus-node

eduardoemery:
This PR adds Stackdriver exporter package to opencensus-node. Unit tests are included.


packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
```diff
@@ -0,0 +1,167 @@
...
+      });
+    });
+  }
+
+
+
+  /**
+   * Gets the Google Application Credentials from the environment variables,
+   * authenticates the client and calls a method to send the traces data.
+   * @param sendTrace
```
kjin: Nit: Remove this line.

packages/opencensus-exporter-trace-stackdriver/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "mocha": "^5.0.4",
+    "ncp": "^2.0.0",
+    "nock": "^9.2.6",
+    "nyc": "^11.8.0",
+    "ts-node": "^4.0.0",
+    "typescript": "~2.7.2"
+  },
+  "dependencies": {
+    "@opencensus/opencensus-core": "^0.0.1",
+    "google-cloud": "^0.57.0",
```
kjin: Is this dependency used anywhere?

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
```diff
@@ -0,0 +1,167 @@
...
+   * @param stackdriverTraces
+   */
+  private authorize(stackdriverTraces) {
+    return new Promise((resolve, reject) => {
+      return google.auth.getApplicationDefault(
```
kjin: This looks like an older version of Google Auth Library. If you add a dependency on `google-auth-library@1.5` (or 1.4+) I believe it comes with a Promise API out-of-the-box.

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
```diff
@@ -0,0 +1,167 @@
...
+  /**
+   * Translates root span data to Stackdriver's trace format.
+   * @param root
+   */
+  private translateTrace(root: types.RootSpan) {
+    const spanList = root.spans.map(span => this.translateSpan(span));
+    spanList.push(this.translateSpan(root));
+
+    return {
+      'projectId': this.projectId,
```
kjin: Nit: `''` in object keys are not necessary (here and in `translateSpan`).

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
```diff
@@ -0,0 +1,203 @@
...
+            rootSpan.end();
+
+            return exporter.publish([rootSpan]).then((result: string) => {
+              assert.ok(result.indexOf('sendTrace sucessfully') >= 0);
+            });
+          });
+    });
+
+
+    it('should fail exporting by network error', async () => {
```
kjin: If `dryrun` is set to false, do we expect a real network error to occur here?

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
```diff
@@ -0,0 +1,203 @@
...
+import * as nocks from './nocks';
+
+process.env.NODE_TLS_REJECT_UNAUTHORIZED = '0';
+let PROJECT_ID = 'fake-project-id';
+
+function checkEnvironoment(): boolean {
+  return false;
+}
+
+describe('Stackdriver Exporter', function() {
```
kjin: My general belief here is that in this test file a `dryrun` flag is not needed, and we can _always_ use nocks to simulate endpoints. We can add a separate test (in, say, a `system-test` or `test/integration` directory) that then uses real network endpoints to ensure that the module works with the Stackdriver API. That way it's easy to identify and run all non-network tests at once, and separately run all network/authenticated tests at once.

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
eduardoemery: Done

packages/opencensus-exporter-trace-stackdriver/package.json
eduardoemery: Dependency removed.

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
eduardoemery: We updated to the last version. It's working with promises now.

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
eduardoemery: Done

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
kjin: Hmm... I've thought about this a little more and I think it should be OK. That being said, I think `dryrun` should be coming from its own environmental variable -- maybe something like `OPENCENSUS_NETWORK_TESTS` -- so that we can easily "turn on" network accesses for tests in all packages with a single switch.

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
kjin: Hmm... in that case is it necessary to have network access at all for this test? We don't need to authenticate with a real server if we're just going to hit a fake endpoint that will fail.

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
```diff
@@ -0,0 +1,203 @@
...
+    it('should fail exporting by network error', async () => {
+      nock('https://cloudtrace.googleapis.com')
+          .persist()
+          .intercept(
+              '/v1/projects/' + exporterOptions.projectId + '/traces', 'patch')
+          .reply(443, 'Simulated Network Error');
+
+      if (dryrun) {
+        nocks.oauth2((body) => true);
+        const scope = nocks.patchTraces(PROJECT_ID, null, null, false);
```
kjin: Is this line needed? It looks like it is covered by the earlier `nock('https://cloudtrace.googleapis.com')...`

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
```diff
@@ -0,0 +1,203 @@
...
+                  failExporter.failBuffer[0].traceId,
+                  rootSpan.spanContext.traceId);
+            });
+          });
+    });
+
+    it('should export traces to stackdriver', () => {
+      if (dryrun) {
+        nocks.oauth2((body) => true);
+        const scope = nocks.patchTraces(PROJECT_ID, null, null, false);
```
kjin: When using nocks (including `nocks.oauth2`), can you use `scope.done` to ensure that the endpoint was hit? If you think it's not necessary, then I believe the `const scope = ` is not needed.

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
```diff
@@ -0,0 +1,157 @@
...
+   */
+  onEndSpan(root: types.RootSpan) {
+    this.exporterBuffer.addToBuffer(root);
+  }
+
+  /**
+   * Publishes a list of root spans to Stackdriver.
+   * @param rootSpans
+   */
+  publish(rootSpans: types.RootSpan[]) {
```
kjin: The base class seems to return `void`?

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
eduardoemery: In fact it's not necessary to authenticate with a real server. Changed to always authenticate with nock.

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
eduardoemery: That's a nice idea, just implemented that. Used `OPENCENSUS_NETWORK_TESTS` as you suggested.

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
eduardoemery: Indeed it isn't necessary. Line removed.

packages/opencensus-exporter-trace-stackdriver/test/test-stackdriver.ts
eduardoemery: Removed the `const scope =` since we are not using.

packages/opencensus-exporter-trace-stackdriver/src/stackdriver.ts
eduardoemery: TODO let's review the base class in future reviews.

PR #19, opened 2018-05-25
feat: add package exporter-trace-zipkin to opencensus-node

djonathascardoso:
This PR adds zipkin trace exporter package to opencensus-node. Unit tests are included.


packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+   */
+  onEndSpan(root: types.RootSpan) {
+    this.buffer.addToBuffer(root);
+  }
+
+  /**
+   * Send a trace to zipkin service
+   * @param zipkinTraces Trace translated to Zipkin Service
+   */
+  private async sendTraces(zipkinTraces: TranslatedSpan[]) {
```
kjin: This method doesn't need the `async` keyword.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+          res.on('end', () => {
+            resolve(
+                {statusCode: res.statusCode, statusMessage: res.statusMessage});
+          });
+        });
+
+        /** Request error event */
+        req.on('error', (e) => {
+          reject({
+            statusCode: 500,
```
kjin: Is the status code defined on `e`? Can we use that status code instead -- since `500` may not always be accurate?

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+            statusMessage: `Problem with request: ${e.message}`
+          });
+        });
+
+        /** Request body */
+        const spansJson: string[] =
+            zipkinTraces.map((span) => JSON.stringify(span));
+        spansJson.join('');
```
kjin: Should this be `spansJson.join(',');`? Or maybe just `const outputJson = JSON.stringify(zipkinTraces)`?

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+      port: this.zipkinUrl.port,
+      path: this.zipkinUrl.path,
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      }
+    };
+
+    return new Promise((resolve, reject) => {
+      try {
```
kjin: Can we narrow the `try` to just what can throw synchronously? Which I think is just the `JSON.stringify`, as far as I can tell.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+        reject(e);
+      }
+    });
+  }
+
+  /**
+   * Mount a list (array) of spans translated to Zipkin format
+   * @param rootSpan Trace to be translated
+   */
+  private mountSpanList(spanList: TranslatedSpan[], rootSpan: types.RootSpan):
```
kjin: It seems like `spanList` is not really necessary, since you can just construct a new array within this function instead.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+      kind: 'SERVER',
+      timestamp: (span.startTime.getTime() * 1000).toFixed(),
+      duration: (span.duration * 1000).toFixed(),
+      debug: true,
+      shared: true,
+      localEndpoint: {serviceName: this.serviceName}
+    } as TranslatedSpan;
+
+    if (rootSpan) {
+      spanTraslated.parentId = rootSpan.id;
```
kjin: Doesn't `span` expose a `parentSpanId` field? Is it possible to use that?

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  /**
+   * Send the rootSpans to zipkin service
+   * @param rootSpans RootSpan array
+   */
+  publish(rootSpans: types.RootSpan[]) {
+    let zipkinTraces: TranslatedSpan[] = [];
+    for (const trace of rootSpans) {
+      zipkinTraces = this.mountSpanList(zipkinTraces, trace);
```
kjin: In conjunction with the suggested change to `mountSpanList`:

`zipkinTraces.push(...this.mountSpanList(/*...*/))`

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    return this.sendTraces(zipkinTraces)
+        .then((result) => {
```
kjin: `then` statement seems unnecessary.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    return spanTraslated;
+  }
+
+  /**
+   * Send the rootSpans to zipkin service
+   * @param rootSpans RootSpan array
+   */
+  publish(rootSpans: types.RootSpan[]) {
```
kjin: I believe `publish` returns `void` in the `Exporter` interface? Why is a Promise being returned here?

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
```diff
@@ -0,0 +1,145 @@
...
+      .post('/api/v2/spans')
+      .reply(202)
+      .post('/api/v2/spanswrong')
+      .reply(404);
+};
+
+/**
+ * Checks if zipkin service is not running for tests
+ */
+const isNotZipkinRunning = (callback: Function) => {
```
kjin: Instead of using the existence of a `Zipkin` server to alter test behavior, can we use an environmental variable like in https://github.com/census-instrumentation/opencensus-node/pull/18#discussion_r190952198

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
```diff
@@ -0,0 +1,145 @@
...
+});
+
+/** Zipkin tests */
+describe('Zipkin Exporter', function() {
+  /** Desabling the timeout for tests */
+  this.timeout(0);
+  /** Should create a Zipkin instance */
+  describe('new Zipkin()', () => {
+    it('should create a Zipkin instance', () => {
+      const zipkin = new ZipkinTraceExporter(zipkinOptions);
```
kjin: I think this test might be unnecessary. `new T` will definitely give you an `instanceof T`.

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
```diff
@@ -0,0 +1,145 @@
...
+      assert.ok(zipkin instanceof ZipkinTraceExporter);
+    });
+  });
+
+  /** Should called when a rootSpan ended */
+  describe('onEndSpan()', () => {
+    it('Should called when a rootSpan ended', () => {
```
kjin: This seems like it is testing `Tracer` rather than the exporter. Can you re-word this to say something like `'should add spans to the exporter buffer'` or similar.

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
```diff
@@ -0,0 +1,145 @@
...
+            rootSpan.end();
+            return exporter.publish([rootSpan]).then((result) => {
+              const resultObject = result as RequestResponse;
+              assert.equal(resultObject.statusCode, 202);
+            });
+          });
+    });
+  });
+
+  /** Should send traces to Zipkin service */
```
kjin: Nit: These comments can be removed if they repeat what the string passed to `it`.

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
```diff
@@ -0,0 +1,145 @@
...
+  describe('publish() with a wrong Zipkin url', () => {
+    it('shouldn\'t send traces to Zipkin service and return an 404 error',
+       () => {
+         /** adding a string to get a wrong url */
+         zipkinOptions.url += 'wrong';
```
kjin: Prefer cloning over mutating `zipkinOptions`. Otherwise this test will have an impact on the next test (if there is one).

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
```diff
@@ -0,0 +1,145 @@
...
+         const tracer = new classes.Tracer();
+         tracer.start(defaultConfig);
+
+         return tracer.startRootSpan(
+             {name: 'root-test'}, async (rootSpan: types.RootSpan) => {
+               const span = rootSpan.startChildSpan('spanTest', 'spanType');
+               span.end();
+               rootSpan.end();
+               return exporter.publish([rootSpan]).then((result) => {
+                 const resultObject = result as RequestResponse;
```
kjin: What is the uncasted type of `result` here?

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: ok

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: Ok. Done.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: Done.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: I put this loop into the mountSpanList method, now it gets the rootSpan array as a parameter. The result was:
```const spanList = this.mountSpanList(rootSpans);```

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
djonathascardoso: Ok, created the environmental variable.

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
djonathascardoso: Done.

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
djonathascardoso: Done.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: Done.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: Agree. Done.

packages/opencensus-exporter-trace-zipkin/src/zipkin.ts
djonathascardoso: Add a TODO

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
djonathascardoso: Removed.

packages/opencensus-exporter-trace-zipkin/test/test-zipkin.ts
djonathascardoso: Cast removed. It was not necessary.

PR #24, opened 2018-06-04
wip: feat: add package instrumentation-http2 to opencensus-node

eduardoemery:
This PR adds HTTP2 instrumentation package to opencensus-node. Unit tests are included.


packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+    }
+
+    if (this.patchedConnect) {
+      shimmer.unwrap(plugin.client, 'request');
+    } else {
+      shimmer.unwrap(plugin.moduleExports, 'connect');
+    }
+  }
+
+  patchConnect() {
```
kjin: These functions should be renamed. They don't patch the corresponding functions, but rather return a function that patches the corresponding functions. A name like `getPatchConnectFn` would be more appropriate.

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+export type ConnectFunction = typeof http2.connect;
+type RequestFunction =
+    (this: http2.ClientHttp2Session, headers?: http2.OutgoingHttpHeaders,
+     options?: http2.ClientSessionRequestOptions) => http2.ClientHttp2Stream;
+export type CreateServerFunction = typeof http2.createServer;
+
+/** Http2 instrumentation plugin for Opencensus */
+export class Http2Plugin extends HttpPlugin {
+  server: http2.Http2Server|http2.Http2SecureServer;
+  client: http2.ClientHttp2Session;
```
kjin: If `server` and `client` are singular instances of HTTP2 servers and clients... how will you handle multiple instances?

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+            return plugin.client;
+          };
+    };
+  }
+
+  patchRequest() {
+    const plugin = this;
+    return (original: RequestFunction): types.Func<http2.ClientHttp2Stream> => {
+      return function patchedRequest(
+                 headers: http2.IncomingHttpHeaders): http2.ClientHttp2Stream {
```
kjin: Aren't these outgoing HTTP headers?

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+        // Do not trace ourselves
+        if (headers['x-opencensus-outgoing-request']) {
+          return original.apply(this, arguments);
+        }
+
+        const request = original.apply(this, arguments);
+        plugin.tracer.wrapEmitter(request);
+
+        const traceOptions = {
+          name: `${headers[':method'] ? headers[':method'] : 'GET'} ${
```
kjin: Nit: Can use `headers[':method'] || 'GET'` here.

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+    };
+  }
+
+  makeHttp2RequestTrace(
+      request: http2.ClientHttp2Stream, headers: http2.IncomingHttpHeaders,
+      plugin: Http2Plugin): types.Func<http2.ClientHttp2Stream> {
+    return (span: types.Span): http2.ClientHttp2Stream => {
+      if (!span) return request;
+
+      const setter = {
```
kjin: Does `const setter: types.HeaderSetter = ...` work here? Prefer that to typecasting.

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+
+      request.on('response', (responseHeaders: http2.IncomingHttpHeaders) => {
+        span.addAttribute(
+            Http2Plugin.ATTRIBUTE_HTTP_STATUS_CODE,
+            `${responseHeaders[':status']}`);
+        span.status = plugin.traceStatus(+responseHeaders[':status']);
+      });
+
+      request.on('end', () => {
+        const userAgent =
```
kjin: Should this be `string|null`? Then wrap the `addAttribute` for user agent in a conditional.

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+
+        const traceOptions = {
+          name: headers[':path'],
+          type: 'SERVER',
+          spanContext: propagation ? propagation.extract(getter) : null
+        } as types.TraceOptions;
+
+        let statusCode: number = null;
+        const originalRespond = stream.respond;
+        stream.respond = function(this: http2.Http2Stream) {
```
kjin: Can you add a comment here as to why you are wrapping `respond`, and why it only needs to be wrapped once? Where does it get called? Is it possible for it to be called more than once?

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+          return stream.respond.apply(this, arguments);
+        };
+
+        return plugin.tracer.startRootSpan(traceOptions, rootSpan => {
+          if (!rootSpan) return original.apply(this, arguments);
+
+          plugin.tracer.wrapEmitter(stream);
+
+          const originalEnd = stream.end;
+          stream.end = function(this: http2.Http2Stream) {
```
kjin: Where is `end` being called? How is this different from listening on the `'end'` event?

Also, are errors handled as well?

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+        span.addAttribute(HttpPlugin.ATTRIBUTE_HTTP_ERROR_MESSAGE, err.message);
+
+        span.end();
+      });
+
+      return request;
+    };
+  }
+
+  patchCreateServer() {
```
kjin: All `patch` functions should be `private`.

packages/opencensus-instrumentation-http2/test/test-http2.ts
```diff
@@ -0,0 +1,308 @@
...
+         await http2Request.get(client, requestOptions).then((result) => {
+           assert.strictEqual(result, `${statusCode}`);
+           assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 0);
+         });
+       });
+  });
+
+
+  /** Should intercept incoming requests */
+  describe('patchEmit()', () => {
```
kjin: We should be testing user-visible behavior here. `patchEmit` is not meant to be user-visible. Can you make this (and other `describe` clauses) a more public-friendly description of what is being tested here?

packages/opencensus-instrumentation-http2/tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "noImplicitThis": false,
```
kjin: Is it possible to remove this compiler option?

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+  }
+
+  patchConnect() {
+    const plugin = this;
+    return (original: ConnectFunction):
+               types.Func<http2.ClientHttp2Session> => {
+      return function patchedConnect(authority: string):
+          http2.ClientHttp2Session {
+            plugin.client = original.apply(this, arguments);
+            shimmer.wrap(plugin.client, 'request', plugin.patchRequest());
```
kjin: Is `request` the only way to communicate with an HTTP2 server? I believe there is a low-level API as well.

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -0,0 +1,283 @@
...
+
+export type Http2Module = typeof http2;
+export type ConnectFunction = typeof http2.connect;
+type RequestFunction =
+    (this: http2.ClientHttp2Session, headers?: http2.OutgoingHttpHeaders,
+     options?: http2.ClientSessionRequestOptions) => http2.ClientHttp2Stream;
+export type CreateServerFunction = typeof http2.createServer;
+
+/** Http2 instrumentation plugin for Opencensus */
+export class Http2Plugin extends HttpPlugin {
```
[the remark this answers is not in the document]
eduardoemery: We use HttpPlugin to reuse the atributes and `traceStatus()`, which translates the status code to [trace status code](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes). We like this approach because it reuses code, but I understand your point of view. Would you prefer to remove the dependency or keep it this way?

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: Makes total sense. We choose this name because the inner function is anonymous and we didn't see a problem on calling it this way. Also, HTTP uses this "name pattern", if you think it's best to change it here, I suggest we change it on HTTP as well.

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: It works, just changed, but `userAgent` comes as `string | string[]`, so I had to cast to `string` before passing as attribute.

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: Added a comment explaining. Please let me know if it doesn't make it clearer.

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: Unfortunately, server stream has no end event, it calls `stream.end()` when it's done responding. That's why we wrap it.

Errors are catched by wraping the respond event that sends the status code.

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: Ok, just changed that. As we needed `patchRequest()` and `patchEmit()` in the tests, I created a method `patchForTest()`that does what we need. It is intended for testing purposes only.

packages/opencensus-instrumentation-http2/test/test-http2.ts
eduardoemery: What do you think of changing `'patchEmit()'` to `'Patch incoming requests'` and `'patchRequest()'` to `'Patch outgoing requests'`?

packages/opencensus-instrumentation-http2/tsconfig.json
eduardoemery: As we use `original.apply(this, argumets)` a lot, removing this compiler option causes a lot of trouble in this plugin.

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: Hmm I couldn't find it. Please let me know if there is, cause we should patch it too.

packages/opencensus-instrumentation-http2/tsconfig.json
kjin: TypeScript has a workaround for this -- anywhere `this` is needed in an ES5 function, you can use this syntax:

```ts
// `this` is now explicit
function (this: Http2Server, arg1: string, /*...*/) {
  this.connect();
}
```

packages/opencensus-instrumentation-http2/src/http2.ts
kjin: Yep, I would be in favor of that.

packages/opencensus-instrumentation-http2/src/http2.ts
kjin: Ok, makes sense. I forgot that outgoing headers can be `string[]` as well.

packages/opencensus-instrumentation-http2/test/test-http2.ts
kjin: Yep, that's an improvement. `Instrumenting incoming requests` would be more accurate, because even patching is not user-visible -- however, instrumenting is (because users will observe traces being reported)

packages/opencensus-instrumentation-http2/src/http2.ts
kjin: `server` and `client` always point toward the most-recently constructed instances, so won't `unpatch` only unpatch those, and not previously constructed ones?

A secondary concern here is that the `clientUrl` value will not be accurate if there are concurrent client requests. It seems like it will always hold the URL of the most recently connected client, but its value is used when a client request ends.

Generally, I think all of these should not be member variables. They aren't properties of the plugin, after all.

packages/opencensus-instrumentation-http2/src/http2.ts
kjin: Never mind -- I was mistaken.

packages/opencensus-instrumentation-http2/src/http2.ts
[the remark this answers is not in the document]
eduardoemery: I agree, specially `traceStatus` and, maybe, other things that are defined in OC spec should be in `opencensus-core` if it helps on code re usability.

packages/opencensus-instrumentation-http2/test/test-http2.ts
eduardoemery: Great idea! I'll update that.

packages/opencensus-instrumentation-http2/test/test-http2.ts
```diff
@@ -130,7 +130,7 @@ describe('Http2Plugin', () => {
     });
 
     after(() => {
-      shimmer.wrap(server.constructor.prototype, 'emit', plugin.patchEmit());
+      plugin.patchForTest('emit');
```
kjin: Actually, I'm wondering why we need to unpatch the server to test the client in the first place (and vice versa). Is it possible to just ignore traces from the server? Whether or not we are using `patchForTest` we are still depending on class-private behavior.

packages/opencensus-instrumentation-http2/tsconfig.json
eduardoemery: Great to know. Updated and removed the compiler option.

packages/opencensus-instrumentation-http2/test/test-http2.ts
eduardoemery: I see your point, makes sense. I removed that method and did just that, ignored the traces we don't care.

packages/opencensus-instrumentation-http2/src/http2.ts
eduardoemery: In fact `clientUrl` where only holding the url from the first client instance. I removed the `clientUrl` and the unpatch of `Connect`, now, every time a new `client` is created `request` is patched passing the right one.

I don't think `server` and `client` should be properties of plugin too. I only used them to later unpatch `request` and `emit`. I removed them, but now only the `client` and `server` constructors will be unwrapped, any existing `client` or `server` instances will still trace.

PR #25, opened 2018-06-04
fix: unref ExporterBuffer flush timer

kjin:
The core unit tests take an unexpectedly long time before exiting... this change fixes that by `unref`'ing the `Timer` object used when doing periodic trace buffer flushing.


kjin:
TBR=@ofrobots

PR #27, opened 2018-06-05
refactor: add third party files to opencensus-exporter-jaeger

fabiogomessilva:
This PR adds third-party files used by opencensus-exporter-jaeger according to OSPR requirements


kjin:
Note: changing this to `chore` as this is not considered a "new feature".

PR #29, opened 2018-06-05
doc: update readme.md files and add examples

eduardoemery:


packages/examples/automatic_tracing/README.md
```diff
@@ -0,0 +1,82 @@
+# OpenCensus Node.js Automatic Tracing
```
kjin: Does it make more sense to put `examples` at the top level rather than in `packages`?

packages/opencensus-instrumentation-http/README.md
```diff
@@ -1,15 +1,10 @@
 # OpenCensus HTTP Instrumentation for Node.js
 [![Gitter chat][gitter-image]][gitter-url]
 
-OpenCensus HTTP Instrumentation allows the user to automatically collect trace data.
+OpenCensus HTTP Instrumentation allows the user to automatically collect trace data. This package is used by [OpenCensus Node.js](https://github.com/census-instrumentation/opencensus-node/tree/master/packages/opencensus-nodejs) to instrument HTTP.
```
kjin: We should mention that this is for both servers and clients. (Same with HTTPS.)

packages/opencensus-nodejs/README.md
```diff
@@ -1,12 +1,109 @@
...
+| [`logLevel`](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/config/types.ts#L47) | `number` | Level of logger - 0:disable, 1: error, 2: warn, 3: info, 4: debug |
+| [`maximumLabelValueSize`](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/config/types.ts#L52) | `number` | The maximum number of characters reported on a label value |
+| [`plugin`](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/config/types.ts#L68) | `PluginNames` | A list of trace instrumentations plugins to load |
+| [`exporter`](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/config/types.ts#L70) | `Exporter` | An exporter object |
+| [`logger`](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/config/types.ts#L72) | `Logger` | An instance of a logger |
+
+**TODO: Choose between the two options**
```
kjin: If this refers to the table vs. the bulleted list, I prefer the table.

README.md
```diff
@@ -1,15 +1,111 @@
...
 
+## Installation
+
+Install OpenCensus with:
+
+```bash
+npm install @opencensus/nodejs
+```
+
+## Traces
```
kjin: I think it would be better if info wasn't duplicated across this and the `@o/nodejs` package README as well. I would suggest cutting technical details from this doc, and linking to the other one instead.

packages/opencensus-instrumentation-http/README.md
eduardoemery: Great, updated that.

packages/opencensus-nodejs/README.md
eduardoemery: Awesome, that was my opinion too haha

packages/examples/automatic_tracing/README.md
eduardoemery: Yep. Actually, as we are planning to leave the more technical README at `@opencensus/nodejs` maybe we should put `examples` there. What do you think?

README.md
eduardoemery: Ok, I removed the technical part and added a link to nodejs package. I also left the info about plugin, propagation and exporters to be like a "current state" of the library kinda thing. Let me know if there is anything that would be great to add.

packages/examples/automatic_tracing/README.md
kjin: I have a slight preference for the root directory. Either is better than having it in `packages`.

eduardoemery: Maybe in root makes it more visible. I'm going to change to that.

PR #30, opened 2018-06-06
feat: add package instrumentation-mongodb to opencensus-node

eduardoemery:
This PR adds MongoDB instrumentation package to opencensus-node. Unit tests are included.

### DIFF
--- a/packages/opencensus-instrumentation-mongodb/src/mongodb.ts
+++ b/packages/opencensus-instrumentation-mongodb/src/mongodb.ts
@@ -20,23 +20,28 @@ import {logger} from '@opencensus/opencensus-core';
 import * as mongodb from 'mongodb';
 import * as semver from 'semver';
 import * as shimmer from 'shimmer';
-import {callbackify} from 'util';
 
+export type MongoDB = typeof mongodb;
 
-/** Http instrumentation plugin for Opencensus */
+/** MongoDB instrumentation plugin for Opencensus */
 export class MongoDBPlugin extends classes.BasePlugin {
-  readonly SERVER_FNS = ['insert', 'update', 'remove', 'auth'];
-  readonly CURSOR_FNS_FIRST = ['_find', '_getmore'];
-  readonly SPAN_MONGODB_QUERY_TYPE = 'db.mongodb.query';
-  logger: types.Logger;
+  private readonly SERVER_FNS = ['insert', 'update', 'remove', 'auth'];
+  private readonly CURSOR_FNS_FIRST = ['_find', '_getmore'];
+  private readonly SPAN_MONGODB_QUERY_TYPE = 'db.mongodb.query';
+  private logger: types.Logger;
 
-  /** Constructs a new HttpPlugin instance. */
+  /** Constructs a new MongoDBPlugin instance. */
   constructor(moduleName: string) {
     super(moduleName);
   }
 
-  // tslint:disable-next-line:no-any
-  applyPatch(moduleExports: any, tracer: types.Tracer, version: string) {
+  /**
+   * Patches MongoDB operations.
+   * @param moduleExports The mongodb module exports.
+   * @param tracer A tracer instance to create spans on.
+   * @param version The package version.
+   */
+  applyPatch(moduleExports: MongoDB, tracer: types.Tracer, version: string) {
     this.setPluginContext(moduleExports, tracer, version);
     this.logger = tracer.logger || logger.logger('debug');
 
@@ -45,11 +50,13 @@ export class MongoDBPlugin extends classes.BasePlugin {
     if (moduleExports.Server) {
       this.logger.debug('patching mongodb-core.Server.prototype.command');
       shimmer.wrap(
-          moduleExports.Server.prototype, 'command', this.patchCommand());
+          moduleExports.Server.prototype, 'command' as never,
+          this.getPatchCommand());
       this.logger.debug(
           'patching mongodb-core.Server.prototype functions:', this.SERVER_FNS);
       shimmer.massWrap(
-          [moduleExports.Server.prototype], this.SERVER_FNS, this.patchQuery());
+          [moduleExports.Server.prototype], this.SERVER_FNS as never[],
+          this.getPatchQuery());
     }
 
     if (moduleExports.Cursor) {
@@ -57,14 +64,14 @@ export class MongoDBPlugin extends classes.BasePlugin {
           'patching mongodb-core.Cursor.prototype functions:',
           this.CURSOR_FNS_FIRST);
       shimmer.massWrap(
-          [moduleExports.Cursor.prototype], this.CURSOR_FNS_FIRST,
-          this.patchCursor());
+          [moduleExports.Cursor.prototype], this.CURSOR_FNS_FIRST as never[],
+          this.getPatchCursor());
     }
 
     return moduleExports;
   }
 
-  /** Unpatches all HTTP patched function. */
+  /** Unpatches all MongoDB patched functions. */
   applyUnpatch(): void {
     shimmer.unwrap(this.moduleExports.Server.prototype, 'command');
     shimmer.massUnwrap(this.moduleExports.Server.prototype, this.SERVER_FNS);
@@ -72,34 +79,54 @@ export class MongoDBPlugin extends classes.BasePlugin {
         this.moduleExports.Cursor.prototype, this.CURSOR_FNS_FIRST);
   }
 
-  patchCommand() {
+  /** Creates spans for Command operations */
+  private getPatchCommand() {
+    const plugin = this;
+    return (original: types.Func<mongodb.Server>) => {
+      return function(
+                 // tslint:disable-next-line:no-any
+                 this: mongodb.Server, ns: string, command: any,
+                 // tslint:disable-next-line:no-any
+                 ...args: any[]): mongodb.Server {
+        let resultHandler = args[args.length - 1];
+        if (plugin.tracer.currentRootSpan && arguments.length > 0 &&
+            typeof resultHandler === 'function') {
+          let type: string;
+          if (command.createIndexes) {
+            type = 'createIndexes';
+          } else if (command.findandmodify) {
+            type = 'findAndModify';
+          } else if (command.ismaster) {
+            type = 'isMaster';
+          } else if (command.count) {
+            type = 'count';
+          } else {
+            type = 'command';
+          }
+
+          const span = plugin.tracer.startChildSpan(
+              ns + '.' + type, plugin.SPAN_MONGODB_QUERY_TYPE);
+          resultHandler = plugin.patchEnd(span, resultHandler);
+        }
+
+        return original.apply(this, arguments);
+      };
+    };
+  }
+
+  /** Creates spans for Query operations */
+  private getPatchQuery() {
     const plugin = this;
     return (original: types.Func<mongodb.Server>) => {
       // tslint:disable-next-line:no-any
-      return function(this: mongodb.Server, ns: string, command: any):
+      return function(this: mongodb.Server, ns: string, ...args: any[]):
           mongodb.Server {
-            let span: types.Span;
-            const index = arguments.length - 1;
-            const resultHandler = arguments[index];
-
+            let resultHandler = args[args.length - 1];
             if (plugin.tracer.currentRootSpan && arguments.length > 0 &&
                 typeof resultHandler === 'function') {
-              let type: string;
-              if (command.createIndexes) {
-                type = 'createIndexes';
-              } else if (command.findandmodify) {
-                type = 'findAndModify';
-              } else if (command.ismaster) {
-                type = 'ismaster';
-              } else if (command.count) {
-                type = 'count';
-              } else {
-                type = 'command';
-              }
-
-              span = plugin.tracer.startChildSpan(
-                  ns + '.' + type, plugin.SPAN_MONGODB_QUERY_TYPE);
-              arguments[index] = plugin.patchEnd(span, resultHandler);
+              const span = plugin.tracer.startChildSpan(
+                  ns + '.query', plugin.SPAN_MONGODB_QUERY_TYPE);
+              resultHandler = plugin.patchEnd(span, resultHandler);
             }
 
             return original.apply(this, arguments);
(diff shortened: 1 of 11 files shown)


packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+ * limitations under the License.
+ */
+
+import {types} from '@opencensus/opencensus-core';
+import {classes} from '@opencensus/opencensus-core';
+import {logger} from '@opencensus/opencensus-core';
+import * as mongodb from 'mongodb';
+import * as semver from 'semver';
+import * as shimmer from 'shimmer';
+import {callbackify} from 'util';
```
kjin: FYI, `util.callbackify` is not available in Node 6.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+
+
+/** Http instrumentation plugin for Opencensus */
+export class MongoDBPlugin extends classes.BasePlugin {
+  readonly SERVER_FNS = ['insert', 'update', 'remove', 'auth'];
```
kjin: Make all fields private.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+  readonly SPAN_MONGODB_QUERY_TYPE = 'db.mongodb.query';
+  logger: types.Logger;
+
+  /** Constructs a new HttpPlugin instance. */
+  constructor(moduleName: string) {
+    super(moduleName);
+  }
+
+  // tslint:disable-next-line:no-any
+  applyPatch(moduleExports: any, tracer: types.Tracer, version: string) {
```
kjin: Would `moduleExports: typeof mongodb` work?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+    shimmer.massUnwrap(this.moduleExports.Server.prototype, this.SERVER_FNS);
+    shimmer.massUnwrap(
+        this.moduleExports.Cursor.prototype, this.CURSOR_FNS_FIRST);
+  }
+
+  patchCommand() {
+    const plugin = this;
+    return (original: types.Func<mongodb.Server>) => {
+      // tslint:disable-next-line:no-any
+      return function(this: mongodb.Server, ns: string, command: any):
```
kjin: Can you add the type of `command`?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+
+  /** Unpatches all HTTP patched function. */
+  applyUnpatch(): void {
+    shimmer.unwrap(this.moduleExports.Server.prototype, 'command');
+    shimmer.massUnwrap(this.moduleExports.Server.prototype, this.SERVER_FNS);
+    shimmer.massUnwrap(
+        this.moduleExports.Cursor.prototype, this.CURSOR_FNS_FIRST);
+  }
+
+  patchCommand() {
```
kjin: Naming -- `getPatchCommandFunction`?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+    const plugin = this;
+    return (original: types.Func<mongodb.Server>) => {
+      // tslint:disable-next-line:no-any
+      return function(this: mongodb.Server, ns: string, command: any):
+          mongodb.Server {
+            let span: types.Span;
+            const index = arguments.length - 1;
```
kjin: Is it possible to use spread args instead of `arguments`?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -0,0 +1,162 @@
...
+
+            if (plugin.tracer.currentRootSpan && arguments.length > 0 &&
+                typeof resultHandler === 'function') {
+              let type: string;
+              if (command.createIndexes) {
+                type = 'createIndexes';
+              } else if (command.findandmodify) {
+                type = 'findAndModify';
+              } else if (command.ismaster) {
+                type = 'ismaster';
```
kjin: Make this camelcase for consistency (or make others not camelcase)

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+  assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 1);
+  assert.strictEqual(rootSpanVerifier.endedRootSpans[0].spans.length, 1);
+  assert.strictEqual(
+      rootSpanVerifier.endedRootSpans[0].spans[0].name, expectedName);
+  assert.strictEqual(
+      rootSpanVerifier.endedRootSpans[0].spans[0].type, expectedType);
+}
+
+describe('MongoDBPlugin', () => {
+  // For these tests, mongo must be runing. Add OPENCENSUS_MONGODB_TESTS to run
```
kjin: Can you update the CircleCI file to start mongo? These tests are currently getting skipped.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+  // these tests.
+  const OPENCENSUS_NETWORK_TESTS =
+      process.env.OPENCENSUS_NETWORK_TESTS as string;
+  let shouldTest = true;
+  if (!OPENCENSUS_NETWORK_TESTS) {
+    console.log('Skipping test-mongodb. Run MongoDB to test');
+    shouldTest = false;
+  }
+
+  const url = 'mongodb://localhost:27017';
```
kjin: Since they are constants, can all of these be capitalized?

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+    plugin.applyPatch(mongodb, tracer, VERSION);
+    accessCollection(url, dbName, collectionName)
+        // tslint:disable-next-line:no-any
+        .then((result: any) => {
+          client = result.client;
+          collection = result.collection;
+          done();
+        })
+        .catch((err: Error) => {
+          console.log(
```
kjin: Shouldn't an error condition fail this test rather than allowing it to skip?

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+  let client: mongodb.MongoClient;
+  let collection: mongodb.Collection;
+
+  before((done) => {
+    tracer.start({samplingRate: 1});
+    tracer.registerSpanEventListener(rootSpanVerifier);
+    plugin.applyPatch(mongodb, tracer, VERSION);
+    accessCollection(url, dbName, collectionName)
+        // tslint:disable-next-line:no-any
+        .then((result: any) => {
```
kjin: Why is `result` of type `any`? Is it possible to infer it automatically?

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+    rootSpanVerifier.endedRootSpans = [];
+    // Non traced insertion of basic data to perform tests
+    const insertData = [{a: 1}, {a: 2}, {a: 3}];
+    collection.insertMany(insertData, (err, result) => {
+      done();
+    });
+  });
+
+  afterEach((done) => {
+    collection.remove({}, (err, result) => {
```
kjin: Nit: `collection.remove({}, done)`

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+  /** Should intercept query */
+  describe('Instrumenting query operations', () => {
+    it('should create a child span for insert', (done) => {
+      const insertData = [{a: 1}, {a: 2}, {a: 3}];
+
+      tracer.startRootSpan({name: 'insertRootSpan'}, (rootSpan) => {
+        collection.insertMany(insertData, (err, result) => {
+          assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 0);
+          rootSpan.end();
+          assert.equal(err, null);
```
kjin: Nit: `assert.ifError(err)`

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+
+import {plugin} from '../src/';
+import {MongoDBPlugin} from '../src/';
+
+type MongoDBAccess = {
+  client: mongodb.MongoClient,
+  collection: mongodb.Collection
+};
+
+class RootSpanVerifier implements types.SpanEventListener {
```
kjin: Can you comment on each class and function what they are doing?

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -0,0 +1,280 @@
...
+          assertSpan(
+              rootSpanVerifier, `${dbName}.$cmd.count`, 'db.mongodb.query');
+          done();
+        });
+      });
+    });
+  });
+
+  /** Should intercept command */
+  describe('Removing Instrumentation', () => {
```
kjin: Is it possible to make each test isolated? In other words, this test should still work (and be meaningful) if I replaced `describe` with `describe.only`.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: That was an old dependency that I let trought. My mistake.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: Done. Also made patching functions private as we discussed before.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: Yes, changed it. But a strange type error appears when wrapping. Had to cast to `never` as a workaround.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: It is passed as a plain object, that's why I had to use `any`.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: Currently, other plugins are not following this pattern. As we were discussing before, should we update the other ones too or keep it this way? Which one is best?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: I made a few changes to the arguments. Please, tell me if it's better another way.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: Done.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: Sure, updated.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: I made it skip so it doesn't block the other tests in case the user doesn't have mongo running. I feel like failing here implies that there is a problem with the plugin behavior rather than with the environment.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: It wasn't inferring, although I created a type `MongoDBAccess` for this. So I added a return type of `Promise<MongoDBAccess>` to `accessCollection()` and it was able to infer `result` type properly. Thanks for noticing.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: Changed, thanks.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: Nice, changed.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: Sure. Please let me know if they are not clear enough.

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: In the `beforeEach()` we insert sufficient data to run the tests independently of each other. Right now, it is possible to run with `describe.only()` or `describe.skip()` at any inner `describe` block.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -20,23 +20,28 @@ import {logger} from '@opencensus/opencensus-core';
...
+export type MongoDB = typeof mongodb;
 
-/** Http instrumentation plugin for Opencensus */
+/** MongoDB instrumentation plugin for Opencensus */
 export class MongoDBPlugin extends classes.BasePlugin {
-  readonly SERVER_FNS = ['insert', 'update', 'remove', 'auth'];
-  readonly CURSOR_FNS_FIRST = ['_find', '_getmore'];
-  readonly SPAN_MONGODB_QUERY_TYPE = 'db.mongodb.query';
-  logger: types.Logger;
+  private SERVER_FNS = ['insert', 'update', 'remove', 'auth'];
```
kjin: `private readonly` would be more accurate.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
kjin: I don't remember whether we already decided in the other PRs -- let's incrementally rename them starting here. The others can be renamed when it's convenient.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -45,109 +50,110 @@ export class MongoDBPlugin extends classes.BasePlugin {
...
-              arguments[index] = plugin.patchEnd(span, resultHandler);
-            }
-
-            return original.apply(this, arguments);
-          };
+      return function(
+                 // tslint:disable-next-line:no-any
+                 this: mongodb.Server, ns: string, command: any,
+                 preferences: object, resultHandler: Function): mongodb.Server {
+        if (plugin.tracer.currentRootSpan && arguments.length > 0 &&
```
kjin: This will only work if there is guaranteed to be exactly 4 arguments. My comment earlier was on whether we could do something like

```js
function(this, ...args: any[])
```

and using `args` in place of `arguments`. Not sure if syntactically that will work -- if it doesn't, I think the previous approach (`arguments.length - 1`) is better.

kjin:
Did you get a chance to enable MongoDB in CircleCI?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: Changed

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
eduardoemery: I agree. Changed

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
eduardoemery: Sure, done that.

eduardoemery:
Sorry for updating the CircleCI late. Just did that this morning.

PR #32, opened 2018-06-08
fix: set currentRootSpan before start and set parentSpanId at span constructor

fabiogomessilva:
Fix three minor issues.

### DIFF
--- a/packages/opencensus-core/src/trace/model/span-base.ts
+++ b/packages/opencensus-core/src/trace/model/span-base.ts
@@ -116,10 +116,10 @@ export abstract class SpanBase implements types.Span {
   /** Gives the TraceContext of the span. */
   get spanContext(): types.SpanContext {
     return {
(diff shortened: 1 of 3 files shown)


packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -118,7 +118,7 @@ export abstract class SpanBase implements types.Span {
     return {
       traceId: this.traceId.toString(),
       spanId: this.id.toString(),
-      parentSpanId: this.parentSpanId
+      options: 1  // always traced
```
kjin: Is the below cast still needed?

kjin: Is the cast below still needed?

fabiogomessilva: No, it is not. Removed!

PR #33, opened 2018-06-08
chore: start reporting coverage

kjin:
This PR starts reporting code coverage. If all goes well, coverage should be at https://codecov.io/gh/census-instrumentation/opencensus-node


kjin:
Closing. I needed to re-think this. If anyone wants to pick up where I left off, feel free.

PR #34, opened 2018-06-10
feat: add instrumentation-grpc package to opencensus-node

fabiogomessilva:
This PR adds instrumentation-grpc package to opencensus-node. Unit tests are included.

### DIFF
--- a/packages/opencensus-instrumentation-all/src/index.js
+++ b/packages/opencensus-instrumentation-all/src/index.js
@@ -15,4 +15,11 @@
  */
 "use strict";
 
-exports.DEFAULT_INSTRUMENTATION_MODULES = ['http', 'https'];
+exports.DEFAULT_INSTRUMENTATION_MODULES = [
+    ,'http' 
+    ,'https'
+    ,'http2'
+    ,'grpc'
+    ,'mongodb'
+];
+
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -213,8 +213,7 @@ export class GrpcPlugin extends classes.BasePlugin {
   private clientStreamAndUnaryHandler<RequestType, ResponseType>(
       plugin: GrpcPlugin, rootSpan: types.RootSpan, call: ServerCallWithMeta,
       callback: SendUnaryDataCallback,
-      original: grpcTypes.handleCall<RequestType, ResponseType>,
-      self: grpcTypes.handleCall<RequestType, ResponseType>) {
+      original: grpcTypes.handleCall<RequestType, ResponseType>, self: {}) {
     function patchedCallback(
         err: grpcTypes.ServiceError,
         // tslint:disable-next-line:no-any
@@ -246,8 +245,7 @@ export class GrpcPlugin extends classes.BasePlugin {
    */
   private serverStreamAndBidiHandler<RequestType, ResponseType>(
       plugin: GrpcPlugin, rootSpan: types.RootSpan, call: ServerCallWithMeta,
-      original: grpcTypes.handleCall<RequestType, ResponseType>,
-      self: grpcTypes.handleCall<RequestType, ResponseType>) {
+      original: grpcTypes.handleCall<RequestType, ResponseType>, self: {}) {
     let spanEnded = false;
     const endSpan = () => {
       if (!spanEnded) {
@@ -486,7 +484,9 @@ export class GrpcPlugin extends classes.BasePlugin {
   }
 
   /**
+   * Convert Grpc Status to Span Status
    *
+   * At this current version of Opencensus Specs they are the same.
    */
   static convertGrpcStatusToSpanStatus(statusCode: grpcTypes.status): number {
     return statusCode;
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -1,0 +1,478 @@
+/**
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      gRPC://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {classes, logger, types} from '@opencensus/opencensus-core';
+import {EventEmitter} from 'events';
+import * as grpcModule from 'grpc';
+import * as lodash from 'lodash';
+import * as path from 'path';
+import * as semver from 'semver';
+import * as shimmer from 'shimmer';
+
+
+const findIndex = lodash.findIndex;
+
+function requireModule(module: string, nodeName: string) {
+  const indexPath = path.dirname(require.resolve(module));
+  return require(path.join(indexPath, nodeName));
+}
+const grpcClient = requireModule('grpc', 'src/client.js');
+
+
+//
+// Types definitions
+//
+export type GrpcModule = typeof grpcModule;
+type MakeClientConstructor = typeof grpcModule.makeGenericClientConstructor;
+type RegisterMethod = typeof grpcModule.Server.prototype.register;
+
+type Status = {
+  code: number, details: string; metadata: grpcModule.Metadata;
+};
+
+type ServerCall = typeof grpcModule.ServerUnaryCall|
+                  typeof grpcModule.ServerReadableStream|
+                  typeof grpcModule.ServerWriteableStream|
+                  typeof grpcModule.ServerDuplexStream;
+
+type ServerCallWithMeta = ServerCall&{
+  metadata: grpcModule.Metadata;
+  status: Status
+}
+&EventEmitter;
+
+export type SendUnaryDataCallback =
+    (error: grpcModule.ServiceError,
+     // tslint:disable-next-line:no-any
+     value?: any, trailer?: grpcModule.Metadata,
+     flags?: grpcModule.writeFlags) => void;
+
+
+type HandlerSet = {
+  // tslint:disable-next-line:no-any
+  func: grpcModule.handleCall<any, any>;
+  // tslint:disable-next-line:no-any
+  serialize: grpcModule.serialize<any>;
+  // tslint:disable-next-line:no-any
+  deserialize: grpcModule.deserialize<any>;
+  type: string;
+};
+
+
+/** gRPC instrumentation plugin for Opencensus */
+export class GrpcPlugin extends classes.BasePlugin {
+  /**
+   * Span grpc attributes
+   */
+  static readonly ATTRIBUTE_GRPC_KIND = 'grpc.kind';  // SERVER or CLIENT
+  static readonly ATTRIBUTE_GRPC_METHOD = 'grpc.method';
+  static readonly ATTRIBUTE_GRPC_STATUS_CODE = 'grpc.status_code';
+  static readonly ATTRIBUTE_GRPC_ERROR_NAME = 'grpc.error_name';
+  static readonly ATTRIBUTE_GRPC_ERROR_MESSAGE = 'grpc.error_message';
+
+  private logger: types.Logger;
+
+  /** Constructs a new GrpcPlugin instance. */
+  constructor() {
+    super('grpc');
+  }
+
+  /**
+   * Patches gRPC incoming and outcoming request functions.
+   * @param moduleExports The gRPC module exports
+   * @param tracer A tracer instance to create spans on.
+   * @param version The package version.
+   */
+  applyPatch(moduleExports: GrpcModule, tracer: types.Tracer, version: string) {
+    this.setPluginContext(moduleExports, tracer, version);
+    this.logger = tracer.logger || logger.logger('debug');
+
+    this.logger.debug('applying pacth to %s@%s', this.moduleName, this.version);
+
+    shimmer.wrap(
+        moduleExports.Server.prototype, 'register', this.patchServer());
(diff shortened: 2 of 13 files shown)


packages/opencensus-instrumentation-grpc/tsconfig.json
```diff
@@ -0,0 +1,21 @@
+
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "noImplicitThis": false,
```
kjin: Can you turn on this option?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+  applyUnpatch(): void {
+    shimmer.unwrap(this.moduleExports.Server.prototype, 'register');
+    shimmer.unwrap(grpcClient, 'makeClientConstructor');
+  }
+
+
+  /**
+   * Returns a function that patches the gRPC server register function in order
+   * to create trace spans for gRPC service methods.
+   * @returns {Function}  The register patched function.
```
kjin: `{Function}` annotation not needed here.

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+   */
+  private patchServer() {
```
kjin: What is the return type of this function?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+    return (originalRegister: RegisterMethod) => {
+      const plugin = this;
+      plugin.logger.debug('pathcServer');
+      return function register(
```
kjin: If you specify this as `register<RequestType, ResponseType>` then all of the function arguments can have non-any type annotations. (Also true for `clientStreamAndUnaryHandler`, etc.)

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+          name: `grpc.${original.path.replace('/', '')}`,
+          type: 'CLIENT',
+        };
+        const args = Array.prototype.slice.call(arguments);
+        // Checks if this remote function call is part of an operation by
+        // checking if there is a current root span, if so, we create a child
+        // span. In case there is no root span, this means that the remote
+        // function call is the first operation, therefore we create a root
+        // span.
+        if (!plugin.tracer.currentRootSpan) {
```
kjin: I am starting to not like having to special-case creating a root span for any client if there isn't already a root span. This complicates things for future plugin writers, as this makes it unclear what the role of a root span is, and _every_ RPC plugin will have to implement it this way. In addition, it masks async context propagation bugs.

I'd like to propose instead that we only start child spans in RPC calls, and add a config option `{ allowTopLevelChildSpans: boolean }` that users can turn on to achieve similar behavior. What do you think?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+      };
+    };
+  }
+
+  /**
+   * This method handels the client remote call
+   */
+  private makeGrpcClientRemoteCall(
+      // tslint:disable-next-line:no-any
+      original: any, args: any[], self: any, plugin: GrpcPlugin) {
```
kjin: Can you specify the actual types here?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+    };
+  }
+
+
+  /**
+   * This function starts a span (child or root) immediately before the
+   * client method is invoked, and ends it either in a callback or stream
+   * event handler, depending on the method type.
+   */
+  private patchAllClientsMethods() {
```
kjin: This is a misnomer -- `patchClientMethod` would be more accurate. (Or, `createClientPatchMethod`, following convention that I've stated preference for in other PRs.)

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+      span.addAttribute(GrpcPlugin.ATTRIBUTE_GRPC_METHOD, original.path);
+      span.addAttribute(GrpcPlugin.ATTRIBUTE_GRPC_KIND, span.type);
+
+      const call = original.apply(self, args);
+
+      plugin.tracer.wrapEmitter(call);
+
+      // if server stream or bidi
+      if (original.responseStream) {
+        let spanEnded = false;
```
kjin: Can you add a comment saying that `spanEnded` is needed because it's possible that both `error` and `status` events are emitted?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+import * as shimmer from 'shimmer';
+
+
+const findIndex = lodash.findIndex;
+
+function requireModule(module: string, nodeName: string) {
+  const indexPath = path.dirname(require.resolve(module));
+  return require(path.join(indexPath, nodeName));
+}
+const grpcClient = requireModule('grpc', 'src/client.js');
```
kjin: We should avoid a dependency on the module being patched.

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+    // A possible issue that could occur is if the 'options' parameter from
+    // the user contains an '_internal_repr' as well as a 'getMap' function,
+    // but this is an extremely rare case.
+    // tslint:disable-next-line:no-any
+    let metadataIndex = findIndex(args, (arg: any) => {
+      return arg && typeof arg === 'object' && arg._internal_repr &&
+          typeof arg.getMap === 'function';
+    });
+    if (metadataIndex === -1) {
+      metadata = new grpcModule.Metadata();
```
kjin: We should avoid using `grpcModule` for constructing an actual `Metadata` object, if possible.

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -0,0 +1,478 @@
...
+        metadataIndex = 0;
+      }
+      args.splice(metadataIndex, 0, metadata);
+    } else {
+      metadata = args[metadataIndex];
+    }
+    return metadata;
+  }
+
+  traceStatus(statusCode: grpcModule.status): number {
```
kjin: What is the purpose of this function? If it's a utility function, can it be `private static`?

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
```diff
@@ -0,0 +1,493 @@
...
+  endedRootSpans: types.RootSpan[] = [];
+
+  onStartSpan(span: types.RootSpan): void {}
+  onEndSpan(root: types.RootSpan) {
+    this.endedRootSpans.push(root);
+  }
+}
+
+
+describe('GrppcPlugin() ', () => {
```
kjin: `GrpcPlugin`

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
```diff
@@ -0,0 +1,493 @@
...
+    assert.strictEqual(
+        span.attributes[GrpcPlugin.ATTRIBUTE_GRPC_STATUS_CODE], `${status}`);
+    assert.ok(span.attributes[GrpcPlugin.ATTRIBUTE_GRPC_METHOD]);
+    if (status !== grpcModule.status.OK) {
+      assert.ok(span.attributes[GrpcPlugin.ATTRIBUTE_GRPC_ERROR_NAME]);
+      assert.ok(span.attributes[GrpcPlugin.ATTRIBUTE_GRPC_ERROR_MESSAGE]);
+    }
+  }
+
+  function assertPropagation(sourceSpan: types.Span, targetSpan: types.Span) {
```
kjin: Can you comment on what this function is asserting? It is hard to tell at a glance which is the server or client.

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
```diff
@@ -0,0 +1,493 @@
...
+  bidiStreamMethod: any;
+};
+
+type TestRequestResponse = {
+  num: number;
+};
+
+const grpcClient = {
+  // tslint:disable-next-line:no-any
+  unaryMethod: (client: TestGrpcClient, request: TestRequestResponse): any => {
```
kjin: Don't these return `Promise<{num: number}>`, etc.?

packages/opencensus-instrumentation-grpc/tsconfig.json
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: Comment improved!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: Thanks. Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: patchServer -> getPatchServer
patchClient -> getPatchClient
patchAllClientsMethods -> getPatchedClientMethods

packages/opencensus-instrumentation-grpc/src/grpc.ts
kjin: `self: Client`
I think `original` can be Function `args` can be any for now, diving into types here doesn't help us.

fabiogomessilva: Thanks. Done!

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: PR #52

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: PR #52

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
fabiogomessilva: TestRequestResponse or TestRequestResponse[]
Done.

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: This function was renamed to 'convertGrpcStatusToSpanStatus'

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -231,12 +244,10 @@ export class GrpcPlugin extends classes.BasePlugin {
...
    */
-  private serverStreamAndBidiHandler(
+  private serverStreamAndBidiHandler<RequestType, ResponseType>(
       plugin: GrpcPlugin, rootSpan: types.RootSpan, call: ServerCallWithMeta,
-      // tslint:disable-next-line:no-any
-      original: grpcModule.handleCall<any, any>,
-      // tslint:disable-next-line:no-any
-      self: grpcModule.handleCall<any, any>) {
+      original: grpcTypes.handleCall<RequestType, ResponseType>,
+      self: grpcTypes.handleCall<RequestType, ResponseType>) {
```
kjin: I think `self: {}` would probably be sufficient here. It doesn't make sense for `original` and `self` to be the same type.

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -469,7 +485,10 @@ export class GrpcPlugin extends classes.BasePlugin {
     return metadata;
   }
 
-  traceStatus(statusCode: grpcModule.status): number {
+  /**
+   *
```
kjin: Can you fill in the docs here?

packages/opencensus-instrumentation-all/src/index.js
```diff
@@ -15,4 +15,11 @@
  */
 "use strict";
 
-exports.DEFAULT_INSTRUMENTATION_MODULES = ['http', 'https'];
+exports.DEFAULT_INSTRUMENTATION_MODULES = [
```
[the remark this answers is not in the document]
fabiogomessilva: I thought about that at the time, but it seems to be too much overhead for just dependencies at package.json and export a constant.

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-grpc/src/grpc.ts
fabiogomessilva: Done!

PR #36, opened 2018-06-12
refactor: address instrumentation-http TODOs

fabiogomessilva:
This PR address instrumentation-http TODOs

### DIFF
--- a/packages/opencensus-instrumentation-http/src/http.ts
+++ b/packages/opencensus-instrumentation-http/src/http.ts
@@ -108,7 +108,7 @@ export class HttpPlugin extends classes.BasePlugin {
   /**
    * Creates spans for incoming requests, restoring spans' context if applied.
    */
-  patchIncomingRequest() {
+  protected patchIncomingRequest() {
     return (original: RequestFunction) => {
       const plugin = this;
       return function incomingRequest(
@@ -194,7 +194,7 @@ export class HttpPlugin extends classes.BasePlugin {
    * Creates spans for outgoing requests, sending spans' context for distributed
    * tracing.
    */
-  patchOutgoingRequest() {
+  protected patchOutgoingRequest() {
     return (original: types.Func<httpModule.ClientRequest>):
                types.Func<httpModule.ClientRequest> => {
       const plugin = this;
@@ -254,7 +254,7 @@ export class HttpPlugin extends classes.BasePlugin {
    * @param original The original patched function.
    * @param options The arguments to the original function.
    */
-  makeRequestTrace(
+  private makeRequestTrace(
       // tslint:disable-next-line:no-any
       request: httpModule.ClientRequest, options: httpModule.RequestOptions,
       plugin: HttpPlugin): types.Func<httpModule.ClientRequest> {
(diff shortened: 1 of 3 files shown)


packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -290,15 +285,15 @@ export class HttpPlugin extends classes.BasePlugin {
         response.on('end', () => {
           plugin.logger.debug('outgoingRequest on end()');
           const method = response.method ? response.method : 'GET';
-          const reqUrl = url.parse(options);
           const userAgent =
-              headers ? (headers['user-agent'] || headers['User-Agent']) : null;
+              headers ? (headers['user-agent'] || headers['User-Agent']) : '';
```
kjin: If there is no user agent... will it append an empty label? We should make the label conditional on there being a user agent.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -24,7 +24,6 @@ import * as url from 'url';
 import * as uuid from 'uuid';
 
 
-// TODO: maybe we should have a setup as a Client or as Server.
```
kjin: Has this TODO been addressed? I don't see any change that reflects that it has.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -80,7 +78,7 @@ export class HttpPlugin extends classes.BasePlugin {
       shimmer.wrap(
           moduleExports && moduleExports.Server &&
               moduleExports.Server.prototype,
-          'emit', this.patchIncomingRequest());
+          'emit' as never, this.patchIncomingRequest());
```
kjin: Why this change?

kjin:
While you're on this file, can you also make `patch*` functions private?

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Yes, there is no need to have specific setups for client and servers.  No action to do.

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: This change is a workaround to solve a strange type cast error for this.patchIncomingRequest when we set the type for moduleExports as HttpModule.

fabiogomessilva: Regarding method visibility, we can not set patch* as private because subclass https needs access to those methods - they were set to protected instead.

packages/opencensus-instrumentation-http/src/http.ts
fabiogomessilva: Done!

packages/opencensus-instrumentation-http/src/http.ts
kjin: Would this work?

```js
shimmer.wrap(moduleExports.Server.prototype, 'emit', /*...*/);
```

fabiogomessilva: Could not make it work.

kjin: Ok. Might be a problem with the shimmer type definitions.

PR #46, opened 2018-06-15
refactor: address instrumentation-https TODOs

eduardoemery:


packages/opencensus-instrumentation-https/test/test-https.ts
```diff
@@ -42,7 +43,10 @@ function doNock(
 const httpRequest = {
   get: (options: {}|string) => {
     return new Promise((resolve, reject) => {
-      return https.get(options, resp => {
+      if (typeof (options) === 'string') {
```
kjin: We should test both `request` and `get`, rather than just one. In Node 6, you may get unexpected results: https://github.com/nodejs/node/blob/master/lib/https.js#L288 `get` may not be patched if only `request` is patched.

eduardoemery: Yes, you are right. I made this change and added tests for both `get()` and `request()` methods. Thanks for pointing it out.

packages/opencensus-instrumentation-https/test/test-https.ts
```diff
@@ -19,16 +19,16 @@ import {classes} from '@opencensus/opencensus-core';
...
 import * as https from 'https';
 import * as mocha from 'mocha';
 import * as nock from 'nock';
 import * as shimmer from 'shimmer';
+import * as url from 'url';
 
 import {plugin} from '../src/';
 import {HttpsPlugin} from '../src/';
 
-// TODO: evaluate how to reuse test code from http plugin
```
eduardoemery: Tests are fairly different.

packages/opencensus-instrumentation-https/test/test-https.ts
```diff
@@ -267,35 +286,31 @@ describe('HttpsPlugin', () => {
...
-        assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 1);
+        assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 2);
         const span = rootSpanVerifier.endedRootSpans[0];
         assertSpanAttributes(
             span, 200, 'GET', 'localhost', testPath, 'Android');
       });
     });
   });
 
-  // TODO: This tests relies on a specific order in which tests are executed.
```
eduardoemery: Changed incoming requests test to not need unpatching.

packages/opencensus-instrumentation-https/test/test-https.ts
```diff
@@ -54,8 +61,10 @@ const httpRequest = {
           reject(err);
         });
       });
+      req.end();
+      return req;
```
kjin: Return value will never be read... better to not have it at all.

packages/opencensus-instrumentation-https/test/test-https.ts
kjin: Ok, but why are they different? Don't `http` and `https` have nearly identical APIs?

eduardoemery: In fact, `https` has only a small wrapper to `http` outgoing requests, and gets all the rest from `http`. We would only need to guarantee that `http` works properly along side the `https`' `request()`/`get()` wrapper.

Currently `https` uses `http` attributes defined by the [OpenCensus specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md), but we believe that soon enough it will have its own set of attributes, and tests for those attributes and behavior will be made. If we reuse code here, it may result in future rework for a member/contributor that may want to update `https` of even change the approach to one that doesn't depend on `http`.

kjin: > Currently `https` uses `http` attributes defined by the [OpenCensus specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md), but we believe that soon enough it will have its own set of attributes, and tests for those attributes and behavior will be made. If we reuse code here, it may result in future rework for a member/contributor that may want to update `https` of even change the approach to one that doesn't depend on `http`.

Correct me if I'm wrong, but it sounds like you are expecting that there will be separate attributes for `https` in the future, such as `https.user_agent`. I personally would not anticipate this to happen... since for instrumentation purposes HTTPS is not really different from HTTP at all.

Because they exist as tests for separate packages, I'll admit there is no clear way to share code between the `instrumentation-http` and `instrumentation-https` tests that I can think of. But I don't think they should be different at all, except for passing options that are specific to HTTPS (like the tls option), because they otherwise have identical APIs.

Not asking for any changes here... just want to make sure we are on the same page.

eduardoemery: You understood me, I was just supposing. Regarding the code sharing, I think making sure `http` worked would be sufficient, maybe, gotta think more about this. Anyways, great to know we are on the same page :D

PR #47, opened 2018-06-15
refactor: address opencensus-core TODOs

eduardoemery:


packages/opencensus-core/src/common/console-logger.ts
```diff
@@ -27,9 +27,6 @@ export class ConsoleLogger implements types.Logger {
   static LEVELS = ['silent', 'error', 'warn', 'info', 'debug', 'silly'];
   level: string;
 
-
-  // TODO: reevaluate options to accept numbers as a parameter
```
kjin: For each TODO that is removed without changes, please document (as a PR comment) why you decided not to do it.

packages/opencensus-core/src/exporters/types.ts
```diff
@@ -25,7 +25,8 @@ export interface Exporter extends modelTypes.SpanEventListener {
    * Sends a list of root spans to the service.
    * @param rootSpans A list of root spans to publish.
    */
-  publish(rootSpans: modelTypes.RootSpan[]): void;
+  // tslint:disable-next-line:no-any
+  publish(rootSpans: modelTypes.RootSpan[]): Promise<any>|void;
```
kjin: Two notes:
1. What is the meaning of the deferred value being returned here? How are users expected to treat this value? If there are no good answers to this question it should just be `Promise<void>`.
2. Returning either a Promise or `void` complicates handling the return value. Can we only return `Promise`?

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -54,16 +54,16 @@ export abstract class BasePlugin implements types.Plugin {
...
-  }
+  /**
+   * Patches module exports functions.
+   * @param moduleExports nodejs module exports to be patched
+   * @param tracer a tracer instance to create spans from
+   * @param version module version description
+   */
+  abstract applyPatch(
+      // tslint:disable:no-any
+      moduleExports: any, tracer: modelTypes.Tracer, version: string): any;
```
kjin: While we are here... maybe it makes sense to parameterize the return value. We know the `moduleExports` and the return value should be the same type, so we can make it `applyPatch<T>(moduleExports: T, ...): T;`.

packages/opencensus-core/package.json
```diff
@@ -48,14 +47,13 @@
     "gts": "^0.5.4",
     "mocha": "^5.0.4",
     "ncp": "^2.0.0",
+    "nyc": "11.6.0",
     "ts-node": "^4.0.0",
-    "typescript": "~2.6.1",
-    "nyc": "11.6.0"
+    "intercept-stdout": "^0.1.2",
```
kjin: Can you preserve the alphabetical ordering here?

packages/opencensus-core/src/exporters/types.ts
eduardoemery: The ideia is that by returning `Promise<any> | void` we account for `publish` methods that returns a Promise, while giving freedom to future collaborators to make a method that returns `void`. Currently, `NoopExporter` and `ConsoleExporter` return void, but if you prefer, I can change them to return `Promise<void>` and restrict the interface, no problems with that IMO.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
eduardoemery: Makes a lot of sense. As `oc-core` changed since this PR was opened, I made this change to the `enable` method.

packages/opencensus-core/package.json
eduardoemery: Sure.

packages/opencensus-core/src/common/console-logger.ts
eduardoemery: Ok, I'll comment those. This one in specific was removed because being able to use numbers as parameters gives more freedom to the end user. Also, passing number as parameters is being currently used across the implementation.

packages/opencensus-core/src/common/console-logger.ts
```diff
@@ -104,10 +101,6 @@ export class ConsoleLogger implements types.Logger {
   }
 }
 
-
-// TODO: reevaluate the need to create a new logger instance on every call to
```
eduardoemery: By creating a new logger instance on every call to it we allow distinct classes to have distinct log levels.

packages/opencensus-core/src/internal/util.ts
```diff
@@ -19,9 +19,6 @@
 
 import * as crypto from 'crypto';
 
-
-// TODO: rethink this snippet aproach
```
eduardoemery: There was no need for change here.

packages/opencensus-core/src/common/console-logger.ts
kjin: I think the original comment that this was in reference to that every time `logger` was called, the field `logger['logger']` was being replaced. This field is also never getting read. The `logger` method is otherwise just a call-through to `new ConsoleLogger`.

I would prefer that `logger` not even be used in the first place, and that anywhere it's used, `new ConsoleLogger` is used instead. After all, why do we need the wrapper method?

kjin: See: https://github.com/census-instrumentation/opencensus-node/pull/2#discussion_r185030617

packages/opencensus-core/src/exporters/types.ts
kjin: It's OK for `publish` to inherently not contain anything asynchronous, but it would still be more consistent to return a `Promise.resolve()` instead of `void` anyway. If we consider the code that will call this function:
```js
const result = exporter.publish();
if (result) {
  result.then(next);
} else {
  next(); // note: zalgo alert
}
```
Why even have this branching logic, when we could just do
```js
publish().then(next);
```

That was the original context behind the second question. If you could make the change in `*Exporter` that would be great.

packages/opencensus-core/src/common/console-logger.ts
kjin: Makes sense, thanks.

packages/opencensus-core/src/exporters/types.ts
eduardoemery: Hmm, got it. Made the change but had to make `Promise<number|string|void>` to account for some exporter implementations that resolve `number` or `string`.

packages/opencensus-core/src/common/console-logger.ts
eduardoemery: My mistake, I got the TODO confused. Regarding this one, our idea here is to return an instance of logger, although calling `new ConsoleLogger()` works too. We drove inspiration from [this logDriver](https://github.com/cainus/logdriver/blob/bba1761737ca72f04d6b445629848538d038484a/index.js#L50).

I removed the `logger['logger'] = aLogger;` that had no need to be there at all.

eduardoemery:
Sorry for the delay. Took a look at your comments, thanks for reviewing. Made a few changes.

packages/opencensus-core/src/exporters/types.ts
kjin: `Promise<number|string|void>` goes back to my original first question. Do we use the return value in a meaningful way? If so, it would be better to standardize now what that return value is (maybe a boolean or status code). Otherwise, if we don't use it, we should just make it `void`.

BTW, the answer to this question might not exist today. It's hard to know how to standardize the return value if we are currently not using it. So I can merge this PR, just something to think about.

kjin:
@eduardoemery Maybe you're in the middle of it right now, but currently the review comment changes only include `package-lock.json`. Did you miss adding a file to the commit?

packages/opencensus-core/src/exporters/types.ts
eduardoemery: I could see the return value being used in tests only, but not in applications. I'm not sure if we should standardize it tho, goes back to that thing about future contributors. But I understand your point of view. Would be great to discuss that latter.

eduardoemery:
Yesterday morning I had some trouble with my git. Got it fixed but apparently it impacted here. Fixed now, sorry for the trouble caused.

packages/opencensus-core/src/common/console-logger.ts
```diff
@@ -104,19 +101,14 @@ export class ConsoleLogger implements types.Logger {
...
+ * Function logger exported to others classes. Inspired by:
+ * https://github.com/cainus/logdriver/blob/bba1761737ca72f04d6b445629848538d038484a/index.js#L50
  * @param options A logger options or strig to logger in console
  */
 // tslint:disable-next-line:no-any
 const logger: any = (options?: types.LoggerOptions|string|number) => {
-  const aLogger = new ConsoleLogger(options);
-  logger['logger'] = aLogger;
-  return aLogger;
+  return new ConsoleLogger(options);
```
kjin: Nit: Make this function have return type `ConsoleLogger`

eduardoemery: Sure.

kjin: `types.Logger` is better. Thanks!

PR #48, opened 2018-06-15
refactor: address b3 propagation TODOs

eduardoemery:
The TODO left was: 

> TODO: Review this logic, maybe make it more robust. Question: are this logic valid if any of the `getHeader` operations returns a `string[]`.

I checked and `string[]` has the same behavior in this logic. `getHeader()` returns a `string | string[]`, the verification `isNaN(Number())` works for both.


eduardoemery:
Added a few lines to make sure we get the first header if it comes as an Array.

PR #50, opened 2018-06-15
refactor: address opencensus-nodejs TODOs

eduardoemery:


packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -175,6 +175,17 @@ export class Tracer implements types.Tracer {
     this.eventListenersLocal.push(listner);
   }
 
+  /**
+   * Unregisters an end span event listener.
+   * @param listener The listener to unregister.
+   */
+  unregisterSpanEventListener(listner: types.SpanEventListener) {
```
kjin: Nit: `listener` (and in other files as well).

packages/opencensus-core/src/trace/types.ts
```diff
@@ -28,10 +28,13 @@ export interface Tracing {
...
   readonly exporter: exportersTypes.Exporter;
 
+  /** Gets active status  */
+  active: boolean;
+
   /**
-   * Enables the tracing process.
-   * @param userConfig A configuration object.
-   * @returns The tracing object.
+   * Starts the tracing.
```
kjin: s/`tracing`/`Tracing instance`/g

packages/opencensus-core/src/trace/model/tracer.ts
eduardoemery: Thanks for noticing.

packages/opencensus-core/src/trace/types.ts
eduardoemery: Ok. Updated here and in the implementation.

PR #51, opened 2018-06-15
feat: add mechanism to load internal module files to patch

fabiogomessilva:
Sometimes, there is a need to patch internal module files that are not exported by default. This PR aims to add a mechanism to make it possible to load internal files from a module and make it available to a plugin.


packages/opencensus-core/src/trace/config/types.ts
```diff
@@ -59,13 +59,31 @@ export interface TracingConfig {
...
+   *
+   * Ex.: (internal files for grpc plugin that need patch)
+   *
+   *  pluginsExtra: {
+   *      'grpc':{'client':'src/client.js','metadata':'src/metadata.js'
+   *       ...
+   * }
+   *
+   */
+  pluginsExtra?: PluginExtraFiles2Patch;
```
kjin: This is an implementation detail for each instrumentation module. I think we shouldn't add another configuration option for this reason.

Is it possible to make patching internal files a feature of the `BasePlugin`/`PluginLoader` classes?

PR #52, opened 2018-06-16
feat: add mechanism to load internal module files to patch

fabiogomessilva:
Sometimes, there is a need to patch internal module files that are not exported by default. This PR aims to add a mechanism to make it possible to load internal files from a module and make it available to a plugin.

### DIFF
--- a/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
+++ b/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
@@ -22,8 +22,14 @@ import * as modelTypes from '../model/types';
 
 import * as types from './types';
 
+/**
+ * Maps a name (key) representing a internal file module and its exports
+ */
+export type ModuleExportsMapping = {
+  // tslint:disable:no-any
+  [key: string]: any;
+};
 
-// TODO: improve Jsdoc comments
 
 /** This class represent the base to patch plugin. */
 export abstract class BasePlugin implements types.Plugin {
@@ -39,9 +45,9 @@ export abstract class BasePlugin implements types.Plugin {
   /** a logger */
   protected logger: Logger;
   /** list of internal files that need patch and are not exported by default */
-  protected internalFileList: types.PluginInternalFiles;
+  protected readonly internalFileList: types.PluginInternalFiles;
   /**  internal files loaded */
-  protected internalFilesExports: types.ModuleExports;
+  protected internalFilesExports: ModuleExportsMapping;
   /** module directory - used to load internal files */
   protected basedir: string;
 
@@ -58,43 +64,63 @@ export abstract class BasePlugin implements types.Plugin {
    * @param moduleExports nodejs module exports to set as context
    * @param tracer tracer relating to context
    * @param version module version description
+   * @param basedir module absolute path
    */
   // tslint:disable:no-any
-  protected setPluginContext(
+  private setPluginContext(
       moduleExports: any, tracer: modelTypes.Tracer, version: string,
       basedir?: string) {
     this.moduleExports = moduleExports;
     this.tracer = tracer;
     this.version = version;
     this.basedir = basedir;
-    this.logger = tracer.logger || logger.logger();
+    this.logger = tracer.logger;
     this.internalFilesExports = this.loadInternalFiles();
   }
 
 
-  // TODO: review this implementation
-  // From the perspective of an instrumentation module author,
-  // that applyUnpatch is abstract makes it seem like patching is optional,
-  // while unpatching is not. It should be the other way around
-
-  // tslint:disable:no-any
-  applyPatch(
+  /**
+   * Method to apply the instrumentation patch.
+   *
+   * GoF Template Method Pattern - this is the invariant part of the patthern
+   *
+   * @param moduleExports nodejs module exports from the module to patch
+   * @param tracer a tracer instance
+   * @param version version of the current instaled module to patch
+   * @param basedir module absolute path
+   */
+  applyPluginPatch(
+      // tslint:disable:no-any
       moduleExports: any, tracer: modelTypes.Tracer, version: string,
-      basedir?: string): any {
+      basedir: string) {
     this.setPluginContext(moduleExports, tracer, version, basedir);
+    return this.applyPatch();
   }
 
-  abstract applyUnpatch(): void;
+
+  applyPluginUnPatch() {
+    this.applyUnpatch();
+  }
+
+  /**
+   * applyPatch() and applyUnpatch()
+   *
+   * GoF Template Method Pattern - this is the variant part of the pattern
+   * Each plugin should implement his own version
+   */
+  // tslint:disable:no-any
+  protected abstract applyPatch(): any;
+  protected abstract applyUnpatch(): void;
 
 
   /**
    * Load internal files according to version range
    */
-  private loadInternalFiles(): types.ModuleExports {
-    let result: types.ModuleExports = null;
-    this.logger.debug('loadInternalFiles %o', this.internalFileList);
+  private loadInternalFiles(): ModuleExportsMapping {
+    let result: ModuleExportsMapping = null;
     if (this.internalFileList) {
-      Object.keys(this.internalFileList).map(versionRange => {
+      this.logger.debug('loadInternalFiles %o', this.internalFileList);
+      Object.keys(this.internalFileList).forEach(versionRange => {
         if (semver.satisfies(this.version, versionRange)) {
           result = this.loadInternalModuleFiles(
               this.internalFileList[versionRange], this.basedir);
@@ -115,18 +141,18 @@ export abstract class BasePlugin implements types.Plugin {
    */
   private loadInternalModuleFiles(
       extraModulesList: types.PluginNames,
-      basedir: string): types.ModuleExports {
-    const extraModules: types.ModuleExports = {};
+      basedir: string): ModuleExportsMapping {
+    const extraModules: ModuleExportsMapping = {};
     if (extraModulesList) {
-      Object.keys(extraModulesList).map(modulename => {
+      Object.keys(extraModulesList).forEach(moduleName => {
         try {
-          this.logger.debug('loading File %s', extraModulesList[modulename]);
-          extraModules[modulename] =
-              require(path.join(basedir, extraModulesList[modulename]));
+          this.logger.debug('loading File %s', extraModulesList[moduleName]);
+          extraModules[moduleName] =
+              require(path.join(basedir, extraModulesList[moduleName]));
         } catch (e) {
           this.logger.error(
               'Could not load internal file %s of module %s. Error: %s',
-              path.join(basedir, extraModulesList[modulename]), this.moduleName,
+              path.join(basedir, extraModulesList[moduleName]), this.moduleName,
               e.message);
         }
       });
--- a/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
+++ b/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
@@ -13,9 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import * as shimmer from 'shimmer';
+import * as path from 'path';
+import * as semver from 'semver';
+
+import {logger} from '../../common/console-logger';
+import {Logger} from '../../common/types';
 import * as modelTypes from '../model/types';
+
 import * as types from './types';
+
 
 // TODO: improve Jsdoc comments
 
@@ -30,6 +36,14 @@ export abstract class BasePlugin implements types.Plugin {
   protected tracer: modelTypes.Tracer;
   /** The module version. */
   protected version: string;
+  /** a logger */
+  protected logger: Logger;
+  /** list of internal files that need patch and are not exported by default */
+  protected internalFileList: types.PluginInternalFiles;
+  /**  internal files loaded */
+  protected internalFilesExports: types.ModuleExports;
+  /** module directory - used to load internal files */
+  protected basedir: string;
 
   /**
    * Constructs a new BasePlugin instance.
@@ -47,10 +61,14 @@ export abstract class BasePlugin implements types.Plugin {
    */
   // tslint:disable:no-any
   protected setPluginContext(
-      moduleExports: any, tracer: modelTypes.Tracer, version: string) {
+      moduleExports: any, tracer: modelTypes.Tracer, version: string,
+      basedir?: string) {
     this.moduleExports = moduleExports;
     this.tracer = tracer;
     this.version = version;
+    this.basedir = basedir;
+    this.logger = tracer.logger || logger.logger();
+    this.internalFilesExports = this.loadInternalFiles();
   }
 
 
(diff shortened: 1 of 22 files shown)


packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -60,10 +78,59 @@ export abstract class BasePlugin implements types.Plugin {
   // while unpatching is not. It should be the other way around
 
   // tslint:disable:no-any
-  applyPatch(moduleExports: any, tracer: modelTypes.Tracer, version: string):
-      any {
-    this.setPluginContext(moduleExports, tracer, version);
+  applyPatch(
+      moduleExports: any, tracer: modelTypes.Tracer, version: string,
+      basedir?: string): any {
```
kjin: `basedir` is optional here but is passed to functions that require it to have a value later. I would suggest making it required here as well.

packages/opencensus-core/src/trace/instrumentation/types.ts
```diff
@@ -39,3 +41,18 @@ export interface Plugin {
...
+ * a mapping of a property name to a internal plugin file name.
+ */
+export type PluginInternalFiles = {
+  [versions: string]: PluginNames;
+};
+
+/**
+ * Maps a name (key) representing a internal file module and its exports
+ */
+export type ModuleExports = {
```
kjin: The type name here should be more descriptive. Something like `ModuleExportsMapping` or something like that. But I wonder if you really need this type to be defined explicitly, because it's so broad. (Why not just define it locally in `base-plugin.ts`.)

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -60,10 +78,59 @@ export abstract class BasePlugin implements types.Plugin {
...
+  /**
+   * Load internal files according to version range
+   */
+  private loadInternalFiles(): types.ModuleExports {
+    let result: types.ModuleExports = null;
+    this.logger.debug('loadInternalFiles %o', this.internalFileList);
+    if (this.internalFileList) {
+      Object.keys(this.internalFileList).map(versionRange => {
+        if (semver.satisfies(this.version, versionRange)) {
+          result = this.loadInternalModuleFiles(
```
kjin: Should you warn if multiple versions match? It sounds like users would be depending on implementation-private behavior.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -60,10 +78,59 @@ export abstract class BasePlugin implements types.Plugin {
...
+
+
+  /**
+   * Load internal files according to version range
+   */
+  private loadInternalFiles(): types.ModuleExports {
+    let result: types.ModuleExports = null;
+    this.logger.debug('loadInternalFiles %o', this.internalFileList);
+    if (this.internalFileList) {
+      Object.keys(this.internalFileList).map(versionRange => {
```
kjin: Prefer `forEach` over `map` if you are not using the return value.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -30,6 +36,14 @@ export abstract class BasePlugin implements types.Plugin {
   protected tracer: modelTypes.Tracer;
   /** The module version. */
   protected version: string;
+  /** a logger */
+  protected logger: Logger;
+  /** list of internal files that need patch and are not exported by default */
+  protected internalFileList: types.PluginInternalFiles;
```
kjin: This is `readonly`, right?

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -60,10 +78,59 @@ export abstract class BasePlugin implements types.Plugin {
...
+
+  /**
+   * Load internal files from a module and  set internalFilesExports
+   */
+  private loadInternalModuleFiles(
+      extraModulesList: types.PluginNames,
+      basedir: string): types.ModuleExports {
+    const extraModules: types.ModuleExports = {};
+    if (extraModulesList) {
+      Object.keys(extraModulesList).map(modulename => {
```
kjin: Use a camelcase variable name (`modulename`).

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -60,10 +78,59 @@ export abstract class BasePlugin implements types.Plugin {
...
+        try {
+          this.logger.debug('loading File %s', extraModulesList[modulename]);
+          extraModules[modulename] =
+              require(path.join(basedir, extraModulesList[modulename]));
```
kjin: I am not certain that this will work in all cases. What does `basedir` usually look like -- is it absolute or relative?

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -47,10 +61,14 @@ export abstract class BasePlugin implements types.Plugin {
...
   // tslint:disable:no-any
   protected setPluginContext(
-      moduleExports: any, tracer: modelTypes.Tracer, version: string) {
+      moduleExports: any, tracer: modelTypes.Tracer, version: string,
+      basedir?: string) {
     this.moduleExports = moduleExports;
     this.tracer = tracer;
     this.version = version;
+    this.basedir = basedir;
+    this.logger = tracer.logger || logger.logger();
```
kjin: Why not just `tracer.logger`?

packages/opencensus-nodejs/test/instrumentation/node_modules/@opencensus/instrumentation-simple-module/simple-module.ts
```diff
@@ -20,18 +20,36 @@ import * as shimmer from 'shimmer';
 
 
 export class SimpleModulePlugin extends classes.BasePlugin {
+  protected internalFileList: types.PluginInternalFiles = {
```
kjin: This module doesn't seem very simple anymore. Can you have a separate instrumentation fixture for something that requires an internal module to be patched.

packages/opencensus-nodejs/test/test-plugin-loader.ts
```diff
@@ -98,6 +94,24 @@ describe('Plugin Loader', () => {
...
+        assert.strictEqual(pluginLoader.plugins.length, 1);
+        assert.strictEqual(simpleModule.name(), 'patched-' + modulename);
+        assert.strictEqual(simpleModule.value(), 101);
+
+        const extraModuleName = 'extra-module';
+        const indexPath = path.dirname(require.resolve(modulename));
+        const extraFile = 'src/extra-module';
+        const extraModule = require(path.join(indexPath, extraFile));
+        assert.strictEqual(extraModule.name(), 'patched-' + extraModuleName);
+        assert.strictEqual(extraModule.value(), 121);
```
kjin: `extraModule` is already not user-visible. I would rather see a test based around a module that uses the module exports of an internal file.

kjin:
Overall, this is much better than before. I think it's good to not expose details about module internals to users.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: It will not be used by all implementations, just the ones that need to patch internal files that are not exported. The parameter as optional does not affect current implementations.

packages/opencensus-core/src/trace/instrumentation/types.ts
fabiogomessilva: Agree. Done !

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: If multiple versions match, the last one will prevail.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
kjin: TS optionality doesn't refer to whether it is required to be used, but whether it is required to be given a value. In the latter case it is not optional because some plugins just won't work if you don't pass it.

I understand that you'd have to update the plugins to add a `basedir` argument -- if you agree that it should be required then it should be done in this PR.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Done!

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Done!

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: The basedir parameter is passed to the callback of the hook function from lib "require-in-the-middle", it is an absolute path.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Agree.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Method loadPlugins from PluginLoader is always passing this parameter value.
```
 result = plugin.applyPatch(exports, this.tracer, version, basedir);
```

fabiogomessilva: The implementations that will require this parameter will have it available.

kjin: I think you might be confounding optional assignment vs. optional usage:

__optional assignment__
```js
// it is OK for us not to pass a value for `basedir`.
plugin.applyPatch(exports, this.tracer, version);
```
__optional usage__
```js
class MyPlugin extends BasePlugin {
  applyPatch(exports, tracer, version, basedir) {
    // it is OK for us not to use `basedir`.
  }
}
```

The syntax `basedir?` suggests that `basedir`'s _assignment_ is optional. But as we have both said, the real intent that we want to convey is that its _usage_ is optional. (There is no syntax for this; the only way to show this intent is to document it.)

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Done!

packages/opencensus-nodejs/test/instrumentation/node_modules/@opencensus/instrumentation-simple-module/simple-module.ts
fabiogomessilva: A new one was created.

packages/opencensus-nodejs/test/test-plugin-loader.ts
fabiogomessilva: I think if the internal file export is accessible, whats the point to load it this way?

fabiogomessilva: Test improved.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -58,43 +64,63 @@ export abstract class BasePlugin implements types.Plugin {
...
+
+  applyPluginUnPatch() {
+    this.applyUnpatch();
+  }
+
+  /**
+   * applyPatch() and applyUnpatch()
+   *
+   * GoF Template Method Pattern - this is the variant part of the pattern
+   * Each plugin should implement his own version
```
kjin: I wasn't aware that this pattern had a specific name, TIL. However, I don't think this necessary belongs in docs, it may be confusing to users. Not everyone has read GoF.

I would just mention that this method will be called when `applyPluginPatch` is called.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
kjin: Can you add a logger warning? It should be brought to the user's attention because it is mostly a bug for someone to user overlapping version ranges.

packages/opencensus-core/src/trace/instrumentation/types.ts
```diff
@@ -23,13 +23,15 @@ export interface Plugin {
    * @param moduleExports nodejs module exports from the module to patch
    * @param tracer a tracer instance
    * @param version version of the current instaled module to patch
+   * @param basedir module absolute path
    */
-  // tslint:disable:no-any
-  applyPatch(
+  applyPluginPatch(
```
kjin: Suggestion: `enable` and `disable` to distinguish this from `applyPatch` and `applyUnpatch`.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Design Patterns are meant to be a kind of universal language, I think it is not confusing at all, it adds more information and makes things clear.  I think our audience is pretty much aware of them.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -118,19 +121,28 @@ export abstract class BasePlugin implements types.Plugin {
    */
   private loadInternalFiles(): ModuleExportsMapping {
     let result: ModuleExportsMapping = null;
+    let overlap = false;
     if (this.internalFileList) {
       this.logger.debug('loadInternalFiles %o', this.internalFileList);
       Object.keys(this.internalFileList).forEach(versionRange => {
         if (semver.satisfies(this.version, versionRange)) {
+          if (result) {
```
kjin: Not to be pedantic, but this can yield strange results if `result` just happens to be a falsey value (`0` or `''`). For correctness' sake I would just make `overlap` a counter (or otherwise just print the warning right in this function).

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -44,26 +64,111 @@ export abstract class BasePlugin implements types.Plugin {
...
-  // while unpatching is not. It should be the other way around
+  disablePluginPatch() {
+    this.applyUnpatch();
+  }
 
+  /**
+   * applyPatch() and applyUnpatch()
+   *
+   * GoF Template Method Pattern - this is the variant part of the pattern
+   * Each plugin should implement his own version.
```
kjin: I would not consider this universal language. I did not know what this was referring to and had to look it up. It might be true that I am not the common case, but it is better not to assume something like this about the audience. After all, it is possible to understand the concept without knowing its name.

Feel free to document the design pattern at the top of the file, but I do not think it belongs here.

As another tidbit, we should not use gender pronouns in code, especially when referring to impersonal objects. Please use `its`.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: Refactory done.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
fabiogomessilva: > ... document the design pattern at the top of the file ...

At the top, it loses context. I have improved the message.

Issue #11, opened 2018-05-08
Package naming


ofrobots:
Looking into the package.json for opencensus-core, it seems the package name is: `@opencensus/opencensus-core`. Why not `@opencensus/core`? I want to make sure this is a conscious decision and the pros/cons are documented somewhere.

kjin:
We've changed the package names to be shorter as you suggested. As for pros/cons, there aren't really any pros for keeping it as `@opencensus/opencensus-core` as far as I know. `@opencensus/core` is better because it is less redundant.

ofrobots:
👍

PR #61, opened 2018-06-23
refactor: rename core package and remove exporter namespaces: types and classes

fabiogomessilva:
This PR is addressing issue #11 and also removing namespaces types and classes from core exports.


packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -30,7 +30,7 @@ import * as types from './types';
 /**
  * This class represent a tracer.
  */
-export class Tracer implements types.Tracer {
+export class TracerAgent implements types.Tracer {
```
kjin: I much prefer `Tracer` as the class name, especially because this class is publicly exposed. In other words, I would rather the interface name change rather than the class. Or, I believe that if there is a 1:1 association between an interface and an implementation, that there not be an interface at all -- and this is one of those situations.

packages/opencensus-core/src/trace/sampler/sampler.ts
```diff
@@ -15,61 +15,45 @@
...
-  private idUpperBound: number;
-  description: string;
-
-  /**
-   * Constructs a new Sampler instance.
-   */
-  constructor() {}
+/**  Sampler that samples every trace. */
+export class AlwaysSampler implements Sampler {
+  description = 'always';
```
kjin: Do we still need the `description` field? Now if you want to figure out what a sampler is doing or the probability, you can use
```
if (sampler instanceof ProbabilitySampler) {
  let prob = sampler.getProbability();
}
```

packages/opencensus-core/src/trace/sampler/sampler.ts
```diff
@@ -87,4 +71,28 @@ export class Sampler implements types.Sampler {
       return false;
     }
   }
-}
\ No newline at end of file
+}
+
+
+/** Builder class of Samplers */
+export class SamplerBuilder {
```
kjin: +1 to having separate `Sampler` classes. This is much better.

packages/opencensus-core/src/trace/sampler/sampler.ts
fabiogomessilva: > Do we still need the description field? 

Maybe we don't, but Java implementation has a description also.  I would like to keep it if you don't mind.

packages/opencensus-core/src/trace/sampler/sampler.ts
fabiogomessilva: Agree.

packages/opencensus-core/src/trace/model/tracer.ts
fabiogomessilva: We can identify a general convention at @opencensus/core, there are always interfaces for exported classes.  I personly like the idea of a Tracer interface here. All instrumentations rely on that interface, I think that is a good approach to rely on interfaces rather than concrete implementations.  If we think of interfaces as 'contracts' between the system and its clients, the way it is been used by the core package is very natural.

packages/opencensus-core/src/trace/sampler/sampler.ts
kjin: Ok. Let's make it and the corresponding interface field `readonly`. Also, please add code documentation on its purpose in the interface.

packages/opencensus-core/src/trace/model/tracer.ts
kjin: > I think that is a good approach to rely on interfaces rather than concrete implementations

That is fair. Then I would suggest changing the interface name. For instrumentation, we don't use the entire `Tracer` interface; as an example, we actually don't want instrumentation to start or stop the tracer, access the sampler, or register span event listeners (AFAIK). Maybe we can define an interface that specifies the type of the object passed to instrumentation (like `SpanCreator`), and a different interface for the type of object that is being managed by the `Tracing` instance (like `SpanEventManager`). Then we would just have `class Tracer implements SpanCreator, SpanEventManager`. For anything that doesn't fit in either of these categories, we should ask whether it needs to be publicly exposed at all. (The names are just suggestions, I don't have strong feelings about them.)

I don't want to force my opinion on this one, so if you disagree that's fine; I would still prefer that a different name be given to the class. Maybe `BaseTracer` (it's not entirely accurate, but I can't think of anything better than that).

packages/opencensus-core/src/trace/sampler/types.ts
```diff
@@ -14,36 +14,13 @@
...
-import {Exporter} from '../../exporters/types';
-import {PluginNames} from '../instrumentation/types';
 
-/** This interface represent the probability of a tracer. */
+/** This interface represent a sampler . */
 export interface Sampler {
   /**
    * A Sampler description name
    */
   description: string;
```
kjin: Mark this as `readonly` as well. Also, let's expand on the documentation a little bit:
```
A string that uniquely describes the sampling behavior of this instance.
```

packages/opencensus-core/src/trace/model/tracer.ts
fabiogomessilva: > .. Maybe we can define an interface that specifies the type of the object passed to instrumentation (like SpanCreator), and a different interface for the type of object that is being managed by the Tracing instance (like SpanEventManager).  and a different interface for the type of object that is being managed by the Tracing instance (like SpanEventManager). Then we would just have class Tracer implements SpanCreator, SpanEventManager.

I understand the reasons, but I'm really not sure if creating this new set of interfaces could bring any benefit to the code, on the contrary, maybe add complexity. The "tracer" name representing the manager/controller class of Spans is a common name in the domain space, it is used by Opencensus and others Trace solutions.

Regarding the name of the class, I've chosen 'CoreTracer' over 'BaseTracer', cause the word Base feels like we have to build something over it.

packages/opencensus-core/src/trace/sampler/types.ts
fabiogomessilva: Done.

packages/opencensus-core/src/trace/model/tracer.ts
kjin: Thanks. `CoreTracer` is fine.

PR #65, opened 2018-06-27
chore: bump require-in-the-middle to ^3.0.0

watson:
This new version allows for multiple versions of require-in-the-middle to be present in the same dependency-tree without interfering.


watson:
I signed it!

watson:
I think the CI failures are unrelated right?

PR #71, opened 2018-07-02
test: improve console exporter tests

eduardoemery:


packages/opencensus-core/src/exporters/console-exporter.ts
```diff
@@ -33,8 +33,8 @@ export class NoopExporter implements types.Exporter {
 /** Format and sends span data to the console. */
 export class ConsoleExporter implements types.Exporter {
   /** Buffer object to store the spans. */
-  private buffer: ExporterBuffer;
   private logger: loggerTypes.Logger;
+  buffer: ExporterBuffer;
```
kjin: Please keep this private.

packages/opencensus-core/test/test-console-exporter.ts
```diff
@@ -60,11 +44,15 @@ describe('NoopExporter', () => {
...
       const rootSpan = new RootSpan(tracer);
-      const queue: RootSpan[] = [];
-      queue.push(rootSpan);
+      const queue: RootSpan[] = [rootSpan];
 
-      exporter.publish(queue);
-      assert.ok(true);
+      exporter.publish(queue)
+          .then(() => {
+            assert.ok(true);
```
kjin: I would like to avoid this line entirely. If there are no conditions to test, that strongly suggests that there is no point in exercising this code path at all.

packages/opencensus-core/test/test-console-exporter.ts
```diff
@@ -60,11 +44,15 @@ describe('NoopExporter', () => {
     it('should do anything', () => {
```
kjin: `'should do nothing'` should be more accurate. After all, it's a no-op Exporter.

packages/opencensus-core/test/test-console-exporter.ts
```diff
@@ -60,11 +44,15 @@ describe('NoopExporter', () => {
...
+      const queue: RootSpan[] = [rootSpan];
 
-      exporter.publish(queue);
-      assert.ok(true);
+      exporter.publish(queue)
+          .then(() => {
+            assert.ok(true);
+          })
+          .catch((err: Error) => {
+            assert.fail(err.message);
```
kjin: Instead of using `.catch` and failing in the catch handler, just return `exporter.publish(queue)`.

packages/opencensus-core/test/test-console-exporter.ts
```diff
@@ -75,12 +63,11 @@ describe('ConsoleLogExporter', () => {
...
       tracer.registerSpanEventListener(exporter);
-      // const rootSpan = new RootSpan(tracer);
-      const rootSpans = createRootSpans();
-      for (const rootSpan of rootSpans) {
+
+      tracer.startRootSpan({name: 'testRootSpans'}, rootSpan => {
         rootSpan.end();
-      }
-      assert.ok(true);
+        assert.ok(exporter.buffer.getQueue().length > 0);
```
kjin: Is this testing the tracer, the exporter buffer, or the exporter? It's not clear which one, but it doesn't seem like it is really testing functionality that is unique to `ConsoleLogExporter`.

packages/opencensus-core/test/test-console-exporter.ts
```diff
@@ -75,12 +63,11 @@ describe('ConsoleLogExporter', () => {
     it('should end a span', () => {
       const exporter = new ConsoleExporter(defaultBufferConfig);
       tracer.registerSpanEventListener(exporter);
-      // const rootSpan = new RootSpan(tracer);
-      const rootSpans = createRootSpans();
-      for (const rootSpan of rootSpans) {
+
+      tracer.startRootSpan({name: 'testRootSpans'}, rootSpan => {
```
kjin: I know this wasn't part of the change, but I don't think we need to use a `Tracer` object to ensure that a `ConsoleLogExporter` works. You can just call `onEndSpan` directly rather than having it called through when the span is ended.

packages/opencensus-core/src/exporters/console-exporter.ts
```diff
@@ -61,19 +61,19 @@ export class ConsoleExporter implements types.Exporter {
...
-      const result: string[] = [];
 
       result.push(
           ROOT_STR + '\n\tChildSpans:\n' +
           `${SPANS_STR.join('\n')}`);
-      console.log(`${result}`);
     });
-    return Promise.resolve();
+    console.log(`${result}`);
+    return Promise.resolve(`${result}`);
```
kjin: Why do we return the stringified result?

packages/opencensus-core/src/exporters/console-exporter.ts
eduardoemery: I had to make it public so I could access the buffer queue from the tests, also, I think the exporter buffer should be accessible outside the exporter. But we can accomplish this in different ways, if you prefer, I could set it private and make a get method or return the buffer queue at `onEndSpan()` (this last one is not my favorite). We could also think about other ways or if accessing the exporter buffer is really necessary.

packages/opencensus-core/test/test-console-exporter.ts
eduardoemery: In fact there is no need. I changed to just return `exporter.publish(queue)` as you suggested just to make sure we don't get any errors.

packages/opencensus-core/test/test-console-exporter.ts
eduardoemery: Sure. Updated.

packages/opencensus-core/test/test-console-exporter.ts
eduardoemery: Way better this way. Just changed, thanks :D

packages/opencensus-core/test/test-console-exporter.ts
eduardoemery: This is just to check if the span was added to the buffer, that should be the proper behaviour of `onEndSpan()`. I also added a few more assertions to make sure the span in the exporter buffer is right.

packages/opencensus-core/test/test-console-exporter.ts
eduardoemery: Makes sense to me, updated to your suggestion.

packages/opencensus-core/src/exporters/console-exporter.ts
eduardoemery: The stringified result is what we end up sending to `console.log()`. I think it makes sense to return a `string` here too, but returning just `result` works fine. Do you prefer `Promise.resolve(result)`? It's ok for me to change.

packages/opencensus-core/src/exporters/console-exporter.ts
kjin: I am typically against making class members public for testing, because they are implementation details, and we end up testing behavior that is not observable by normal users of the class. The same goes for getters for class members.

There are a few ways to work around this:
1) We can test _only_ observable behavior: initialize the `ConsoleExporter` with a buffer of size `X`, add `X-1` spans, verify that nothing has been written to stdout yet, then add one more and verify that all spans were written to stdout.
2) We can use dependency injection: Instead of initializing `this.buffer` in the constructor with a `new ExporterBuffer`, pass it as a parameter and have the caller create the `ExporterBuffer` object. This way, you don't have to test the behavior of `ConsoleExporter` that relies on `ExporterBuffer`: for example, the behavior in (1) doesn't really pertain to `ConsoleExporter` at all, it is a property of `ExporterBuffer`. Instead, the behavior you are testing is that `ConsoleExporter` correctly uses `ExporterBuffer` -- and this is possible by just passing in a mock `ExporterBuffer` implementation.

I prefer (2) and think it is a more longer-term solution.

packages/opencensus-core/test/test-console-exporter.ts
```diff
@@ -90,11 +73,19 @@ describe('ConsoleLogExporter', () => {
...
-      assert.ok(true);
+      exporter.publish(queue)
+          .then((result: string) => {
+            assert.ok(result.indexOf(rootSpan.traceId) >= 0);
+            assert.ok(result.indexOf(rootSpan.id) >= 0);
+            assert.ok(result.indexOf(rootSpan.name) >= 0);
+            assert.ok(result.indexOf(rootSpan.spans[0].name) >= 0);
+            assert.ok(result.indexOf(rootSpan.spans[0].id) >= 0);
+          })
+          .catch((err: Error) => {
```
kjin: I think I have the same comment as earlier, `.catch` is not needed if we just return `exporter.publish(queue).then()`. (BTW, without that change this test will pass synchronously before the asserts are run.)

packages/opencensus-core/src/exporters/console-exporter.ts
eduardoemery: I understand your point. I really like the dependency injection approach but to create a buffer I need an exporter. I could change the `ExporterBuffer` constructor to accept an exporter optionally and set it latter, but that simple change would affect many exporters, if not all. So, I ended up doing your first suggestion, I intercepted stdout before and after the `bufferSize` threshold.

PR #72, opened 2018-07-03
fix: add states to plugin loader mechanism

kjin:
If tests pass, this should allow #65 to land.

### DIFF
--- a/packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
+++ b/packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
@@ -115,35 +115,31 @@ export class PluginLoader {
   loadPlugins(pluginList: PluginNames) {
     if (this.hookState === HookState.UNINITIALIZED) {
       hook(Object.keys(pluginList), (exports, name, basedir) => {
-        if (this.hookState === HookState.ENABLED) {
-          const version = this.getPackageVersion(name, basedir as string);
-          this.logger.info('trying loading %s.%s', name, version);
-          let moduleExports = exports;
-          if (!version) {
-            return moduleExports;
-          } else {
-            this.logger.debug('applying patch to %s@%s module', name, version);
-            this.logger.debug(
-                'using package %s to patch %s', pluginList[name], name);
-            // Expecting a plugin from module;
-            try {
-              const plugin: Plugin = require(pluginList[name]).plugin;
-              this.plugins.push(plugin);
-              moduleExports =
-                  plugin.enable(exports, this.tracer, version, basedir);
-            } catch (e) {
-              this.logger.error(
-                  'could not load plugin %s of module %s. Error: %s',
-                  pluginList[name], name, e.message);
-            }
-            return moduleExports;
-          }
-        } else {
+        if (this.hookState !== HookState.ENABLED) {
+          return exports;
+        }
+        const version = this.getPackageVersion(name, basedir as string);
+        this.logger.info('trying loading %s.%s', name, version);
+        if (!version) {
+          return exports;
+        }
+        this.logger.debug('applying patch to %s@%s module', name, version);
+        this.logger.debug(
+            'using package %s to patch %s', pluginList[name], name);
+        // Expecting a plugin from module;
+        try {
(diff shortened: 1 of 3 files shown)


packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -102,30 +113,37 @@ export class PluginLoader {
...
-              pluginList[name], name, e.message);
+    if (this.hookState === HookState.UNINITIALIZED) {
+      hook(Object.keys(pluginList), (exports, name, basedir) => {
+        if (this.hookState === HookState.ENABLED) {
+          const version = this.getPackageVersion(name, basedir as string);
+          this.logger.info('trying loading %s.%s', name, version);
+          let moduleExports = exports;
+          if (!version) {
+            return moduleExports;
+          } else {
```
ofrobots: Get rid of `else` and unindent as the `then` block has a return.

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -102,30 +113,37 @@ export class PluginLoader {
...
-              'could not load plugin %s of module %s. Error: %s',
-              pluginList[name], name, e.message);
+    if (this.hookState === HookState.UNINITIALIZED) {
+      hook(Object.keys(pluginList), (exports, name, basedir) => {
+        if (this.hookState === HookState.ENABLED) {
+          const version = this.getPackageVersion(name, basedir as string);
+          this.logger.info('trying loading %s.%s', name, version);
+          let moduleExports = exports;
+          if (!version) {
+            return moduleExports;
```
ofrobots: return `exports`. Move `moduleExports` declaration down.

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -102,30 +113,37 @@ export class PluginLoader {
...
+          } else {
+            this.logger.debug('applying patch to %s@%s module', name, version);
+            this.logger.debug(
+                'using package %s to patch %s', pluginList[name], name);
+            // Expecting a plugin from module;
+            try {
+              const plugin: Plugin = require(pluginList[name]).plugin;
+              this.plugins.push(plugin);
+              moduleExports =
```
ofrobots: return `plugin.enable(...)`. Get rid of variable `moduleExports`.

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -102,30 +113,37 @@ export class PluginLoader {
...
+                  plugin.enable(exports, this.tracer, version, basedir);
+            } catch (e) {
+              this.logger.error(
+                  'could not load plugin %s of module %s. Error: %s',
+                  pluginList[name], name, e.message);
```
ofrobots: return `exports` here.

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -102,30 +113,37 @@ export class PluginLoader {
...
+            }
+            return moduleExports;
+          }
+        } else {
```
ofrobots: Switch the order of the `then` and `else` blocks. Observe that the new then block has a return in it. Remove else and unident.

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -102,30 +113,37 @@ export class PluginLoader {
...
+          return exports;
         }
-        return moduleExports;
-      }
-    });
+      });
+      this.hookState = HookState.ENABLED;
```
ofrobots: Make this the first line of the if uninitialize block; otherwise early returns miss this.

kjin: Actually the hook callback is executed on a different tick. But that being said, thanks for highlighting this as I should be bringing this out of the `if` block entirely (otherwise re-activation won't work).

PR #74, opened 2018-07-09
feat: add jaeger uber-trace-id header propagation

tcolgate:
Jaeger uses a custom header for propagating span context. They are intending
to move to the w3c spec, but there is no clear time frame on that, lots of
existinng things using jaeger tracing will need this for now.

### DIFF
--- a/packages/opencensus-propagation-jaeger/src/jaeger-format.ts
+++ b/packages/opencensus-propagation-jaeger/src/jaeger-format.ts
@@ -1,0 +1,105 @@
+/**
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {HeaderGetter, HeaderSetter, Propagation, SpanContext} from '@opencensus/core';
+
+import * as crypto from 'crypto';
+import * as uuid from 'uuid';
+
+const TRACE_ID_HEADER = 'uber-trace-id';
+const DEBUG_ID_HEADER = 'jaeger-debug-id';
+
+const DEBUG_VALUE = 2;
+const SAMPLED_VALUE = 1;
+
+/**
+ * Propagates span context through Jaeger trace-id propagation.
+ *  The format of the header is described in the jaeger documentation:
+ *  (https://www.jaegertracing.io/docs/client-libraries/)
+ */
+export class JaegerFormat implements Propagation {
+  /**
+   * Gets the trace context from a request headers. If there is no trace context
+   * in the headers, null is returned.
+   * @param getter
+   */
+  extract(getter: HeaderGetter): SpanContext|null {
+    if (getter) {
+      let debug = 0;
+      if (getter.getHeader(DEBUG_ID_HEADER)) {
+        debug = SAMPLED_VALUE;
+      }
+
+      const spanContext = {traceId: '', spanId: '', options: debug};
+
+      let header = getter.getHeader(TRACE_ID_HEADER);
+      if (!header) {
+        return spanContext;
+      }
+      if (header instanceof Array) {
+        header = header[0];
+      }
+      const parts = header.split(':');
+      if (parts.length !== 4) {
+        return spanContext;
+      }
+
+      spanContext.traceId = parts[0];
+      spanContext.spanId = parts[1];
+
+      const jflags = Number('0x' + parts[3]);
+      const sampled = jflags & SAMPLED_VALUE;
+
+      debug = (jflags & DEBUG_VALUE) || debug;
+
+      spanContext.options = (sampled || debug) ? SAMPLED_VALUE : 0;
+
+      return spanContext;
+    }
+    return null;
+  }
+
+  /**
+   * Adds a trace context in a request headers.
+   * @param setter
+   * @param spanContext
+   */
+  inject(setter: HeaderSetter, spanContext: SpanContext): void {
+    if (setter) {
+      let flags = '0';
+      if (spanContext.options) {
+        flags = (spanContext.options & SAMPLED_VALUE ? SAMPLED_VALUE : 0)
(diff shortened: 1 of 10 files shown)


tcolgate:
There seems to be some level of dissatisfaction with the idea of having this in tree.

kjin:
I wasn't aware of the discussion in the Python repo, thanks for the context.

I currently can't spare the bandwidth to review the code yet. Sorry for the inconvenience -- I hope to get to it soon!

tcolgate:
@kjin no probs. Do you have any (very approximate), time frame for the next likely release? We were going to wait on that rather than publishing a version of this package internally.

kjin:
I'll be able to answer that question later today. If you don't mind, please ping again if I don't have an answer by tomorrow.

kjin:
@tcolgate I can release the module on its own once I merge it, actually. That being said, I'll need to get confirmation from the OC team given the resistance in the Python thread.

tcolgate:
@kjin we're also waiting on the fixes for the jaeger exporter, and http tracing. I think the current release also has the issue where get isn't wrapped in https? (though that is already fixed on master). We more or less need a release of -core -node -exporter-jaeger and this. I don't mean that to sound pushy, that's just my "go live" list, the time frame is all down to you of course.

re: the python thing, the spec says alternate propagations can be supported. Seems odd to say that and then not actually accept any. Of course, if the decision goes against this, we can always publish separately.

kjin:
@tcolgate Ok, thanks for the heads up.

packages/opencensus-propagation-jaeger/test/test-jaeger-format.ts
```diff
@@ -0,0 +1,91 @@
...
+
+const UBER_TRACE_ID = 'uber-trace-id';
+
+const SAMPLED_VALUE = 0x1;
+const NOT_SAMPLED_VALUE = 0x0;
+
+const jaegerFormat = new JaegerFormat();
+
+describe('JaegerPropagation', () => {
+  /** Should get the singleton trancing instance. */
```
kjin: This comment seems like it doesn't apply, right?

packages/opencensus-propagation-jaeger/src/jaeger-format.ts
```diff
@@ -0,0 +1,97 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {HeaderGetter, HeaderSetter, Propagation, SpanContext} from '@opencensus/core';
+
+import * as crypto from 'crypto';
+import * as uuid from 'uuid';
+
+const UBER_TRACE_ID = 'uber-trace-id';
```
kjin: Is it more accurate to call this the "uber Trace ID" or the "Jaeger Trace ID"?

packages/opencensus-propagation-jaeger/test/test-jaeger-format.ts
```diff
@@ -0,0 +1,91 @@
...
+        }
+      };
+
+      jaegerFormat.inject(setter, spanContext);
+      assert.deepEqual(jaegerFormat.extract(getter), spanContext);
+    });
+  });
+
+
+  // Same test as propagation-stackdriver.
```
kjin: This seems to suggest that we don't really need `generate` to exist as a function on `Propagation` mechanisms. Not really anything to do in this PR, but what do you think?

packages/opencensus-propagation-jaeger/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "pretty": true,
+    "module": "commonjs",
+    "target": "es6",
+    "strictNullChecks": false
```
kjin: Not sure if this `tsconfig.json` was cloned from somewhere else, but either way, would the code still compile if this option was removed?

packages/opencensus-propagation-jaeger/src/jaeger-format.ts
tcolgate: We can't really change the header value, as the client libraries all assume uber-trace-id. Dropping the prefix seems reasonable.

packages/opencensus-propagation-jaeger/test/test-jaeger-format.ts
tcolgate: I couldn't see where that function actually every got called from. I've updated the jaeger version to produce 64 bit spanIds. I think the exporter gets to generate brand new traces/spans maybe? I don't think the propagator is involved at all.

tcolgate:
@kjin 
- Fixed a couple of errors in README
- Added link to jaeger client docs in the source (there was already one in the readme)
- Fixed things to work with strict null checks
- Updated traceId generation to use a 64 bit spanId

packages/opencensus-propagation-jaeger/src/jaeger-format.ts
kjin: Thanks. I was referring to the variable name only.

packages/opencensus-propagation-jaeger/src/jaeger-format.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  /**
+   * Generate SpanContexts
+   */
+  generate(): SpanContext {
+    return {
+      traceId: uuid.v4().split('-').join(''),
+      // tslint:disable-next-line:ban Needed to parse hexadecimal.
+      spanId: parseInt(crypto.randomBytes(8).toString('hex'), 16).toString(),
```
kjin: Just a warning here -- we used 6 bytes in other `generate()` functions because not all 8-byte unsigned integers are representable in JS. It probably won't have any end-user visible effect but `crypto.randomBytes(6)` has a better guarantee of uniformity.

packages/opencensus-propagation-jaeger/test/test-jaeger-format.ts
```diff
@@ -0,0 +1,90 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {HeaderGetter, HeaderSetter} from '@opencensus/core';
+import * as assert from 'assert';
+
+import {JaegerFormat} from '../src/';
+
+const UBER_TRACE_ID = 'uber-trace-id';
```
kjin: Can you change this to `TRACE_ID_HEADER` as well.

packages/opencensus-propagation-jaeger/src/jaeger-format.ts
tcolgate: Cool, I'll revert to six as it also verifies the padding.

kjin:
Thanks!

PR #77, opened 2018-07-14
fix: properly split low/high for 128-bit  spans

tcolgate:


packages/opencensus-exporter-jaeger/src/jaeger-driver/index.ts
```diff
@@ -59,13 +59,16 @@ export function spanToThrift(span: Span) {
 
   const spanLogs = [];
   const unsigned = true;
-  const length = span.spanContext.traceId.length;
   const parentSpan = span.parentSpanId ? Utils.encodeInt64(span.parentSpanId) :
                                          ThriftUtils.emptyBuffer;
 
+  const numericTraceId = Number('0x' + span.spanContext.traceId);
+  const traceIdLow = numericTraceId & (2 ^ 64 - 1);
```
kjin: Evaluating in an interpreter, `2 ^ 64 - 1` yields `61`. I get a feeling that the operation here won't work for JavaScript. Maybe we should be using Buffers here instead to manipulate byte arrays?

tcolgate: @kjin sorry, I raised a couple of these a bit prematurely. I need to add tests or this and actally try it out.

tcolgate:
@kjin I think the attached patch is good. The thirft encoding is build on node-int64, so this looks like a valid use.

packages/opencensus-exporter-jaeger/src/jaeger-driver/index.ts
```diff
@@ -59,13 +59,21 @@ export function spanToThrift(span: Span) {
...
   const spanLogs = [];
   const unsigned = true;
-  const length = span.spanContext.traceId.length;
   const parentSpan = span.parentSpanId ? Utils.encodeInt64(span.parentSpanId) :
                                          ThriftUtils.emptyBuffer;
 
+  let traceId = span.spanContext.traceId;
+  if (traceId.length < 32) {
+    const len = 32 - traceId.length + 1;
+    traceId = new Array(len).join('0') + traceId;
```
kjin: Suggestion: `` `00000000000000000000000000000000${traceId}`.slice(-32) ``

PR #78, opened 2018-07-14
feature: add support for annotations and messageevents to jaeger

tcolgate:


packages/opencensus-exporter-jaeger/src/jaeger-driver/index.ts
```diff
@@ -48,16 +48,39 @@ export type SenderCallback = (numSpans: number, err?: string) => void;
...
     Object.keys(span.attributes).forEach(key => {
       tags.push({'key': key, 'value': span.attributes[key]});
     });
-    spanTags = ThriftUtils.getThriftTags(tags);
   }
 
-  const spanLogs = [];
+  if (span.messageEvents) {
+    span.messageEvents.forEach(msg => {
+      tags.push({'key': 'message.id', 'value': msg.id});
```
kjin: Is it OK for multiple values to exist under the same key here? Here and for annotations as well.

packages/opencensus-exporter-jaeger/test/test-jaeger.ts
```diff
@@ -80,6 +80,18 @@ describe('Jaeger Exporter', () => {
...
+        span.addAttribute('testNum', 3.142);
+        span.addAnnotation(
+            'something happened',
+            Date.now(),
+            {
+              'error': true,
+              'severity': 'high',
+              'priority': 4,
+            },
+        );
```
kjin: Is it possible to tell that these attributes and annotations were written as Thrift tags?

packages/opencensus-exporter-jaeger/src/jaeger-driver/index.ts
tcolgate: Since these are getting translated to individual log events with tags, rather than just tags, it should be fine. I'm less sure what happens if you have a log event without a timestamp (which would be the case for message events).
 I don't really have a good grasp of messageEvents in general, but dropping them would seem like the wrong thing to do.
  I've not had a chance to actually run this yet, shouldn't have PR'd, sorry.

packages/opencensus-exporter-jaeger/test/test-jaeger.ts
tcolgate: I'll try and improve the test.

tcolgate:
Added a test, though I'm in the process of making sure it actually works too :)

tcolgate:
One thing, message events should have a timestamp, according to the spec.
https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/trace/trace.proto

tcolgate:
I'm going to add timestamp to addMessageEvent, and raise that in a different PR.

tcolgate:
@kjin this requires https://github.com/census-instrumentation/opencensus-node/pull/81

kjin:
I've merged #81.

tcolgate:
@kjin thanks for tidying that up.

PR #82, opened 2018-07-23
refactor: make adjustments to stats design

eduardoemery:
Changes design to better suit OpenCensus Specs.


packages/opencensus-core/src/exporters/types.ts
```diff
@@ -25,8 +24,13 @@ export interface Exporter extends modelTypes.SpanEventListener {
    * Sends a list of root spans to the service.
    * @param rootSpans A list of root spans to publish.
    */
-
   publish(rootSpans: modelTypes.RootSpan[]): Promise<number|string|void>;
 }
 
+/** Defines a StatsEventListener interface */
```
kjin: Not sure if this is within the scope of this change, but could you make this more descriptive?

packages/opencensus-core/src/stats/aggregation/types.ts
```diff
@@ -0,0 +1,108 @@
...
+ */
+
+import {Measurement, Tags} from '../types';
+
+/**
+ * Informs the type of the aggregation. It can be: count, sum, lastValue or
+ * distribution.
+ */
+export const enum AggregationType {
+  count = 0,
```
kjin: Could you make enum fields all caps?

packages/opencensus-core/src/stats/stats.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  /**
+   * Updates all views with the new measurements.
+   * @param measurements A list of measurements to record
+   */
+  record(measurements: Measurement[]) {
+    // TODO: To be implemented
+  }
+}
```
kjin: Can you add a newline here?

packages/opencensus-core/src/stats/types.ts
```diff
@@ -0,0 +1,104 @@
...
+  byte = 'by',   // bytes
+  kbyte = 'kb',  // Kbytes
+  sec = 's',     // seconds
+  ms = 'ms',     // millisecond
+  ns = 'ns'      // nanosecond
+}
+
+/** Describes the types of a Measure. It can be Int64 or a Double type. */
+export const enum MeasureType {
+  int64 = 'INT64',
```
kjin: As with other places, let's keep enum fields all caps.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -0,0 +1,71 @@
...
+    // TODO: To be implemented
+    return null;
+  }
+
+  /** Returns a list of all AggregationData in the view */
+  getSnapshots(): AggregationData[] {
+    // TODO: To be implemented
+    return null;
+  }
+}
```
kjin: Newline

packages/opencensus-core/src/stats/view.ts
```diff
@@ -0,0 +1,71 @@
...
+    // TODO: To be implemented
+  }
+
+  /**
+   * Returns a snapshot of an AggregationData for that tags/labels values.
+   * @param tags The desired data's tags
+   */
+  getSnapshot(tags: Tags): AggregationData {
+    // TODO: To be implemented
+    return null;
```
kjin: You don't need to change this since TODOs should get resolved soon, but for TODOs that have to do with a method not being implemented, let's throw instead of just letting it no-op. Better to fail fast if we end up missing an unimplemented function for some reason.

packages/opencensus-core/test/test-tracer.ts
```diff
@@ -299,4 +299,4 @@ describe('Tracer', () => {
       });
     });
   });
-});
\ No newline at end of file
+});
```
kjin: 👍

packages/opencensus-core/src/stats/stats.ts
```diff
@@ -0,0 +1,93 @@
...
+    return null;
+  }
+
+  /**
+   * Creates a measure of type Int64.
+   * @param name The measure name
+   * @param unit The measure unit
+   * @param description The measure description
+   */
+  createMeasureInt(name: string, unit: MeasureUnit, description?: string):
```
kjin: Can you actually call this `createMeasureInt64` and write as a comment that only integers up to `Number.MAX_SAFE_INTEGER` are allowed?

kjin:
Re: googlebot -- I will be rebasing.

packages/opencensus-core/src/stats/aggregation/types.ts
eduardoemery: Makes sense to me. Updated here and in other enums

packages/opencensus-core/src/stats/stats.ts
eduardoemery: My bad. Updated here and in other occurrences.

packages/opencensus-core/src/stats/types.ts
eduardoemery: Great. Updated.

packages/opencensus-core/src/stats/stats.ts
eduardoemery: Great. I also added this comment in `measurement.value` just as a reminder.

packages/opencensus-core/src/stats/view.ts
eduardoemery: Nice approach. Changed them to throw "Not Implemented".

packages/opencensus-core/src/exporters/types.ts
eduardoemery: This is related to the stats exporter. Changed the comments a little bit to be more clear. Please, tell me if it's still confusing.

eduardoemery:
Thanks for reviewing. I made a few changes based on what we discussed yesterday about `AggregationData`.

I removed the `SumAggregation`, `CountAggregation`, `LastValueAggregation` and `DistributionAggregation`, that were the implementations of the `AggregationData`s. As we discussed, I created a class called `Recorder` that will have this one method to update the aggregations.

We can discuss this on our next standup to see if there are better approaches.

packages/opencensus-core/src/stats/view.ts
kjin: Thanks for doing it thoroughly 👍

packages/opencensus-core/src/exporters/types.ts
kjin: `stats exporter` is better, as it shows who should be implementing it. To make it slightly more explicit:

```
An interface that describes the possible events that will be emitted from a Stats instance. Stats exporters should implement this interface.
```

PR #84, opened 2018-07-26
feat: Map Span attributes to labels for Stackdriver export

novemberborn:
I couldn't figure out how to get tests to pass locally, even without these changes. It seems to work if I copy the relevant lines into my project's dependency though.


kjin:
Thanks for the PR! You can run `npm run fix` in the `opencensus-exporter-stackdriver` directory to fix linting errors, which looks like it is the source of CI failures.

novemberborn:
@kjin thanks, done. Let's see what CI says 😄

PR #85, opened 2018-07-30
feat: add stats recorder implementation

eduardoemery:
- Add the `recorder` implementation
- Adds `recorder` tests
- Refactors code
    - removes `bucketsBoundaries`
    - changes `bucket.max` to `bucket.highBoundary`
    - changes `bucket.min` to `bucket.lowBoundary`

### DIFF
--- a/packages/opencensus-core/src/stats/recorder.ts
+++ b/packages/opencensus-core/src/stats/recorder.ts
@@ -45,7 +45,8 @@ export class Recorder {
     distributionData.count += 1;
 
     const inletBucket = distributionData.buckets.find((bucket) => {
-      return bucket.min <= measurement.value && measurement.value < bucket.max;
+      return bucket.lowBoundary <= measurement.value &&
+          measurement.value < bucket.highBoundary;
     });
     inletBucket.count += 1;
 
--- a/packages/opencensus-core/src/stats/recorder.ts
+++ b/packages/opencensus-core/src/stats/recorder.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {AggregationData, AggregationType, CountData, DistributionData, LastValueData, Measurement, SumData} from './types';
+import {AggregationData, AggregationType, CountData, DistributionData, LastValueData, Measurement, MeasureType, SumData} from './types';
 
 export class Recorder {
   static addMeasurement(
@@ -39,36 +39,37 @@ export class Recorder {
       distributionData: DistributionData,
       measurement: Measurement): DistributionData {
     distributionData.timestamp = Date.now();
+    const value = measurement.measure.type === MeasureType.DOUBLE ?
+        measurement.value :
+        Math.trunc(measurement.value);
 
     distributionData.count += 1;
 
     const inletBucket = distributionData.buckets.find((bucket) => {
-      return bucket.lowBoundary <= measurement.value &&
-          measurement.value < bucket.highBoundary;
+      return bucket.lowBoundary <= value && value < bucket.highBoundary;
     });
     inletBucket.count += 1;
 
-    if (measurement.value > distributionData.max) {
-      distributionData.max = measurement.value;
+    if (value > distributionData.max) {
+      distributionData.max = value;
     }
 
-    if (measurement.value < distributionData.min) {
-      distributionData.min = measurement.value;
+    if (value < distributionData.min) {
+      distributionData.min = value;
     }
 
     if (distributionData.count === 1) {
-      distributionData.mean = measurement.value;
+      distributionData.mean = value;
     }
 
-    distributionData.sum += measurement.value;
+    distributionData.sum += value;
 
     const oldMean = distributionData.mean;
     distributionData.mean = distributionData.mean +
-        (measurement.value - distributionData.mean) / distributionData.count;
+        (value - distributionData.mean) / distributionData.count;
     distributionData.sumSquaredDeviations =
         distributionData.sumSquaredDeviations +
-        (measurement.value - oldMean) *
-            (measurement.value - distributionData.mean);
+        (value - oldMean) * (value - distributionData.mean);
     distributionData.stdDeviation = Math.sqrt(
         distributionData.sumSquaredDeviations / distributionData.count);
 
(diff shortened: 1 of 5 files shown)


packages/opencensus-core/src/stats/recorder.ts
```diff
@@ -14,12 +14,86 @@
...
+    }
+
+    if (distributionData.count === 1) {
+      distributionData.mean = measurement.value;
+    }
+
+    distributionData.sum += measurement.value;
+
+    const oldMean = distributionData.mean;
+    distributionData.mean = distributionData.mean +
```
kjin: (1) I think this will introduce drift from roundoff errors over time. It might be best to re-compute the mean from `sum / count`. Though I guess there is no way to avoid the same issue for stdev.
(2) Is this formula correct? Do you have a source for this?

packages/opencensus-core/src/stats/types.ts
```diff
@@ -176,16 +176,14 @@ export interface DistributionData extends AggregationData {
...
-  readonly bucketsBoundaries: number[];
 }
 
 /** A simple histogram bucket interface. */
 export interface Bucket {
   /** Number of occurrences in the domain */
   count: number;
   /** The maximum bucket limit in domain */
-  readonly max: number;
+  readonly highBoundary: number;
```
kjin: For the comments here, can we say "the maximum (min) possible value for a data point to fall in this bucket"?

packages/opencensus-core/src/stats/recorder.ts
eduardoemery: I like the idea of using just `sum/count`, it's way simpler, but it may introduce overflow problems when `sum` gets too large, really large, which may not be a use case. If this specific situation won't happen, or will be really hard to happen, I'm okay with `sum/count`.

This formula comes from manipulating the mean definition itself, there is a nice explanation [here](http://datagenetics.com/blog/november22017/index.html). This is also [how Go does](https://github.com/census-instrumentation/opencensus-go/blob/master/stats/view/aggregation_data.go#L138). But, again, I'm on board with `sum/count` if that problem won't happen.

packages/opencensus-core/src/stats/types.ts
eduardoemery: Sure, I liked it.

packages/opencensus-core/src/stats/recorder.ts
kjin: Overflow is a good concern that I didn't think about. In that case I am more ambivalent, and incrementing might just as good given that it stays consistent with stdev. It will not be completely accurate for Double values but maybe that is more acceptable.

packages/opencensus-core/src/stats/recorder.ts
```diff
@@ -14,12 +14,86 @@
...
 
 export class Recorder {
   static addMeasurement(
       aggregationData: AggregationData,
       measurement: Measurement): AggregationData {
-    throw new Error('Not Implemented');
+    switch (aggregationData.type) {
+      case AggregationType.DISTRIBUTION:
+        return this.addToDistribution(
+            aggregationData as DistributionData, measurement);
```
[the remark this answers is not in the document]
eduardoemery: I made the changes that we discussed yesterday. But I still have to cast. The problem with union is that we create a type that has all the values, so, when we pass an `AggregationData` to a functions that asks for a `DistributionData`, for example, it points out that there is no attribute `value` in `DistributionData`, which makes sense. I think we can discuss this problem more today.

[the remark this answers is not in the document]
eduardoemery: Fixed now! Thanks for the help. Adding the `type: ...` really worked, TS was able to know which type we were working with.

eduardoemery:
I made one update that was missing. It addresses truncating the `INT64` measures before actually recording them. I also added a few tests for those.

If you prefer, I can remove this last commit and open a PR to address that. Please let me know.

packages/opencensus-core/src/stats/recorder.ts
```diff
@@ -14,63 +14,62 @@
...
-            aggregationData as LastValueData, measurement);
+        return this.addToLastValue(aggregationData, measurement);
     }
   }
 
   private static addToDistribution(
       distributionData: DistributionData,
       measurement: Measurement): DistributionData {
     distributionData.timestamp = Date.now();
+    const value = measurement.measure.type === MeasureType.DOUBLE ?
```
kjin: Suggestion: have `addTo*` take in `value` directly as an argument, and do this boilerplate in `addMeasurement`.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -28,7 +28,7 @@ export class BaseView implements View {
   readonly measure: Measure;
   /**
    * A map of stringified tags representing columns labels or tag keys, concept
-   * similar to dimensions on multidimensional modeling, to AggregationData.
+   * similar to dimensions on multidimensional modeling, to AggregationMetadata.
```
kjin: `AggregationMetadata` -> `AggregationData` in this file.

packages/opencensus-core/test/test-recorder.ts
```diff
@@ -83,101 +82,133 @@ describe('Recorder', () => {
...
-        }
-      });
-    }
-  });
+  for (const measure of measures) {
+    describe(
+        `Add measurements of type ${measure.type} to a Count Aggregation Data`,
+        () => {
+          for (const testCase of testCases) {
+            const countData: CountData = {
```
kjin: `countData` should be in the `it` callback clause... here and in other tests.

packages/opencensus-core/test/test-recorder.ts
```diff
@@ -83,101 +82,133 @@ describe('Recorder', () => {
...
-              DistributionData;
-
-          assertDistributionData(distributionData, sentValues);
-        }
-      });
-    }
-  });
+  for (const measure of measures) {
+    describe(
+        `Add measurements of type ${measure.type} to a Count Aggregation Data`,
```
kjin: Simplify description: `for count aggregation data of ${measure.type} values`.

packages/opencensus-core/src/stats/recorder.ts
eduardoemery: Ways simpler. Thanks for the suggestion.

packages/opencensus-core/src/stats/view.ts
eduardoemery: Thanks, there were a few other occurrences of this mistake, fixed.

packages/opencensus-core/test/test-recorder.ts
eduardoemery: Done.

packages/opencensus-core/test/test-recorder.ts
eduardoemery: Way better, thanks again.

PR #88, opened 2018-08-01
 chore(multiple): 0.0.3 release proposal

kjin:
[Release Notes](https://github.com/census-instrumentation/opencensus-node/releases/tag/untagged-33851390fb4b58408668)


kjin:
Thanks, I agree that once we hit 0.1.0 we will need to consider separate versioning.

PR #89, opened 2018-08-02
feat: add stats view implementation

eduardoemery:
Add stats view implementation and its tests.

### DIFF
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -80,6 +80,11 @@ export class BaseView implements View {
     this.bucketBoundaries = bucketBoundaries;
   }
 
+  /** Gets the view's tag keys */
+  getColumns(): string[] {
+    return this.columns;
+  }
+
   /**
    * Records a measurement in the proper view's row. This method is used by
    * Stats. User should prefer using Stats.record() instead.
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -24,7 +24,7 @@ export class BaseView implements View {
    */
   readonly name: string;
   /** Describes the view, e.g. "RPC latency distribution" */
-  readonly description?: string;
+  readonly description: string;
   /** The Measure to which this view is applied. */
   readonly measure: Measure;
   /**
@@ -46,7 +46,7 @@ export class BaseView implements View {
   /** The start time for this view */
   readonly startTime: number;
   /** The bucket boundaries in a Distribution Aggregation */
-  private bucketBoudaries?: number[];
+  private bucketBoundaries?: number[];
   /**
    * The end time for this view - represents the last time a value was recorded
    */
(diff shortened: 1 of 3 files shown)


packages/opencensus-core/src/stats/view.ts
```diff
@@ -32,32 +33,142 @@ export class BaseView implements View {
...
+   * @param tagsKeys The Tags' keys that view will have
+   * @param description The view description
+   * @param bucketBoudaries The view bucket boundaries for a distribution
+   * aggregation type
+   */
   constructor(
       name: string, measure: Measure, aggregation: AggregationType,
-      tagKeys: string[], description?: string) {
-    throw new Error('Not Implemented');
+      tagsKeys: string[], description?: string, bucketBoudaries?: number[]) {
```
kjin: `bucketBoundaries` (you can do a global replace)

packages/opencensus-core/src/stats/view.ts
```diff
@@ -32,32 +33,142 @@ export class BaseView implements View {
...
+   * Records a measurement in the proper view's row. This method is used by
+   * Stats. User should prefer using Stats.record() instead.
+   *
+   * Measurements with measurement type INT64 will have its value truncated.
+   * @param measurement The measurement to record
+   */
+  recordMeasurement(measurement: Measurement) {
+    // Checks if measurements has all tags in views
+    for (const tagKey of this.columns) {
+      if (!Object.keys(measurement.tags).find((key) => key === tagKey)) {
```
kjin: Prefer `some` over `find` here, as the return result could be an empty string. Of course any empty string key is not a real tag, but this will at least "work" if `some` is used instead.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -32,32 +33,142 @@ export class BaseView implements View {
...
+    Recorder.addMeasurement(this.rows[encodedTags], measurement);
+  }
+
+  /**
+   * Encodes a Tags object into a key sorted string.
+   * @param tags The tags to encode
+   */
+  private encodeTags(tags: Tags): string {
+    const encodedTags = Object.keys(tags).sort().reduce((strTags, tagKey) => {
+      return strTags + `${tagKey}:${tags[tagKey]},`;
```
kjin: Is `:` a reasonable delineator here?

packages/opencensus-core/src/stats/view.ts
eduardoemery: Thanks! Updated here and all occurrences.

packages/opencensus-core/src/stats/view.ts
eduardoemery: Great suggestion. Updated.

eduardoemery:
Made two small changes as we discussed:
- Makes view description mandatory
- Adds a method to get the view's columns

packages/opencensus-core/src/stats/view.ts
kjin: I see. It would be safer to take advantage of the [restrictions on tag keys/values](https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagContext.md). For example, to use non-printable characters as delineators, since both keys and values must use printable characters. That way there is guaranteed to be no name collision in weird cases where the value contains `:` or `,`.

Also, `map` + `join` might be clearer than `reduce`.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -112,10 +123,31 @@ export class BaseView implements View {
...
+      return this.invalidString(tagKey) || this.invalidString(tags[tagKey]);
+    });
+  }
+
+  /**
+   * Checks if a string has only printable characters.
+   * @param str The string to be checked
+   */
+  private invalidString(str: string): boolean {
+    return RegExp(`[${LOW_LIMIT_CHAR}-${HIGH_LIMIT_CHAR}]`).test(str) ||
```
kjin: Nit: I would suggest creating this `RegExp` in the file scope or as a static field so it doesn't have to get constructed on every call to `invalidString`. Its value is static anyway. (Same for next line.)

Also, would `/[^\u0020-\u007e]/` work as a regex to encompass both?

eduardoemery: I opted for `RegExp()` for readability when passing `LOW_LIMIT_CHAR` and `HIGH_LIMIT_CHAR` but hadn't considered the performance issue. I think it is way better with `/[^\u0020-\u007e]/`. Thanks again for the tips!

Issue #95, opened 2018-08-16
Support W3C TraceContext Propagation


justindsmith:
https://github.com/w3c/distributed-tracing

justindsmith:
Good catch. Closing now.

PR #97, opened 2018-08-17
feat: add tracecontext propagation

justindsmith:
Based on w3c TraceContext spec: https://w3c.github.io/distributed-tracing/

The `traceparent` header is validated and propagated according to spec.
The only caveat is that the option flags are lost downstream and reset to
"0x1" by the tracer. This will require a refactor of how the tracer treats
the options as certain assumptions are being made that the least-significant
bit will always represent "sampled".

The `tracestate` is propagated downstream, but is currently not mutable by
the tracer. It is simply stored as a string, but types have been generated
to allow this to be refactored into a more structured type in the future,
allowing the tracer to update the tracestate as necessary.

This commit does not attempt to implement the `Correlation-Context` header.

issue #95

### DIFF
--- a/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
+++ b/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
@@ -27,7 +27,7 @@ const TRACE_STATE = 'tracestate';
 const REQUESTED_FLAG = 0x1;
 const RECORDED_FLAG = 0x2;
 
-const DEFAULT_OPTIONS = REQUESTED_FLAG;
+const DEFAULT_OPTIONS = 0x0;
 
 /**
  * Propagates span context through Trace Context format propagation.
--- a/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
+++ b/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
@@ -1,0 +1,143 @@
+/**
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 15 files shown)


packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -0,0 +1,142 @@
...
+      // Read headers
+      let traceParent = getter.getHeader(TRACE_PARENT);
+      if (Array.isArray(traceParent)) {
+        traceParent = traceParent[0];
+      }
+
+      // Parse TraceParent into version, traceId, spanId, and option flags. All
+      // parts of the header should be present or it is considered invalid.
+      const parts = traceParent ? traceParent.split('-') : [];
+      if (parts.length === 4) {
```
kjin: Would it make more sense to return `null` if this condition (or subsequent conditions) isn't true?

kjin:
> This will require a refactor of how the tracer treats the options as certain assumptions are being made that the least-significant bit will always represent "sampled".

I see that 0x1 means "requested" [here](https://w3c.github.io/distributed-tracing/report-trace-context.html#requested-flag-00000001-). My interpretation is that this is the same as being sampled -- would you agree with that?

justindsmith:
> I see that 0x1 means "requested" here. My interpretation is that this is the same as being sampled -- would you agree with that?
I'm not sure.. I had a hard time deciphering what the `requested` vs `recorded` was. My understanding was that `requested` was a more explicitly "user requested" and not necessarily "sampled", but honestly I'm not 100% on either side.

[Here's the pull request](https://github.com/w3c/distributed-tracing/pull/142) that split the "sampled" flag into two different flags, which provides a little more context / discussion around the flags. I just noticed a recent comment about some confusion / disagreement around the proposed flags, so we might need to wait for that to get resolved / fleshed out.

Let me see if I can get some clarification from that thread to at least clear things up in my mind.

justindsmith:
> Let me see if I can get some clarification from that thread to at least clear things up in my mind.

From the thread above, sounds like `requested` is logically equivalent to "sampled". So I think you're right on that.

packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
kjin: I think (1) is a better choice, even if it involves further modifying the Tracer. From the spec:

> Notably, the tracestate field is unreliant on data in the traceparent.

So while it might not happen typically, a `tracestate` isn't necessarily invalid if sent in solation.

kjin:
Sorry for the delay -- I will merge this PR now. We can refactor the Tracer to not drop isolated tracestate in a follow-up.

bogdandrutu:
FYI: This is still discussed in the w3c group. May change.

PR #122, opened 2018-09-07
fix: propagate context, but don't sample

tcolgate:
Currently the code generates, but then ignores, unsampled root spans, if no
existing span context exists. It also drop span context that come in from
propagators if they are not sampled. As a result, any span context that is
propagated through the application is also sent to the exporters, regardless
of the options on the span (because only sampled spans will ever been seen).

This behaviour is incorrect, at least by reference to the python and Go codebases.

The sampling decision should be propagated, even if that decision is not to sample.
Consequently, trace IDs are assigned and propagated (to propagate the sampling
decision).

The attached PR is not meant for merging, but was the easiest way of outlining the
flow.

This makes it easier to write instrumentation, since instrumentation will then never
recieve a null span.

### DIFF
--- a/packages/opencensus-core/src/exporters/exporter-buffer.ts
+++ b/packages/opencensus-core/src/exporters/exporter-buffer.ts
@@ -73,6 +73,10 @@ export class ExporterBuffer {
    * @param root RootSpan to be added in the buffer.
    */
   addToBuffer(root: modelTypes.RootSpan) {
+    if (!(root.spanContext.options & 1)) {
+      return this;
+    }
+
     this.queue.push(root);
     this.logger.debug('ExporterBuffer: added new rootspan');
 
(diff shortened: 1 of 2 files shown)


bogdandrutu:
Completely agree that instrumentation should never receive null span because that will cause all the time to have:
```
if (span != null) {
  span.addAttribute(...);
}
```

Also propagating the ids (traceid, spanid, spampling decision) is good for example to do log correlation.

kjin:
Hi @tcolgate, is this ready for review, or is it still WIP?

tcolgate:
I meant it more for discussion. I think more work is needed. Are you happy with the idea that this needs fixing?

kjin:
I see -- you're proposing that we _always_ "trace", but only end up exporting if the sample bit is set, correct?

Generally, I agree with the notion of avoiding null spans, as well as propagating trace context regardless of sampling decision. In the Stackdriver Trace Agent, we used the concept of "phantom traces" that look like real traces to instrumentations (i.e. had the same interface), but would never actually get published -- this addressed the former but not the latter.

If the Python and Go implementations _always_ propagate trace context, I believe that is a strong argument that we should as well. I am a little wary of the performance overhead, however.

tcolgate:
@kjin I've got a bit more time to work on this now I think.
Beyond this change, Can you think of anything that we need, or anything specific we should test?

PR #126, opened 2018-09-13
Fix SpanId generation for interoperability

jflorencio:
SpanIds in opencensus-node are currently generated, sent in traces, and propagated to other services as numeric values. This isn't consistent with other opencensus libraries (e.g.: [opencensus-java](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/SpanId.java) serializes the span id as lowercase base16). When opencensus-node is setup with B3 propagation, this SpanId format violates the [B3 specification](https://github.com/openzipkin/b3-propagation#spanid-1) which mandates lowercase base16 encoding. Generally speaking, opencensus-java only accepts spans in this format so it will fail to correctly link to any span generated by opencensus-node. The best case scenario is a shared trace id with a missing parent span link.

This PR removes the integer parsing, and leaves it as lowercase hex characters.  I tested locally with B3 propagation where node creates the root span and propagates it to a Java service instrumented with opencensus-java, and verified the data is correctly ingested and correlated in Jaeger.

Update: I updated the stackdriver exporter to convert to and from its numeric spanId encoding.

### DIFF
--- a/packages/opencensus-core/src/internal/util.ts
+++ b/packages/opencensus-core/src/internal/util.ts
@@ -19,8 +19,7 @@
 
 import * as crypto from 'crypto';
 
-// Use 6 bytes of randomness only as JS numbers are doubles not 64-bit ints.
-const SPAN_ID_RANDOM_BYTES = 6;
+const SPAN_ID_RANDOM_BYTES = 8;
 
 // Use the faster crypto.randomFillSync when available (Node 7+) falling back to
 // using crypto.randomBytes.
@@ -32,6 +31,5 @@ const spanRandomBuffer = randomFillSync ?
     () => randomBytes(SPAN_ID_RANDOM_BYTES);
 
 export function randomSpanId() {
-  // tslint:disable-next-line:ban Needed to parse hexadecimal.
-  return parseInt(spanRandomBuffer().toString('hex'), 16).toString();
+  return spanRandomBuffer().toString('hex');
 }
--- a/packages/opencensus-propagation-b3/src/b3-format.ts
+++ b/packages/opencensus-propagation-b3/src/b3-format.ts
@@ -76,8 +76,7 @@ export class B3Format implements Propagation {
   generate(): SpanContext {
     return {
       traceId: uuid.v4().split('-').join(''),
-      // tslint:disable-next-line:ban Needed to parse hexadecimal.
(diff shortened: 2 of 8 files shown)


jflorencio:
Seems like stackdriver needs a numeric span id. Let me come up with a different solution here.

kjin:
> Seems like stackdriver needs a numeric span id. Let me come up with a different solution here.

For propagation through the `x-cloud-trace-context` header, yes. Given the limitation of JS number ranges, I think it is a reasonable trade-off to generate 48-bit integer values rather than use a user-land 64-bit integer string library... what do you think?

jflorencio:
Hmmm... While I agree that it's ideal to not rely on a library that does conversions on 64-bit integers in pure JavaScript, I don't think it's safe to assume a 6 byte span id. In particular, while we can limit the spanIds generated by this library, we still need to support spans that are propagated to opencensus-node as full 8-byte spanId values.

For example, if we're using stackdriver propagation through the `x-cloud-trace-context` header, but then we're exporting traces to most other tracing backends (e.g.: jaeger), or conversely, if we're using b3-propagation, but exporting to stackdriver, we'll need to accept and convert the full 64-bit value to base 16 or base 10 encoding, respectively. If we don't do the full conversion, we end up in our current situation - the spans don't get linked together since the span parentId emitted by opencensus-node is different from the span id emitted by the calling service.

I think there are optimizations that can be done, for example, generating 6 byte span ids by default, and then when a conversion is required, detecting if the value is more than 6 bytes, and doing a native or library-assisted conversion appropriately. Alternatively something like [bignum](https://github.com/justmoon/node-bignum) may be an option for more efficient conversions. Since spanId generation and conversion logic isn't centralized though and I wasn't too sure why, I just wanted to focus on correctness here.

LMK what you think.

kjin:
Thinking about it more, I agree that we should use an 8-byte span ID. Using 6 bytes in the core implementation is making an assumption that propagation libraries are converting span IDs to numerical values, but this assumption is not fair to propagation mechanisms that keep span IDs as hex values.

I may also be overvaluing the performance gain of using a native dec to hex converter. `hex2dec` is probably acceptable given that the operation is likely minuscule in the face of overall request latency. We can certainly pursue optimizations in a follow-up PR. So I think I'm on board with this change.

Issue #124, opened 2018-09-13
Add OC-Agent Node Exporter.


songy23:
The OC-Agent Node exporter should:

1. Act as a client of the [Trace RPC service](https://github.com/census-instrumentation/opencensus-proto/blob/d3b7200dbd14743faa22c6642291d76030833cbd/src/opencensus/proto/agent/trace/v1/trace_service.proto#L32) of OC-Agent;
2. Grab current `Span`s from OpenCensus-Node library, transform them to [Span proto](https://github.com/census-instrumentation/opencensus-proto/blob/d3b7200dbd14743faa22c6642291d76030833cbd/src/opencensus/proto/trace/v1/trace.proto#L37), and apply the incoming [Config](https://github.com/census-instrumentation/opencensus-proto/blob/d3b7200dbd14743faa22c6642291d76030833cbd/src/opencensus/proto/agent/trace/v1/trace_service.proto#L53) from Agent.

For reference, please see [OC-Agent Go Exporter](https://github.com/census-ecosystem/opencensus-go-exporter-ocagent) and [OC-Agent Python Exporter](https://github.com/census-instrumentation/opencensus-python/tree/master/opencensus/trace/exporters/ocagent).

/cc @justindsmith

songy23:
Thanks for the help with the Agent Exporter :)

> The only glaring issue I see right now is that the node sampler doesn’t support rate sampling, but it looks like the go oc-agent exporter treats that config change as a no-op right now, so I’ll just do the same until we can get that work done.

Yes, actually no languages have support for rate-limiting sampler at this moment since it's just proposed to be added to our spec (https://github.com/census-instrumentation/opencensus-specs/pull/158). For now it's safe to just ignore it.

 > How is testing being done with the oc-agent today?

In OC-Agent-Go-Exporter we're testing the Agent exporter against a mock Agent. Please take a look at [mock_agent_test.go](https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/master/v1/mock_agent_test.go). 

> Is there a way for me to validate my changes with a running agent instance?

Unfortunately testing against a real Agent instance is not available at this moment. @odeke-em and I are actively working to add Agent impl to OC-Service repo, please expect that to be done in the following week.

> There are also some other concerns architecturally around an exporter modifying the tracer config, but there are a few options that shouldn’t be too disruptive, so I’ll play around with that.

Exactly, that's also why we don't 100% commit to the config update feature with Agent:

[...]

PR #128, opened 2018-09-19
Initial OpenCensus Agent Exporter implementation

justindsmith:
The OpenCensus Agent Exporter conforms to the Exporter interface, receiving
spans as they complete, adding them to a buffer, and sending the spans from
the buffer to the agent over a grpc stream.

OpenCensus proto files have been included, as well as types for the required
agent proto files. A series of adapters have been written to adapt from the
@opencensus/core internal span representation to the proto definitions.

Grpc streams are monitored for any disconnect and attempt to reconnect if
possible, utilizing a backoff strategy provided by the grpc client.

The exporter also listens to the Config stream, which recieves configuration
updates from the agent. Specifically, the current tracer sampler is updated
to either a ProbabilitySampler or ConstantSampler. The RateLimitingSampler
option is currently a no-op and has no effect on the tracer sampler.


justindsmith:
This is the initial implementation of the agent exporter for issue #124. I will follow this up with a test suite (in this same PR) in a few days, but wanted to at least start the conversation. There are a few call out items here that I’d like some thoughts on.

/cc @kjin 

1. The implementation copies the opencensus proto files into the repo and I’ve hand-crafted the types for the relevant protos for the agent work. I think it makes more sense for these to live in the census-instrumentation/opencensus-proto repository and to auto-generate the types but didn’t want to take that work on just yet. It would involve deploying a new package out of that repo. I don’t think it delays this PR and we can optimize it in a future release. It’s something we should look at though.

2. To handle the configuration updates from the agent, I’m getting a handle to the current `tracer` instance by loading the `@opencensus/nodejs` and getting the instance from there. That binds the exporter to nodejs implementation, which I think is fine, but wanted to call that out explicitly.

/cc @songy23 

3. We are not supporting the stackTrace, sameProcessAsParentSpan, or childSpanCount properties today. The stackTrace isn’t supported by the core node implementation today. The `sameProcessAsParentSpan` I might be able to default to `true` (?) and the `childSpanCount` I can supply for the rootSpans but NOT for the child spans. Just let me know what you think is best with those.

4. I’d like some visibility into the `hexStringToUint8Array()` function in the `adapters.ts` file. Most of the adapters were straight forward, but since this is putting a value into a Buffer, I want to make sure that I’m handling that case properly (plus this is what we use to transport the trace/span id, so it’s pretty critical to get right).

5. In the core node implementation the MetricEvent.id is a string type, while in the agent type it’s a number. Looking at the code, the nodes implementation is creating a hex value, so I’m parsing that to an integer and sending that as the value, but that might be fragile. We might want to unify on the number type if that’s what’s expected.

6. Today we only support the `createInsecure()` credentials, does this need to be configurable by the user? Does the agent run in a secure way or can we assume we can always make an insecure connection?

I’m sure I’ll have more questions, but those should get us started!

songy23:
Thanks @justindsmith for getting a PR ready in such a short time :+1: 

> I think it makes more sense for these to live in the census-instrumentation/opencensus-proto repository and to auto-generate the types but didn’t want to take that work on just yet.

Make sense to me, please file a tracking issue on moving the gen-files to OpenCensus-Proto, similar to https://github.com/census-instrumentation/opencensus-python/issues/318/.

> We are not supporting the stackTrace, sameProcessAsParentSpan, or childSpanCount properties today. The stackTrace isn’t supported by the core node implementation today. 

Note that `StackTrace` is a deprecated concept and is not supported by any language (AFAIK). It's there just because of backwards-compatibility. See https://github.com/census-instrumentation/opencensus-go/issues/912#issuecomment-422505999.

> The sameProcessAsParentSpan I might be able to default to true (?) and the childSpanCount I can supply for the rootSpans but NOT for the child spans. Just let me know what you think is best with those.

FYI in Java data model, we have two optional fields [hasRemoteParent](https://github.com/census-instrumentation/opencensus-java/blob/217a1e86dd1082296b7278253f5f51633191fb06/api/src/main/java/io/opencensus/trace/export/SpanData.java#L95) and [childSpanCount](https://github.com/census-instrumentation/opencensus-java/blob/217a1e86dd1082296b7278253f5f51633191fb06/api/src/main/java/io/opencensus/trace/export/SpanData.java#L105) in `SpanData`, which correspond to `sameProcessAsParentSpan` and `childSpanCount` in proto. I think it's fine to leave a default value for them for now, but ultimately we should support both in the Node `Span` data model.

> I’d like some visibility into the hexStringToUint8Array() function in the adapters.ts file. Most of the adapters were straight forward, but since this is putting a value into a Buffer, I want to make sure that I’m handling that case properly (plus this is what we use to transport the trace/span id, so it’s pretty critical to get right).

Defer the answer to @kjin since I'm not quite familiar with the Node encoding implementation.

> In the core node implementation the MetricEvent.id is a string type, while in the agent type it’s a number. Looking at the code, the nodes implementation is creating a hex value, so I’m parsing that to an integer and sending that as the value, but that might be fragile. We might want to unify on the number type if that’s what’s expected.

Sorry I'm bit lost here - are you talking about `MessageEvent`? I didn't see there's a `MetricEvent` type in Node.  If so, I agree the encoding should be consistent across language and we need to put it under OpenCensus-Specs. (Note that there was a [PR](https://github.com/census-instrumentation/opencensus-specs/pull/2) about this but it was closed. /cc @bogdandrutu ) 

> Today we only support the createInsecure() credentials, does this need to be configurable by the user? Does the agent run in a secure way or can we assume we can always make an insecure connection?

I think it makes sense to make it configurable by users. Note that in Go-Agent-Exporter we have an [option](https://github.com/census-ecosystem/opencensus-go-exporter-ocagent/blob/00af367e65149ff1f2f4b93bbfbb84fd9297170d/options.go#L42-L45) for using `insecure`.

justindsmith:
> Make sense to me, please file a tracking issue on moving the gen-files to OpenCensus-Proto
Sure thing!

> Sorry I'm bit lost here - are you talking about MessageEvent? I didn't see there's a MetricEvent type in Node. If so, I agree the encoding should be consistent across language and we need to put it under OpenCensus-Specs.
Oops! Yeah, I meant `MessageEvent.id`. Luckily I don't think the damage is too broad right now, so I'll try to make the change in this PR assuming @kjin is okay with it. (I didn't see anything in your linked PR about MessageEvent id type though - and a quick look through the specs didn't reveal anything either - but I don't think this is a huge change on the node side.)

> I think it makes sense to make it configurable by users. Note that in Go-Agent-Exporter we have an option for using insecure.
Ok cool. I'll follow that model then.

> I think it's fine to leave a default value for them for now, but ultimately we should support both in the Node Span data model.
I'll take a look at those implementations and see what I can do, but might leave them as default for the time being.

Thanks for the quick replies!

songy23:
> I didn't see anything in your linked PR about MessageEvent id type though - and a quick look through the specs didn't reveal anything either - but I don't think this is a huge change on the node side.

Oh yes that PR is only about `Annotations`. Anyway I've filed https://github.com/census-instrumentation/opencensus-specs/issues/180 to follow up with adding cross-language specs on these fields.

> I'll take a look at those implementations and see what I can do, but might leave them as default for the time being.

Sounds good to me :)

justindsmith:
Updated the review to add unit tests and some other clean-up items. This PR is ready for a full review.

Here's from the commit:
- Add unit testing of exporter and adapters
- Ability to pass in a credentials to use for the grpc client connection.
- Move from `grpc-tools` to `google-proto-files`. The `grpc-tools` included a lot of unnecessary dependencies, we only need the google proto files.
- Added `remoteParent` to `sameProcessAsParentSpan` property, which flips the remoteParent boolean. Note that the `remoteParent` defaults to `false` and is NOT being set by any instrumentation today. // cc @songy23 
- General cleanup.

kjin:
Thanks for the contribution! Taking a look now

packages/opencensus-exporter-ocagent/src/ocagent.ts
```diff
@@ -0,0 +1,313 @@
...
+
+    // Create buffer
+    this.buffer = new ExporterBuffer(this, this._options);
+
+    /**
+     * Get node properties
+     */
+    this.exporterVersion = require('../../package.json').version;
+    this.coreVersion =
+        require('../../node_modules/@opencensus/core/package.json').version;
```
kjin: This seems a little brittle -- do you know of any third-party modules that can help with this -- or, should we have the `@opencensus/core` module directly export this as a field?

packages/opencensus-exporter-ocagent/src/ocagent.ts
```diff
@@ -0,0 +1,313 @@
...
+  private exporterVersion: string;
+  private coreVersion: string;
+  private hostName: string;
+  private processStartTimeMillis: number;
+
+  private activeLocal = false;
+
+  constructor(options: OCAgentExporterOptions) {
+    // Get the complete set of options from the defaults and inputs
+    this._options = Object.assign({}, DEFAULT_OPTIONS, options);
```
kjin: IMO we don't need to prefix this with an underscore.

packages/opencensus-exporter-ocagent/src/ocagent.ts
```diff
@@ -0,0 +1,313 @@
...
+      if (this.shouldAttemptReconnect(status)) {
+        this.logger.error(
+            `OCAgent: config stream disconnected; attempting reconnect`);
+        setImmediate(() => this.connectToConfigStream());
+      }
+    });
+    this.configStream.on(
+        StreamEvent.Error,
+        () => {
+            // no-op: Swallow errors to keep the process alive. Not listening
```
kjin: Should we log errors here?

packages/opencensus-exporter-ocagent/src/ocagent.ts
```diff
@@ -0,0 +1,313 @@
...
+        probability = update.config.probabilitySampler.samplingProbability;
+      } else if (
+          update.config.rateLimitingSampler &&
+          update.config.rateLimitingSampler.qps != null) {
+        // no-op
+        this.logger.warn('OCAgent: RateLimitingSampler is not supported');
+        return;
+      }
+
+      // If we have a valid probabily, create a new sampler
```
kjin: Spelling - `probability`

packages/opencensus-exporter-ocagent/tsconfig.json
```diff
@@ -0,0 +1,18 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "pretty": true,
+    "module": "commonjs",
+    "target": "es6",
+    "strictNullChecks": false
```
kjin: Can we enable `strictNullChecks`?

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -0,0 +1,464 @@
...
+        */
+       const runConstantTest = (decision: boolean): Promise<void> => {
+         server.sendConfigurationChangeConstant(decision);
+         const expectedSamplerDescription = decision ? 'always' : 'never';
+         return new Promise(resolve => {
+           setTimeout(() => {
+             const sampler = tracing.tracer.sampler.description;
+             // Validate sampler type
+             assert.ok(
+                 sampler === expectedSamplerDescription,
```
kjin: Nit -- Prefer `assert.strictEquals`.

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -0,0 +1,464 @@
...
+                 sampler.startsWith('probability'),
+                 `sampler ${sampler} is not of type 'probability'`);
+
+             // Validate set probability
+             assert.ok(
+                 sampler.endsWith(`(${testProbability})`),
+                 `sampler ${sampler} is not using probability ${
+                     testProbability}`);
+             resolve();
+           }, 15);
```
kjin: Can we set the server wait time as a top-level constant? Just for ease-of-use (might be useful to tweak if CI flakes occur)

packages/opencensus-exporter-ocagent/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "@opencensus/exporter-ocagent",
```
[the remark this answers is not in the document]
justindsmith: I generally agree with you, but it looks like other implementations are going with `ocagent` [ [1](https://github.com/census-instrumentation/opencensus-python/tree/master/opencensus/trace/exporters/ocagent) [2](https://github.com/census-ecosystem/opencensus-go-exporter-ocagent) [3](https://github.com/census-instrumentation/opencensus-java/tree/master/exporters/trace/ocagent) ], so I'd recommend  keeping it `ocagent` for consistency. /cc @songy23

packages/opencensus-exporter-ocagent/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, softwares
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export * from './ocagent';
```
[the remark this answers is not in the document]
justindsmith: You're right that I was mostly going for consistency. I _do_ like that `ocagent` is easier to search for in an IDE than `index` (and having to find the right file by subdirectory). So I'm actually in favor with keeping it this way for now. You get the benefit of importing from the root package but makes it easier to search for the right file in dev mode.

packages/opencensus-exporter-ocagent/src/ocagent.ts
justindsmith: > should we have the @opencensus/core module directly export this as a field?
You're right that this would be the better approach. Let me try to get that added.

justindsmith: Looking around a bit more, there is a package that can help expose `package.json` properties, but it might be overkill if we're just looking to expose `version`: https://www.npmjs.com/package/pkginfo

We can expose the version easily by just importing the package.json and exporting from there.

The problem for _this_ PR is that it will require a version bump in the core lib before we can use it here. So I propose we do the following:
* I'm going to leave the existing version code as-is for this PR.
* I'm going to open up a second PR to add the export version to the core lib.
* I'm going to open up an issue to track using the new exported core lib version in the ocagent exporter.

Let me know if that makes sense.

kjin: Thanks, that makes sense. I do agree that `pkginfo` is overkill.

The approach sounds good.

packages/opencensus-exporter-ocagent/package.json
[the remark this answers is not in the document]
kjin: Ok, in that case `ocagent` seems like the better option.

PR #132, opened 2018-10-02
fix: remove const enums

eduardoemery:
This PR addresses the following:

- [Nodejs Quickstart](https://opencensus.io/quickstart/nodejs/metrics/#1) giving a _Cant read property MS of MeasureUnit_ error
- OpenCensus Core compile script giving a _The node cant find 'utils' module_ error.
- Warning for not recompiling the library with `strictNullChecks` enabled

### DIFF
--- a/packages/opencensus-core/package.json
+++ b/packages/opencensus-core/package.json
@@ -41,6 +41,7 @@
     "@types/continuation-local-storage": "^3.2.1",
     "@types/mocha": "^2.2.48",
     "@types/node": "^9.4.7",
+    "@types/once": "^1.4.0",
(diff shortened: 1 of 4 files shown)


packages/opencensus-core/README.md
```diff
@@ -7,6 +7,8 @@ The library is in alpha stage and the API is subject to change.
 
 Please join [gitter](https://gitter.im/census-instrumentation/Lobby) for help or feedback on this project.
 
+_Note: If recompiling the library, the flag `strictNullChecks` should not be enabled._
```
kjin: Do you need to make this note here?

PR #142, opened 2018-10-11
Removes redundant range min/max from distributedValues

isaikevych:
[exporter-stackdriver]

Solves the error:
UnhandledPromiseRejectionWarning: Error: Field timeSeries[0].points[0].distributionValue.range is not allowed: Distribution range is not yet supported.


isaikevych:
I signed it!

PR #146, opened 2018-10-12
Fix export issue

grantwwu:
When re-exporting like this, I get this error message when I try to
invoke core.logger:

Cannot invoke an expression whose type lacks a call signature. Type 'typeof import("/file/path/censored/node_modules/@opencensus/core/build/src/common/console-logger")' has no compatible call signatures.

My theory is that using a default import for this breaks type signatures.


grantwwu:
Reviewing the CI logs, it seems that maybe the bug was actually in the example code and not the library code.  Will investigate further on Monday.

PR #150, opened 2018-10-17
Adds 1 minute delay on record for [exporter-stackdriver]

isaikevych:


songy23:
One high-level comment: I think a better way may be: 
1. For every `onRecord`, still generate `timeSeries` synchronously.
2. Buffer the `timeSeries` in an array and export them to backend every 1 minute (or other configured delay) asynchronously.
3. Export should go in parallel with `onRecord`, i.e shouldn't block each other.

This is what we did in other languages. What do you think?

isaikevych:
> One high-level comment: I think a better way may be:
> 
> 1. For every `onRecord`, still generate `timeSeries` synchronously.
> 2. Buffer the `timeSeries` in an array and export them to backend every 1 minute (or other configured delay) asynchronously.
> 3. Export should go in parallel with `onRecord`, i.e shouldn't block each other.
> 
> This is what we did in other languages. What do you think?
I need to invest some time for refactoring, because current implementation uses language specific approach. Also, I am using Promise for this needs, which is readable for nodejs devs.

isaikevych:
TODO: https://github.com/census-instrumentation/opencensus-node/issues/152

PR #160, opened 2018-10-23
Added ignoreUrls support

whs:
This add ignoreUrls support by adding a check to CoreTracer.

Ideally, I would add it to the HTTP tracer directly but that would 1) requires interface change to pass the configuration to instrumenters and 2) it would requires modifying all tracers to honer the flag.

I'm not sure what's the intention of the interface's designer here, so I decided to go with the path of least resistance.


justindsmith:
Hey @whs thanks for the first PR! :)

Looking a bit further, I’m not sure including the `ignoreUrls` in the tracing config is actually the right approach. As you pointed out, it adds too much plugin-specific logic into the root tracing config, which isn’t a good pattern.

You mention that the ideal would be to add it to the http plugin directly, and I think we can actually do that with minimal effects and no interface breaking.

I’d suggest we do the following:

1. Update the `opencensus-instrumentation-http` package to accept a configuration object that includes the `ignoreUrls` pattern and enforces the ignore logic. The default should be no ignoring (so there is no change in default behavior).
2. In your client code, you can instantiate the http plugin with the `ignoreUrls` configuration value you need and pass that through the [`plugins` tracing configuration](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/config/types.ts#L54-L68), which should override the default http plugin and use your instance instead. (Note that this might get a little weird as you actually need to create a file that can be `required` that exports a `plugin` instance object for your configuration that the plugin-loader will load. We might want to improve this if instantiating plugins becomes a more common workflow.)
3. We should remove the `ignoreUrls` configuration from the tracing to remove confusion (since it wasn’t being used before anyways as you point out). 

Generally, I think it's better to keep instrumentation-specific configurations localized to the instrumentation library as much as possible. Things will only get more confusing when we add more instrumentation.

What do you think about that? If you need help with anything don’t hesitate to ask.

@fabiogomessilva Adding you just in case you had different intentions for this config option.

/cc @isaikevych

whs:
I just got back from re:invent to start working on this.

I'm not sure how to pass a configuration object to the instrumenter.

- Looking at the entrypoint of instrumenter: plugin-loader is calling `instrumenter.enable(moduleToPatch, tracer, version, basedir)`
- So the only place we could get any configuration passed is by reading a field off tracer.
- The `Tracing` interface has no field that expose config passed to `start()`
  - I could add one but that means all tracers would need to be updated. It's also an interface not abstract class so future implementors would need to duplicate this functionality.

At this point I'd suggest these changes:

- Change TracingConfig.plugins to `string: string|PluginNameAndOption`, similar to webpack loader configuration
- Change `BasePlugin.enable` to accept one more argument: the option.

Should be mostly compatible with existing code.

justindsmith:
Yeah, my thought was to have the user write a new module that would pass the config to the plugin implementation, but I like your idea much better. It's also backward compatible, which is great. (As long as the plugin can handle an `undefined` configuration by default.)

I think that's the way to go, and having a way to send configuration to the plugins would be a great feature!

Would you have time to implement this? Or would you need some help on some of it?

whs:
Done! New version should be working even better than the original one.

packages/opencensus-core/src/trace/instrumentation/types.ts
```diff
@@ -24,31 +24,45 @@ export interface Plugin {
...
   /** Method to disable the instrumentation  */
   disable(): void;
 }
 
+export type PluginConfig = {
+  // tslint:disable-next-line:no-any
+  [key: string]: any;
+};
+
+export type NamedPluginConfig = {
```
justindsmith: Should we have:
```
module: string;
config: PluginConfig;
```
That way there's no confusion over what is the module name and what the config is. There's no room for name collision.

packages/opencensus-core/src/trace/instrumentation/types.ts
```diff
@@ -24,31 +24,45 @@ export interface Plugin {
...
+}&PluginConfig;
 
 /**
  * Type PluginNames: each key should be the name of the module to trace,
  * and its value should be the name of the package  which has the
  * plugin implementation.
  */
 export type PluginNames = {
-  [pluginName: string]: string;
+  [pluginName: string]: string|PluginConfig;
```
justindsmith: Should this be `string|NamedPluginConfig` instead? That will allow typescript to know that `module` needs to be defined in the config object. Right now there's no compiler check that verifies that property exists. (Also see the object structure comment above.)

whs:
I'm having problem with http2. It seems that the original URL is not recorded in the session object, yet our interface requires the full URI for outgoing requests.

I'm thinking of leaving http2 unsupported for now.

justindsmith:
> I'm having problem with http2. It seems that the original URL is not recorded in the session object, yet our interface requires the full URI for outgoing requests.
> 
> I'm thinking of leaving http2 unsupported for now.

I think that's fine for now as long as we document it. This is more of an advanced feature, so having some corner-cases is going to be expected.

justindsmith:
@whs I _think_ the failing tests are just code style problems. You should be able to run `npm run fix` in any of the packages that are failing and it will fix the code style problems for you.

whs:
Merged master and test passed for Stackdriver. Anything else blocking the merge?

whs:
Added!

PR #162, opened 2018-10-24
fix spanId format

### LINKED ISSUE #161: Invalid spanId error stackdriver-exporter
After updating to `0.0.5` I am getting this error when trying to sendTraces:

```
[error] 2018-10-24T09:24:58.218Z sendTrace error: Invalid value at 'traces.traces[0].spans[0].span_id' (TYPE_FIXED64), "125aee19018a78dd"
```

It seems that the type of the spanId is invalid. I believe the issue is because the spanId is a hex string instead of a [int64 string](https://developers.google.com/discovery/v1/type-format).

I think this error is caused by this update #126.

conord33:
I am not sure if this is the best place to change the spanId format. Another option would be to do it in the core/utils spanId generator, but I am not sure if other exporters or propagators expect the spanId to be in a certain format.

closes #161

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
+++ b/packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
@@ -18,6 +18,7 @@ import {Exporter, ExporterBuffer, ExporterConfig, RootSpan, Span, SpanContext} f
 import {logger, Logger} from '@opencensus/core';
 import {auth, JWT} from 'google-auth-library';
 import {google} from 'googleapis';
+import {hexToDec} from 'hex2dec';
 
 import {StackdriverExporterOptions, TracesWithCredentials, TranslatedSpan, TranslatedTrace} from './types';
 
@@ -87,7 +88,7 @@ export class StackdriverTraceExporter implements Exporter {
     return {
       name: span.name,
       kind: 'SPAN_KIND_UNSPECIFIED',
-      spanId: span.id,
+      spanId: hexToDec(span.id),
       startTime: span.startTime,
       endTime: span.endTime,
       labels: Object.keys(span.attributes)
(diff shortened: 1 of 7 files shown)


conord33:
Yeah will do

conord33:
Looks like this would also be fixed by upgrading to the V2 cloud trace api instead of using V1.

conord33:
@isaikevych I updated the tests. Let me know what you think when you get a chance.

packages/opencensus-exporter-stackdriver/test/test-stackdriver-cloudtrace.ts
```diff
@@ -161,8 +159,16 @@ describe('Stackdriver Trace Exporter', function() {
...
-        nocks.oauth2((body) => true);
-        nocks.patchTraces(PROJECT_ID, null, null, false);
+        nocks.oauth2(body => true);
+        nocks.patchTraces(
+            PROJECT_ID, (body: {traces: TranslatedTrace[]}): boolean => {
+              assert.strictEqual(body.traces.length, 1);
+              const {spans} = body.traces[0];
+              assert.strictEqual(spans.length, 2);
+              assert.ok(spans.every(
+                  span => RegExp('^[0-9]{15,20}$').test(span.spanId)));
```
isaikevych: Could you please use the same package for verify in test?

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
```diff
@@ -87,7 +88,7 @@ export class StackdriverTraceExporter implements Exporter {
     return {
       name: span.name,
       kind: 'SPAN_KIND_UNSPECIFIED',
-      spanId: span.id,
+      spanId: hexToDec(span.id),
```
isaikevych: Can we use builtin function instead?

isaikevych:
Thank you

angelomichel:
Can we have a new release fixing the original issue (Invalid spanId error stackdriver-exporter, which seems to be resolved with this PR)?

We're running into the same issue.

PR #169, opened 2018-10-30
instrumentation-http: Parse URL instead of relying on pathname

whs:
instrumentation-http rely on `options.pathname` for outgoing request to name the span. However, several libraries are passing options directly to http.request which [save it as is](https://github.com/nodejs/node/blob/0d8021e5a4cf0a6aa3a700a361f6d42c2894f2ba/lib/_http_client.js#L90).

Affected libraries include:

- [node-fetch](https://github.com/bitinn/node-fetch) 1.x.x (can be solved by upgrading to 2.x.x)
- [request](https://github.com/request/request)

To reproduce, simply setup a tracer and fire an outgoing request with request. The trace will have the name of `GET undefined`.

This PR ensure that we parse every option for pathname instead of relying that it exists.

### DIFF
--- a/packages/opencensus-instrumentation-http/src/http.ts
+++ b/packages/opencensus-instrumentation-http/src/http.ts
@@ -212,7 +212,7 @@ export class HttpPlugin extends BasePlugin {
             return original.apply(this, arguments);
           }
 
-          pathname = url.parse(options.href).pathname;
+          pathname = url.parse(options.path).pathname;
         }
 
         const request = original.apply(this, arguments);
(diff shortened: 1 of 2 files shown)


packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -209,6 +211,8 @@ export class HttpPlugin extends BasePlugin {
                 'header with "x-opencensus-outgoing-request" - do not trace');
             return original.apply(this, arguments);
           }
+
+          pathname = url.parse(options.path).pathname;
```
OsvaldoRosado: You might want to do `options.pathname` || `url.parse(options.path).pathname`. [According to Node docs][1], the options parameter can be a WHATWG URL object, which [won't have a `path` field but will have `pathname`][2].

[1]: https://nodejs.org/api/http.html#http_http_request_options_callback
[2]: https://nodejs.org/api/url.html#url_the_whatwg_url_api

whs:
Fixed, please check

PR #173, opened 2018-11-01
Adds monitored resource to [core]

isaikevych:
Monitored resource for https://github.com/census-instrumentation/opencensus-node/issues/130


packages/opencensus-core/src/common/monitored-resource/gcp-metadata-config.ts
```diff
@@ -0,0 +1,138 @@
...
+          return Promise.resolve(GcpMetadataConfig.metadata);
+        });
+    return GcpMetadataConfig.promise;
+  }
+
+  private static getAttributes() {
+    let attributes;
+    if (GcpMetadataConfig.KUBERNETES_SERVICE_HOST in process.env) {
+      attributes = monitoredResourceAttributes.GKE;
+    } else {
```
songy23: This is not true. You'll also need to check if `instance-id` exists in the metadata to determine if application is running on GCE.

packages/opencensus-core/src/common/monitored-resource/gcp-metadata-config.ts
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  /**
+   * Fetches the requested instance metadata entry.
+   * @param name Attribute name relative to the computeMetadata/v1 prefix
+   */
+  static getAttribute(name: string): Promise<string> {
+    const options = {
+      host: GcpMetadataConfig.URL,
+      path: '/' + monitoredResourceAttributes.GKE[name],
```
songy23: Why is this always using GKE attributes?

packages/opencensus-core/src/common/monitored-resource/gcp-metadata-config.ts
isaikevych: > This is not true. You'll also need to check if `instance-id` exists in the metadata to determine if application is running on GCE.

This method is called just when we have instance-id

packages/opencensus-core/src/common/monitored-resource/gcp-metadata-config.ts
isaikevych: Done

isaikevych: > Why is this always using GKE attributes?

@songy23 Done

packages/opencensus-core/src/common/monitored-resource/monitored-resource.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Represents aws_ec2_instance type monitored resource.
+ * For definition refer to
+ *  https://cloud.google.com/monitoring/api/resources#tag_aws_ec2_instance
+ */
+export class AwsMonitoredResource implements types.MonitoredResource {
+  type = types.MonitoredResources.AWS_EC2_INSTANCE;
+
+  getLabels() {
+    const aws = new AwsIdentityDocumentUtils();
+    return aws.getMetadata();
```
songy23: Please only extract the needed information from the AWS metadata (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html). The only things needed are account id, instance id and region.

isaikevych: Done

packages/opencensus-core/src/common/monitored-resource/types.ts
```diff
@@ -0,0 +1,74 @@
+export enum MonitoredResources {
+  GCP_GCE_INSTANCE = 'gce_instance',
```
songy23: The type and attribute names should match [Specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/resource/Resource.md). See also [go constants](https://github.com/census-ecosystem/opencensus-go-resource/blob/master/resourcekeys/const.go)

packages/opencensus-core/src/common/monitored-resource/aws-identity-document-utils.ts
```diff
@@ -0,0 +1,105 @@
+/**
```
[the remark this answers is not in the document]
isaikevych: This new requirements blocked by [Resource API](https://github.com/census-instrumentation/opencensus-node/pull/193)

PR #175, opened 2018-11-03
Drops negative bucket bounds and negative values in record API

isaikevych:
Issue https://github.com/census-instrumentation/opencensus-node/issues/172

### DIFF
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -79,7 +79,8 @@ export class BaseView implements View {
    */
   constructor(
       name: string, measure: Measure, aggregation: AggregationType,
-      tagsKeys: string[], description: string, bucketBoundaries?: number[], logger = defaultLogger) {
+      tagsKeys: string[], description: string, bucketBoundaries?: number[],
+      logger = defaultLogger) {
     if (aggregation === AggregationType.DISTRIBUTION && !bucketBoundaries) {
       throw new Error('No bucketBoundaries specified');
     }
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -109,6 +109,12 @@ export class BaseView implements View {
       return;
     }
 
+    if (measurement.value < 0) {
+      this.logger.warn(
+          'Dropping values, value to record must be non-negative.');
+      return;
+    }
+
     // Checks if measurement has all tags in views
     for (const tagKey of this.columns) {
       if (!Object.keys(measurement.tags).some((key) => key === tagKey)) {
@@ -120,6 +126,7 @@ export class BaseView implements View {
     if (!this.rows[encodedTags]) {
       this.rows[encodedTags] = this.createAggregationData(measurement.tags);
     }
+
     Recorder.addMeasurement(this.rows[encodedTags], measurement);
   }
 
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -14,6 +14,9 @@
  * limitations under the License.
  */
 
+import * as logger from '../common/console-logger';
+import * as loggerTypes from '../common/types';
+
 import {Recorder} from './recorder';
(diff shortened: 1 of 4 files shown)


packages/opencensus-core/src/stats/view.ts
```diff
@@ -59,6 +62,8 @@ export class BaseView implements View {
   endTime: number;
   /** true if the view was registered */
   registered = false;
+  /** An object to log information to */
+  private logger: loggerTypes.Logger = logger.logger();
```
OsvaldoRosado: Should this be hardcoded? I've noticed in other classes in this repo the pattern seems to be to have a default logger but allow passing one in as an override.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -181,15 +193,55 @@ export class BaseView implements View {
...
-    return bucketBoundaries.map((boundary, boundaryIndex) => {
-      return {
-        count: 0,
-        lowBoundary: boundaryIndex ? boundary : -Infinity,
-        highBoundary: (boundaryIndex === bucketBoundaries.length - 1) ?
-            Infinity :
-            bucketBoundaries[boundaryIndex + 1]
-      };
-    });
+    const boundaries = bucketBoundaries.filter(Math.clz32);
```
OsvaldoRosado: This doesn't seem to catch all negative numbers. For example, this passes through `-Infinity`. Is that ok?

OsvaldoRosado: It also seems a little compute wasteful to do this `.filter` just for a console log. Can't the logging done during the reduce?

packages/opencensus-core/src/stats/view.ts
```diff
@@ -181,15 +193,55 @@ export class BaseView implements View {
...
+        this.logger.error('Bucket boundaries not sorted.');
+      }
+      if (current === next) {
+        this.logger.error('Bucket boundaries not unique.');
+      }
+    }
+  }
+
+  /**
+   * Creates empty buckeet boundary.
```
OsvaldoRosado: Nit: `buckeet` -> `bucket`

packages/opencensus-core/src/stats/view.ts
isaikevych: Fixed for current class, added [issue 177](https://github.com/census-instrumentation/opencensus-node/issues/177) for improve other parts

packages/opencensus-core/src/stats/view.ts
isaikevych: Improved. Thank you

packages/opencensus-core/src/stats/view.ts
isaikevych: Changed

packages/opencensus-core/src/stats/view.ts
```diff
@@ -104,6 +112,12 @@ export class BaseView implements View {
       return;
     }
 
+    if (measurement.value < 0) {
+      this.logger.warn(
+          'Dropping values, value to record must be non-negative.');
```
songy23: Maybe also log the value?

packages/opencensus-core/src/stats/view.ts
```diff
@@ -181,15 +196,57 @@ export class BaseView implements View {
...
+        const bucket = this.createBucket(boundary, nextBoundary, position, len);
+        accumulator.push(bucket);
+      } else {
+        negative++;
+      }
+      return accumulator;
+    }, []);
+    if (negative) {
+      this.logger.warn(`Dropping ${
+          negative} negative bucket boundaries, the values must be strictly > 0.`);
```
songy23: ditto

isaikevych: > ditto

This is what implemented in java [BucketBoundaries.java line 84](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/stats/BucketBoundaries.java#L84)

songy23: Ignore this comment - just noticed "${negative}" is already in the log.

packages/opencensus-core/src/stats/view.ts
isaikevych: Done

PR #185, opened 2018-11-12
Rename prefix name for [StackdriverStatsExporter]

isaikevych:
Uses `require` instead of `import` for `hex2dec` package due to type package for `hex2dec` will be available

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
+++ b/packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
@@ -45,8 +45,8 @@ export class StackdriverStatsExporter implements StatsEventListener {
         options.period :
         StackdriverStatsExporter.PERIOD;
     this.projectId = options.projectId;
-    this.metricPrefix = options.metricPrefix ||
-        StackdriverStatsExporter.CUSTOM_OPENCENSUS_DOMAIN;
+    this.metricPrefix =
+        options.prefix || StackdriverStatsExporter.CUSTOM_OPENCENSUS_DOMAIN;
     this.logger = options.logger || logger.logger();
     this.viewToUpload = {};
 
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,7 +13,7 @@
     "JSONStream": {
       "version": "1.3.5",
       "resolved": "https://registry.npmjs.org/JSONStream/-/JSONStream-1.3.5.tgz",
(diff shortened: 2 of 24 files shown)


packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
```diff
@@ -45,8 +45,8 @@ export class StackdriverStatsExporter implements StatsEventListener {
         options.period :
         StackdriverStatsExporter.PERIOD;
     this.projectId = options.projectId;
-    this.metricPrefix = options.metricPrefix ||
-        StackdriverStatsExporter.CUSTOM_OPENCENSUS_DOMAIN;
+    this.metricPrefix =
+        options.prefix || StackdriverStatsExporter.CUSTOM_OPENCENSUS_DOMAIN;
```
mayurkale22: Can you please add tests for this? Something like: https://github.com/census-instrumentation/opencensus-java/blob/master/exporters/stats/stackdriver/src/test/java/io/opencensus/exporter/stats/stackdriver/StackdriverExporterWorkerTest.java#L255

songy23:
Please split the changes on `package-lock.json` to another PR. This PR is too large.

isaikevych:
> Please split the changes on `package-lock.json` to another PR. This PR is too large.
Its recommended to push package-lock files According to npm help package-lock.json:

`package-lock.json is automatically generated for any operations where npm modifies either the node_modules tree, or package.json. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, regardless of intermediate dependency updates.

This file is intended to be committed into source repositories, and serves various purposes:

Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.

Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.

To facilitate greater visibility of tree changes through readable source control diffs.

And optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.

One key detail about package-lock.json is that it cannot be published, and it will be ignored if found in any place other than the toplevel package. It shares a format with npm-shrinkwrap.json(5), which is essentially the same file, but allows publication. This is not recommended unless deploying a CLI tool or otherwise using the publication process for producing production packages.`

packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
isaikevych: Done

PR #196, opened 2018-11-20
Metrics: use `number` for timestamp instead of interface

mayurkale22:


mayurkale22:
Are you asking about unit tests or [unit](https://github.com/census-instrumentation/opencensus-node/blob/743c98532d1743f6e2709022f8c92454c82b9eb6/packages/opencensus-core/src/metrics/export/types.ts#L47) type in MetricDescriptor?

bogdandrutu:
@mayurkale22 this is for trace. And timestamp as a number should have a unit "us" or "ns" from the epoch. I am a bit lost in this PR logic.

bogdandrutu:
https://www.geeksforgeeks.org/javascript-date-now/ so it seems that the timestamp you are proposing is millis since epoch. Do we have a way to get better timestamps (ns or us granularity)?

isaikevych:
> https://www.geeksforgeeks.org/javascript-date-now/ so it seems that the timestamp you are proposing is millis since epoch. Do we have a way to get better timestamps (ns or us granularity)?

@mayurkale22 I dont know what are you going to achieve by this changes but anyway
@bogdandrutu nodejs has process.hrtime() which returns high-resolution real time in [seconds, nanoseconds] array then we had plans to use it and that is why protobuf generated Timestamp interface with seconds and nanos

PR #197, opened 2018-11-20
Exporter/Stats/Stackdriver: Fix MetricDescriptors ValueType

### LINKED ISSUE #178: StackdriverStatsExporter: valueType to string
Change valueType to use string instead of number

mayurkale22:
Fixes #178

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/types.ts
+++ b/packages/opencensus-exporter-stackdriver/src/types.ts
@@ -72,13 +72,13 @@ export enum MetricKind {
 }
 
 export enum ValueType {
-  VALUE_TYPE_UNSPECIFIED,
-  BOOL,
-  INT64,
-  DOUBLE,
-  STRING,
-  DISTRIBUTION,
-  MONEY
+  VALUE_TYPE_UNSPECIFIED = 'VALUE_TYPE_UNSPECIFIED',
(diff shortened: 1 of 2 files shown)


packages/opencensus-exporter-stackdriver/src/types.ts
```diff
@@ -72,13 +72,13 @@ export enum MetricKind {
...
-  STRING,
-  DISTRIBUTION,
-  MONEY
+  VALUE_TYPE_UNSPECIFIED = 'VALUE_TYPE_UNSPECIFIED',
+  BOOL = 'BOOL',
+  INT64 = 'INT64',
+  DOUBLE = 'DOUBLE',
+  STRING = 'STRING',
+  DISTRIBUTION = 'DISTRIBUTION',
+  MONEY = 'MONEY'
```
bogdandrutu: I don't think we support "Money", "String" and "Bool". Do we still need them here?

mayurkale22: I don't think we need that, looks like previously copied from https://cloud.google.com/monitoring/api/ref_v3/rest/v3/projects.metricDescriptors#ValueType list.

mayurkale22: Removed.

PR #200, opened 2018-11-26
Enforce "noUnusedLocals" : Report errors on unused locals.

mayurkale22:
Add ```--noUnusedLocals``` Compiler Options to disallows unused imports, variables, functions and
private class members.

### DIFF
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -16,10 +16,9 @@
 
 import * as defaultLogger from '../common/console-logger';
 import * as loggerTypes from '../common/types';
-
 import {BucketBoundaries} from './bucket-boundaries';
 import {Recorder} from './recorder';
-import {AggregationData, AggregationMetadata, AggregationType, CountData, DistributionData, LastValueData, Measure, Measurement, MeasureType, SumData, Tags, View} from './types';
+import {AggregationData, AggregationType, Measure, Measurement, Tags, View} from './types';
 
 const RECORD_SEPARATOR = String.fromCharCode(30);
 const UNIT_SEPARATOR = String.fromCharCode(31);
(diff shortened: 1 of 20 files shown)


packages/opencensus-core/src/stats/view.ts
```diff
@@ -64,6 +63,7 @@ export class BaseView implements View {
   /** true if the view was registered */
   registered = false;
   /** An object to log information to */
+  // @ts-ignore
```
isaikevych: Why we disabling if we want to avoid redundant definitions?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -30,6 +29,7 @@ export abstract class SpanBase implements types.Span {
   /** Indicates if this span was ended */
   private endedLocal = false;
   /** Indicates if this span was forced to end */
+  // @ts-ignore
```
isaikevych: Why we disabling if we want to avoid redundant definitions?

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -40,6 +38,7 @@ export class CoreTracer implements types.Tracer {
   /** A list of end span event listeners */
   private eventListenersLocal: types.SpanEventListener[] = [];
   /** A list of ended root spans */
+  // @ts-ignore
```
isaikevych: Why we disabling if we want to avoid redundant definitions?

mayurkale22: As discussed, I added to suppress errors. I am not completely aware of CoreTracer properties, hence didnt remove that variable. Anyways, I will remove it in next PR.

isaikevych: > As discussed, I added to suppress errors. I am not completely aware of CoreTracer properties, hence didnt remove that variable. Anyways, I will remove it in next PR.

How users will use Typescript for their projects in this case?

justindsmith: Looks like each case it's a private unused variable, so we should clean them up. Let's just remove them in the next PR like @mayurkale22 suggested. They aren't api breaking to remove since they're private.

PR #203, opened 2018-11-27
Add metric registry and validations util

mayurkale22:


packages/opencensus-core/test/test-metric-registry.ts
```diff
@@ -0,0 +1,151 @@
...
+const UNIT = MeasureUnit.UNIT;
+const LABEL_KEYS: LabelKey[] = [{key: 'code', description: 'desc'}];
+const LABEL_KEYS_WITH_NULL: LabelKey[] =
+    [{key: 'code', description: 'desc'}, null];
+
+const registry = new MetricRegistry();
+
+describe('addInt64Gauge', () => {
+  it('should throw an error when the name is null', () => {
+    assert.throws(() => {
```
isaikevych: Could you please add coverage for empty string, 0 and undefined?
The same for other tests in this PR

packages/opencensus-core/src/metrics/metric-registry.ts
```diff
@@ -0,0 +1,104 @@
...
+   *
+   * @param {string} name The name of the metric.
+   * @param {string} description The description of the metric.
+   * @param {MeasureUnit} unit The unit of the metric.
+   * @param {LabelKey[]} labelKeys The list of the label keys.
+   */
+  addDerivedInt64Gauge(
+      name: string, description: string, unit: MeasureUnit,
+      labelKeys: LabelKey[]): void {
+    checkNotNull(name, 'name');
```
isaikevych: Can we use just logical OR instead? 
The same for lines below in this PR

packages/opencensus-core/src/metrics/metric-registry.ts
```diff
@@ -0,0 +1,104 @@
...
+    checkNotNull(description, 'description');
+    checkNotNull(unit, 'unit');
+    checkListElementNotNull(checkNotNull(labelKeys, 'labelKeys'), 'labelKey');
```
isaikevych: you can simplify it by using `Array.every()`
The same for lines below in this PR

packages/opencensus-core/test/test-metric-registry.ts
mayurkale22: Added tests for empty string and undefined values.

packages/opencensus-core/src/metrics/metric-registry.ts
mayurkale22: ```Array.every()``` good option, but it will return boolean values, I have to add another condition to check returned value and throw exception. WDYT?

packages/opencensus-core/src/metrics/metric-registry.ts
mayurkale22: I think this is good for readability and to throw meaningful exception (using missing param name).

packages/opencensus-core/src/metrics/metric-registry.ts
isaikevych: Right - you can use it with `Array.every()` and its builtin function

packages/opencensus-core/src/metrics/metric-registry.ts
isaikevych: OR is also readable than non pure function

packages/opencensus-core/test/test-metric-registry.ts
isaikevych: What about `0` and `'0'`?

packages/opencensus-core/src/common/validations.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+/**
+ * Ensures the truth of an expression involving one or more parameters to the
+ * calling method.
+ *
+ * @param  {boolean} expression A boolean expression.
+ * @param  {string} errorMessage The exception message to use if the check fails.
+ */
+export function checkArgument(expression: boolean, errorMessage: string) {
```
justindsmith: I don't see this function used anywhere, can we remove it?

packages/opencensus-core/src/common/validations.ts
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Ensures that an object reference passed as a parameter to the calling
+ * method is not null.
+ *
+ * @param {T} reference An object reference.
+ * @param {string} errorMessage The exception message to use if the check fails.
+ * @returns {T} An object reference.
+ */
+export function checkNotNull<T>(reference: T, errorMessage: string): T {
```
justindsmith: If these are validators, can we call them `validateNotNull`, etc?

justindsmith: Also, this will throw an error if the `reference` parameter is `0` or `''` (empty string), is that expected? If so this should be documented, but in this case I'd suggest we do a more strict `reference === null || reference === undefined` conditional.

packages/opencensus-core/src/common/validations.ts
```diff
@@ -0,0 +1,57 @@
...
+  }
+}
+
+/**
+ * Throws an Error if any of the list elements is null.
+ *
+ * @param {T} list The argument list to check for null.
+ * @param {string} errorMessage The exception message to use if the check fails.
+ */
+export function checkListElementNotNull<T>(list: T[], errorMessage: string) {
```
justindsmith: Typically "lists" are referred to as "arrays". Can we rename this `validateArrayElementsNotNull`?

packages/opencensus-core/src/common/validations.ts
```diff
@@ -0,0 +1,57 @@
...
+  for (const element of list) {
+    if (!element) {
```
justindsmith: Same problem as the other function with conditional checking. Either improve or make sure documentation explains behavior of `0` and empty string.

packages/opencensus-core/test/test-metric-registry.ts
justindsmith: Typescript will enforce type checking, preventing "0" (unless someone uses an `any` forced type), but if someone is using this class in pure javascript you can't enforce types and could receive any input.

I guess the question is really if this is an external facing interface or an internal only interface where you can expect that the inputs have already been sanitized and validated.

packages/opencensus-core/src/common/validations.ts
isaikevych: Also, its possible to use builtin function instead https://github.com/census-instrumentation/opencensus-node/pull/203/files#r236829149 Array.every()

packages/opencensus-core/src/common/validations.ts
mayurkale22: Removed. This is required for my next PR, I tried to split PR (Gauges API) into smaller ones (this is the first installment of that).

packages/opencensus-core/src/common/validations.ts
mayurkale22: Done

packages/opencensus-core/src/common/validations.ts
mayurkale22: Changed to ```Array.forEach```.

packages/opencensus-core/src/common/validations.ts
mayurkale22: Done.

mayurkale22:
@justindsmith This registry is added to create and manage an application's set of metrics (like gauge, counter), this will also have a metric collector interface. Instead of one huge PR, I tried to split stuffs in smaller PRs. Next PRs, 

- Add Gauges API and Plug-in gauges into the registry.
- Add derived gauges API and Plug-in gauges into the registry.
- Add Metric Collector interface.
- Register Metric-Registry with MetricProducer.

packages/opencensus-core/src/common/validations.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Validates that an array passed as a parameter doesn't contain null element.
+ *
+ * @param {T} list The argument list to check for null.
+ * @param {string} errorMessage The exception message to use if the check fails.
+ */
+export function validateArrayElementsNotNull<T>(
+    array: T[], errorMessage: string) {
+  array.forEach((element) => {
```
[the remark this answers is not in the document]
mayurkale22: Done, Thanks 👍

mayurkale22:
Thank you @justindsmith @isaikevych for reviews.

isaikevych:
> @mayurkale22 I know this has now been merged, and I _definitely_ like the use of smaller PRs (seriously, thank you! ), but maybe we should create a branch for these kinds of features - while still doing small PRs along the way - and then once the feature is stable and ready merge that back into the master branch. That way any iteration in the feature isn't performed in the master and add to the commit log as releases go out while the feature is still in progress.
> 
> What do you think about that?
> 
> /cc @isaikevych

@justindsmith Good idea, thank you

PR #207, opened 2018-11-29
Add Double and INT64 Gauges APIs

mayurkale22:
The Gauges API work is split into multiple PRs. This is the second installment.

Completed:
**- Add metric registry and validations util** (https://github.com/census-instrumentation/opencensus-node/pull/203)

Current:
- Add Double and INT64 Gauges APIs

Next in line:
- Plug-in gauges into the registry.
- Add derived gauges API and Plug-in gauges into the registry.
- Add Metric Collector interface.
- Register Metric-Registry with MetricProducer.

### DIFF
--- a/packages/opencensus-core/src/common/validations.ts
+++ b/packages/opencensus-core/src/common/validations.ts
@@ -37,8 +37,9 @@ export function validateNotNull<T>(reference: T, errorMessage: string): T {
  */
 export function validateArrayElementsNotNull<T>(
     array: T[], errorMessage: string) {
-  if (array.every(
-          element => element !== null || typeof element !== 'undefined')) {
+  const areAllNotNullOrUndefined = array.every(
+      element => element !== null && typeof element !== 'undefined');
+  if (!areAllNotNullOrUndefined) {
     throw new Error(`${errorMessage} elements should not be a NULL`);
   }
 }
--- a/packages/opencensus-core/src/common/validations.ts
+++ b/packages/opencensus-core/src/common/validations.ts
@@ -37,9 +37,9 @@ export function validateNotNull<T>(reference: T, errorMessage: string): T {
  */
 export function validateArrayElementsNotNull<T>(
     array: T[], errorMessage: string) {
-  const areAllNotNullOrUndefined = array.every(
+  const areAllDefined = array.every(
       element => element !== null && typeof element !== 'undefined');
-  if (!areAllNotNullOrUndefined) {
+  if (!areAllDefined) {
     throw new Error(`${errorMessage} elements should not be a NULL`);
   }
 }
--- a/packages/opencensus-core/src/metrics/gauges/gauge.ts
+++ b/packages/opencensus-core/src/metrics/gauges/gauge.ts
@@ -28,6 +28,20 @@ export class Gauge implements types.Meter {
   private defaultLabelValues: LabelValue[];
   private registeredPoints: Map<string, types.Point> = new Map();
 
+  private static readonly LABEL_VALUE = 'labelValue';
+  private static readonly LABEL_VALUES = 'labelValues';
+  private static readonly ERROR_MESSAGE_INVALID_SIZE =
+      'Label Keys and Label Values don\'t have same size';
+
+  /**
+   * Constructs a new Gauge instance.
+   *
+   * @param {string} name The name of the metric.
+   * @param {string} description The description of the metric.
+   * @param {string} unit The unit of the metric.
+   * @param {MetricDescriptorType} type The type of metric.
+   * @param {LabelKey[]} labelKeys The list of the label keys.
+   */
   constructor(
       name: string, description: string, unit: string,
       type: MetricDescriptorType, readonly labelKeys: LabelKey[]) {
@@ -49,7 +63,7 @@ export class Gauge implements types.Meter {
    */
   getOrCreateTimeSeries(labelValues: LabelValue[]): types.Point {
     validateArrayElementsNotNull(
-        validateNotNull(labelValues, 'labelValues'), 'labelValue');
+        validateNotNull(labelValues, Gauge.LABEL_VALUES), Gauge.LABEL_VALUE);
     return this.registerTimeSeries(labelValues);
   }
 
@@ -71,7 +85,7 @@ export class Gauge implements types.Meter {
    * @param {LabelValue[]} labelValues The list of label values.
    */
   removeTimeSeries(labelValues: LabelValue[]): void {
-    validateNotNull(labelValues, 'labelValues');
+    validateNotNull(labelValues, Gauge.LABEL_VALUES);
     this.registeredPoints.delete(hashLabelValues(labelValues));
   }
 
@@ -98,7 +112,7 @@ export class Gauge implements types.Meter {
       return this.registeredPoints.get(hash);
     }
     if (this.labelKeysLength !== labelValues.length) {
-      throw new Error('Label Keys and Label Values don\'t have same size');
+      throw new Error(Gauge.ERROR_MESSAGE_INVALID_SIZE);
     }
 
     const point = new PointEntry(labelValues);
--- a/packages/opencensus-core/src/metrics/gauges/gauge.ts
+++ b/packages/opencensus-core/src/metrics/gauges/gauge.ts
@@ -22,7 +22,7 @@ import {hashLabelValues, initializeDefaultLabels} from '../utils';
 /**
  * Gauge metric
  */
-export class Gauge {
+export class Gauge implements types.Meter {
   private readonly metricDescriptor: MetricDescriptor;
   private labelKeysLength: number;
   private defaultLabelValues: LabelValue[];
@@ -115,13 +115,13 @@ export class Gauge {
     if (this.registeredPoints.size === 0) {
       return null;
     }
-    const hrtimeLocal = process.hrtime();
-    const timeSeriesList: TimeSeries[] = new Array();
-    this.registeredPoints.forEach((point: types.Point) => {
-      timeSeriesList.push(point.getTimeSeries(
-          {seconds: hrtimeLocal[0], nanos: hrtimeLocal[1]}));
-    });
-    return {descriptor: this.metricDescriptor, timeseries: timeSeriesList};
+    const [seconds, nanos] = process.hrtime();
+    return {
+      descriptor: this.metricDescriptor,
+      timeseries: Array.from(
+          this.registeredPoints,
+          ([_, point]) => point.getTimeSeries({seconds, nanos}))
+    };
   }
 }
 
@@ -162,8 +162,7 @@ export class PointEntry implements types.Point {
   getTimeSeries(timestamp: Timestamp): TimeSeries {
     return {
       labelValues: this.labelValues,
-      points: [{value: this.value, timestamp}],
-      startTimestamp: null
+      points: [{value: this.value, timestamp}]
     };
   }
 }
--- a/packages/opencensus-core/src/metrics/gauges/gauge.ts
+++ b/packages/opencensus-core/src/metrics/gauges/gauge.ts
@@ -143,9 +143,10 @@ export class Gauge implements types.Meter {
  * The value of a single point in the Gauge.TimeSeries.
  */
 export class PointEntry implements types.Point {
+  private readonly labelValues: LabelValue[];
   private value = 0;
 
-  constructor(private readonly labelValues: LabelValue[]) {
+  constructor(labelValues: LabelValue[]) {
     this.labelValues = labelValues;
   }
 
--- a/packages/opencensus-core/src/metrics/gauges/gauge.ts
+++ b/packages/opencensus-core/src/metrics/gauges/gauge.ts
@@ -1,0 +1,169 @@
+/**
+ * Copyright 2018, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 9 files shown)


packages/opencensus-core/src/metrics/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (obj1.value > obj2.value) {
+        return 1;
+      }
+      if (obj1.value < obj2.value) {
+        return -1;
+      }
+      return 0;
+    });
+  }
+  const elems = [];
```
justindsmith: you can use `const elems = labelValues.map(lv => lv.value);` here to quickly get an array of the `value` property of each element in the array (instead of the `for` loop following this line).

packages/opencensus-core/src/metrics/gauges/gauge.ts
```diff
@@ -0,0 +1,169 @@
...
+   *
+   * @returns {Metric} The Metric.
+   */
+  getMetric(): Metric {
+    if (this.registeredPoints.size === 0) {
+      return null;
+    }
+    const hrtimeLocal = process.hrtime();
+    const timeSeriesList: TimeSeries[] = new Array();
+    this.registeredPoints.forEach((point: types.Point) => {
```
justindsmith: Instead of `new Array()` and then `.forEach` you can just use `.map`

packages/opencensus-core/src/metrics/gauges/types.ts
```diff
@@ -0,0 +1,50 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Metric, TimeSeries, Timestamp} from '../export/types';
+
+export interface Meter {
```
justindsmith: Is this used anywhere in the PR? Is it for a future feature?

packages/opencensus-core/test/test-utils.ts
```diff
@@ -0,0 +1,49 @@
...
+  });
+  it('should return hash for Value and null', () => {
+    const hash = hashLabelValues(LABEL_VALUES_WITH_NULL);
+    assert.deepStrictEqual(hash, ',200');
+  });
+  it('should return same hash for interchanged labels', () => {
+    assert.deepStrictEqual(
+        hashLabelValues(LABEL_VALUES_1), hashLabelValues(LABEL_VALUES_2));
+  });
+});
```
justindsmith: might want to add a quick "empty array" check here, you cover it in your function impl.

packages/opencensus-core/src/metrics/utils.ts
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Returns default label values.
+ *
+ * @param  {number} count The number of label values.
+ * @returns {LabelValue[]} The list of the label values.
+ */
+export function initializeDefaultLabels(count: number): LabelValue[] {
+  const UNSET_LABEL_VALUE: LabelValue = {value: null};
+  const defaultLabels: LabelValue[] = new Array(count);
```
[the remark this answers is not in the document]
mayurkale22: SGTM, Done

packages/opencensus-core/src/metrics/utils.ts
mayurkale22: Sure, Done.

packages/opencensus-core/src/metrics/gauges/gauge.ts
mayurkale22: Done

packages/opencensus-core/test/test-utils.ts
mayurkale22: Make Sense, Added.

packages/opencensus-core/src/metrics/gauges/types.ts
mayurkale22: Good catch, implemented by Gauge class. Also, this interface is required for DerivedGauge class and metric collector/producer (in future).

packages/opencensus-core/src/metrics/gauges/gauge.ts
```diff
@@ -0,0 +1,170 @@
...
+   *
+   * It is recommended to keep a reference to the Point instead of always
+   * calling this method for manual operations.
+   *
+   * @param {LabelValue[]} labelValues The list of the label values.
+   * @returns {types.Point} The value of single gauge.
+   */
+  getOrCreateTimeSeries(labelValues: LabelValue[]): types.Point {
+    validateArrayElementsNotNull(
+        validateNotNull(labelValues, 'labelValues'), 'labelValue');
```
isaikevych: Should we show to user some message in case if its `null`? The same question for other `trow new Error` in this PR

packages/opencensus-core/src/metrics/gauges/gauge.ts
```diff
@@ -0,0 +1,170 @@
...
+    if (this.registeredPoints.size === 0) {
+      return null;
+    }
+    const hrtimeLocal = process.hrtime();
+    return {
+      descriptor: this.metricDescriptor,
+      timeseries: Array.from(
+          this.registeredPoints,
+          ([_, point]) => point.getTimeSeries(
+              {seconds: hrtimeLocal[0], nanos: hrtimeLocal[1]}))
```
isaikevych: you can destructure instead of calling by index

packages/opencensus-core/src/metrics/gauges/gauge.ts
```diff
@@ -0,0 +1,170 @@
...
+   * Returns the TimeSeries with one or more Point.
+   *
+   * @param {Timestamp} timestamp The time at which the gauge is recorded.
+   * @returns {TimeSeries} The TimeSeries.
+   */
+  getTimeSeries(timestamp: Timestamp): TimeSeries {
+    return {
+      labelValues: this.labelValues,
+      points: [{value: this.value, timestamp}],
+      startTimestamp: null
```
isaikevych: Why its `null`

packages/opencensus-core/src/metrics/gauges/gauge.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+}
+
+/**
+ * The value of a single point in the Gauge.TimeSeries.
+ */
+export class PointEntry implements types.Point {
+  private value = 0;
+
+  constructor(private readonly labelValues: LabelValue[]) {
```
isaikevych: my guess that `private readonly` should not be as argument

packages/opencensus-core/src/metrics/utils.ts
```diff
@@ -0,0 +1,52 @@
...
+ * @param  {LabelValue[]} labelValues The list of the label values.
+ * @returns {string} The hashed label values string.
+ */
+export function hashLabelValues(labelValues: LabelValue[]): string {
+  if (labelValues.length === 0) {
+    return '';
+  }
+  if (labelValues.length > 1) {
+    labelValues = labelValues.sort((obj1, obj2) => {
+      if (obj1.value > obj2.value) {
```
isaikevych: you can simplify it by using `+-` operators instead of this `ifs`

packages/opencensus-core/src/metrics/gauges/gauge.ts
mayurkale22: ```start_timestamp```: Must be present for cumulative metrics only. https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto#L133-L137

packages/opencensus-core/src/metrics/gauges/gauge.ts
mayurkale22: @justindsmith any thoughts?

packages/opencensus-core/src/metrics/gauges/gauge.ts
mayurkale22: Done, thanks.

packages/opencensus-core/src/metrics/utils.ts
mayurkale22: SGTM, done.

mayurkale22:
@isaikevych I see 100% coverage for ```gauges```, ```utils``` and ```validations```. Let me know if I miss anything for tests.

packages/opencensus-core/src/metrics/gauges/gauge.ts
isaikevych: Is it optional param? if yes - please change it in type

isaikevych:
@mayurkale22 Could you please show me lines where are you defining Gauge, Point instance?

packages/opencensus-core/src/metrics/gauges/gauge.ts
mayurkale22: Done

mayurkale22:
> @mayurkale22 Could you please show me lines where you defining Gauge, Point instance?

https://github.com/census-instrumentation/opencensus-node/pull/207/files#diff-89ce3df8e0c8e56beb55435f6596c7ddR104

packages/opencensus-core/src/metrics/utils.ts
isaikevych: still the same

mayurkale22: I changed to ```+(obj1.value > obj2.value) || -(obj1.value < obj2.value) || 0;```  https://github.com/census-instrumentation/opencensus-node/pull/207/files#diff-4ecd32f36720fa9f740ab01cc93ee360R31. Let me know if something else is expected.

isaikevych:
```suggestion
      export function hashLabelValues(labelValues: LabelValue[]): string {
        return labelValues.map(lv => lv.value).sort().join(this.COMMA_SEPARATOR)
      }
```
You can use + operator if you need desc ordered

packages/opencensus-core/src/metrics/gauges/gauge.ts
justindsmith: Did a quick search, looks like we use a mix of both conventions throughout the codebase (fun!).

I'm not sure I would put a blanket "always" or "never" on this approach, but I think readability goes down when you have a mix of class-level and constructor-level properties defined that are both associated with the class. A developer now needs to look at _both_ places to see the complete set of properties. So I would say that if _all_ the properties of a class are passed in through the constructor we can use a visibility modifier to attach it to the class, but if we're defining other properties we shouldn't use that approach.

If we go by the above then in this case we should make it a class-level definition and not use the constructor-visiblity-modifier approach.

What do you all think of that? What are your thoughts on this?

mayurkale22: Just to make sure I understand your point.

> if all the properties of a class are passed in through the constructor we can use a visibility modifier to attach it to the class

something like this?
```
class PointEntry implements types.Point {
    private readonly labelValues: LabelValue[];

    constructor (labelValues: LabelValue[]) {
        this.labelValues = labelValues;
    }
}
```

packages/opencensus-core/src/metrics/utils.ts
mayurkale22: done, thanks

isaikevych: Please remove this redundant
```suggestion
     if (labelValues.length === 0) {
       return '';
    }
```

packages/opencensus-core/src/common/validations.ts
```diff
@@ -37,8 +37,9 @@ export function validateNotNull<T>(reference: T, errorMessage: string): T {
  */
 export function validateArrayElementsNotNull<T>(
     array: T[], errorMessage: string) {
-  if (array.every(
-          element => element !== null || typeof element !== 'undefined')) {
+  const areAllNotNullOrUndefined = array.every(
+      element => element !== null && typeof element !== 'undefined');
+  if (!areAllNotNullOrUndefined) {
```
isaikevych: would you mind to rename it to `areAllDefined`?
in this case it will be more simple to read

packages/opencensus-core/src/metrics/utils.ts
mayurkale22: Done.

packages/opencensus-core/src/common/validations.ts
mayurkale22: SGTM, Done.

packages/opencensus-core/test/test-gauge.ts
```diff
@@ -0,0 +1,407 @@
...
+    description: METRIC_DESCRIPTION,
+    unit: UNIT,
+    type: GAUGE_INT64,
+    labelKeys: LABEL_KEYS
+  };
+
+  beforeEach(() => {
+    instance = new Gauge(
+        METRIC_NAME, METRIC_DESCRIPTION, UNIT, GAUGE_INT64, LABEL_KEYS);
+    process.hrtime = () => [1000, 1e7];
```
isaikevych: You redefining `process.hrtime` Its builtin function of global nodejs object, other functionality can use it.
The best practice is to restore the state.

mayurkale22: Done, here https://github.com/census-instrumentation/opencensus-node/pull/207/commits/164369cf66ad168eaffdbf1422f1b0e0f03c7ba5 
Good catch!

mayurkale22:
@justindsmith friendly ping?

packages/opencensus-core/src/metrics/gauges/gauge.ts
justindsmith: Yeah, in that case I think it's okay to just use the visibility modifier in the constructor since that value is the only class-level property.

mayurkale22:
Thanks @justindsmith @isaikevych for reviews.

PR #208, opened 2018-11-29
Drops whole measurement list if some of value is negative

isaikevych:
Issue [206](https://github.com/census-instrumentation/opencensus-node/issues/206)

### DIFF
--- a/packages/opencensus-core/src/stats/stats.ts
+++ b/packages/opencensus-core/src/stats/stats.ts
@@ -115,8 +115,9 @@ export class Stats {
   }
 
   /**
-   * Virifies whether all measurements has positive value
+   * Verifies whether all measurements has positive value
    * @param measurements A list of measurements
+   * @returns {boolean} Whether values is positive
    */
   private hasNegativeValue(measurements: Measurement[]): boolean {
     return measurements.some(measurement => measurement.value < 0);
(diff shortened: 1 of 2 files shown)


packages/opencensus-core/src/stats/stats.ts
```diff
@@ -114,18 +114,26 @@ export class Stats {
     return {name, unit, type: MeasureType.INT64, description};
   }
 
+  /**
+   * Virifies whether all measurements has positive value
```
mayurkale22: s/Virifies/Verifies?

packages/opencensus-core/src/stats/stats.ts
```diff
@@ -114,18 +114,26 @@ export class Stats {
...
+   * @param measurements A list of measurements
+   */
```
mayurkale22: ```@return``` is missing.

packages/opencensus-core/src/stats/stats.ts
isaikevych: Done

packages/opencensus-core/src/stats/stats.ts
isaikevych: Done

isaikevych:
@justindsmith Done

PR #210, opened 2018-11-29
Add stats examples

mayurkale22:


examples/stats/exporter/stackdriver.js
```diff
@@ -0,0 +1,157 @@
...
+    stats.record({
+      measure: mLineLengths,
+      tags,
+      value: processedLine.length
+    });
+
+    stats.record({
+      measure: mLatencyMs,
+      tags,
+      value: endTime.getTime() - startTime.getTime()
```
isaikevych: `startTime` === `endTime`. Could you please correct me if Im wrong?

mayurkale22: Currently yes, because we are not doing any fancy work in ```processLine()``` method.

mayurkale22:
> Please fix the build.

Build failure is fixed in #215. Rebased.

examples/stats/exporter/stackdriver.js
mayurkale22: Used ```process.hrtime``` to record the latency value. PTAL

PR #216, opened 2018-12-03
Update Core library README

mayurkale22:
1. Add badges (node, npm, license, dependencies and downloads).
2. Add usage example.


packages/opencensus-core/README.md
```diff
@@ -1,11 +1,74 @@
...
+Now we will record the desired metrics. To do so, we will use ```stats.record()``` and pass in measurements.
+
+```javascript
+const endTime = new Date();
+const tags = {method: "repl", status: "OK"};
+
+stats.record({
+  measure: mLatencyMs,
+  tags,
+  value: (new Date()) - startTime.getTime()
```
isaikevych: Its not clear how is `startTime` defined

mayurkale22: Done.

packages/opencensus-core/README.md
```diff
@@ -1,11 +1,74 @@
...
+Now we will record the desired metrics. To do so, we will use ```stats.record()``` and pass in measurements.
+
+```javascript
+let startTime = new Date();
+const tags = {method: "repl", status: "OK"};
+
+stats.record({
+  measure: mLatencyMs,
+  tags,
+  value: new Date().getTime() - startTime.getTime()
```
isaikevych: In this case it will be equal and value always will be 0

mayurkale22: Used ```process.hrtime``` to record the latency value. PTAL

PR #231, opened 2018-12-11
Converting `timestamp` and `duration` to the `number` type to follow Zipkin span types

dsborets:
In order to follow a `Zipkin` [Span](https://github.com/openzipkin/zipkin/blob/9b91acbaa3b90ad1c9e4ab62489db05dfd9edf74/zipkin/src/main/java/zipkin2/Span.java#L289) specification, I'm proposing to change `string` type to `number` for `timestamp` and `duration`.

### DIFF
--- a/packages/opencensus-exporter-zipkin/src/zipkin.ts
+++ b/packages/opencensus-exporter-zipkin/src/zipkin.ts
@@ -31,8 +31,8 @@ interface TranslatedSpan {
   id: string;
   parentId?: string;
   kind: string;
-  timestamp: string;
-  duration: string;
+  timestamp: number;
+  duration: number;
   debug: boolean;
   shared: boolean;
   localEndpoint: {serviceName: string};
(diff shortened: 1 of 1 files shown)


justindsmith:
Thanks @dsborets!

Just to be sure, did you validate this change against a zipkin endpoint to make sure the data is received and handled correctly in zipkin? Just want to make sure there are no breaking change, otherwise looks good.

dsborets:
Hi @justindsmith . Sure, I tested it and found that Zipkin accepting any spans disregards of `timestamp` and `duration` fields type (string or number). So we should be safe. I found this issue when we started to use Zipkin-to-Datadog proxy. And that proxy following Zipkin specification accepting only number type (simply by unmarshalling json to Go structure). Initially I thought it's a proxy's problem, but finally realized type discrepancy in opencensus exporter. Probably nobody used opencensus with any servers with more strict type casting than standard Zipkin server

dsborets:
Thank you 👍

PR #232, opened 2018-12-11
Codecov: start reporting code coverage.

mayurkale22:
This PR starts reporting code coverage. If all goes well, coverage should be at https://codecov.io/gh/census-instrumentation/opencensus-node

For now, I have considered four packages (```nodejs```, ```core```, ```exporter-prometheus``` and ```exporter-stackdriver```) for codecov. If all goes well, I will expand this for other packages.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -16,6 +16,10 @@ unit_tests: &unit_tests
     - run:
         name: Run unit tests.
         command: npm run test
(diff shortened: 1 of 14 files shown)


mayurkale22:
> Just for my own clarification, this will report on every PR update as well as merge, right?

I am not sure about merge, but yes on every PR update.

PR #253, opened 2018-12-21
Add MetricProducerManager to keep track of all MetricProducers

mayurkale22:
Basically, the ```MetricProducerManager``` keeps a set of MetricProducers (```MetricProducerForRegistry``` and ```MetricProducerForStats```) that is used by exporters to determine the metrics that need to be exported. Next step is to rewrite all the stats exporters (Stackdriver and prometheus) to use data from MetricProducerManager instead of viewdata.

The ```MetricProducerManager```, ```MetricsComponent``` and ```ExportComponent``` classes are same as Java's implementation of metrics package. https://github.com/census-instrumentation/opencensus-java/tree/master/api/src/main/java/io/opencensus/metrics/export


packages/opencensus-core/src/metrics/export/export-component.ts
```diff
@@ -0,0 +1,32 @@
...
+  private static readonly METRIC_PRODUCER_MANAGER = new MetricProducerManager();
+
+  /**
+   * Returns the global MetricProducerManager which can be used to register
+   * handlers to export all the recorded metrics.
+   *
+   * @return {MetricProducerManager}.
+   */
+  getMetricProducerManager(): MetricProducerManager {
+    return ExportComponent.METRIC_PRODUCER_MANAGER;
```
OsvaldoRosado: Is this just to keep the MetricProducerManager as a singleton? I'm not sure this needs to be a class at all if that's the case (vs simply exporting a const).

mayurkale22: SGTM, Done. PTAL

packages/opencensus-core/src/metrics/metrics.ts
```diff
@@ -14,21 +14,29 @@
...
 export class Metrics {
-  private static readonly METRIC_REGISTRY = Metrics.newMetricRegistry();
+  private static readonly METRIC_COMPONENT = new MetricsComponent();
+
+  /**
+   * Returns the global MetricProducerManager.
+   *
+   * @return {MetricProducerManager}.
+   */
+  static getMetricProducerManager(): MetricProducerManager {
```
OsvaldoRosado: Depending on what `getMetricRegistry` does, you might be able to make these properties instead of getter methods (or even potentially replacing this class for a plain module with consts - just like the previous PR feedback).

packages/opencensus-core/src/metrics/export/metric-producer-manager.ts
```diff
@@ -0,0 +1,69 @@
...
+ */
+
+import {validateNotNull} from '../../common/validations';
+import {MetricProducer} from '../export/metric-producer';
+
+/**
+ * Keeps a set of MetricProducer that is used by exporters to determine the
+ * metrics that need to be exported.
+ */
+export class MetricProducerManager {
```
OsvaldoRosado: Should users be able to construct their own, or should only the singleton be used? If users don't need their own, we might not want to export this.

mayurkale22: > Should users be able to construct their own, or should only the singleton be used?

should only the singleton be used.

> If users don't need their own, we might not want to export this.

Agreed, removed export and added ```MetricProducerManager``` interface type, which is implemented by ```BaseMetricProducerManager```.

packages/opencensus-core/src/metrics/metric-component.ts
```diff
@@ -0,0 +1,40 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {metricProducerManagerInstance} from './export/metric-producer-manager';
+import {MetricRegistry} from './metric-registry';
+
+// Class that holds the implementation instance for MetricRegistry.
```
draffensperger: Nit: should this be a JSDoc comment with `/** */` (similar for other class level comments below)?

mayurkale22: Done, PTAL

draffensperger: LGTM, thanks!

packages/opencensus-core/src/metrics/metrics.ts
```diff
@@ -14,21 +14,31 @@
  * limitations under the License.
  */
 
+import {metricProducerManagerInstance} from './export/metric-producer-manager';
+import {MetricProducerManager} from './export/types';
+import {MetricsComponent} from './metric-component';
 import {MetricRegistry} from './metric-registry';
 
+
+// Class for accessing the default MetricsComponent.
```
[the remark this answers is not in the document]
mayurkale22: Done.

packages/opencensus-core/src/metrics/metrics.ts
```diff
@@ -14,21 +14,31 @@
...
 export class Metrics {
-  private static readonly METRIC_REGISTRY = Metrics.newMetricRegistry();
+  private static readonly METRIC_COMPONENT = new MetricsComponent();
+
+  /**
```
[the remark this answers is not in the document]
mayurkale22: Done.

packages/opencensus-core/src/metrics/metrics.ts
```diff
@@ -14,21 +14,31 @@
...
+   * Returns the global MetricProducerManager.
+   *
+   * @return {MetricProducerManager}.
+   */
+  static getMetricProducerManager(): MetricProducerManager {
+    return metricProducerManagerInstance;
+  }
 
   /**
```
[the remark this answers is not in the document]
mayurkale22: Done.

PR #261, opened 2019-01-03
Replace process.hrtime with Date.now

mayurkale22:
Although ```process.hrtime()``` provides a high-resolution timer, it is from an **unknown relative time, not epoch** i.e. The time is relative to an arbitrary time in the past (not related to the time of day) and therefore not subject to clock drifts. The whole purpose of ```hrtime``` is to gather information on intervals or durations. That's why it doesn't really matter what the point in time was. Also, it is impossible to construct the ISO timestamp from ```hrtime```, which is required when we export metrics data using stackdriver exporter. 

I raised this concern in https://github.com/census-instrumentation/opencensus-node/pull/196, but I failed to justify reasoning that time. While converting OC metrics data to stackdriver exporter I faced the issue with converting ```process.hrtime()``` to _a timestamp in RFC3339 UTC "Zulu" format_ something like this https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts#L183.

I am proposing to use ```Date.now``` function, as it returns the number of milliseconds elapsed since 1 January 1970 00:00:00 UTC. I have written ```timestampFromMillis ``` function to construct Timestamp object from the given milliseconds. Although this is not accurate to nanoseconds, But I think this is best we can get in nodejs. Correct me if I am wrong and let me know your views.


mayurkale22:
@draffensperger and @justindsmith Thanks for the great suggestion. I managed to write function ```getNowWithProcessHRTime``` to calculate ```[seconds, nanoseconds]``` using now and hrtime. Let me know your views on same.

```
const MILLIS_PER_SECOND = 1e3;
const NANOS_PER_MILLI = 1e3 * 1e3;
const NANOS_PER_SECOND = 1e3 * 1e3 * 1e3;

const refTime = Date.now();
const startSecs = Math.floor(refTime / MILLIS_PER_SECOND);
const startNanos = (refTime % MILLIS_PER_SECOND) * NANOS_PER_MILLI;

const hrtime = process.hrtime;
const origin = hrtime();

function getNowWithProcessHRTime() {
  const [offsetSecs, offsetNanos] = hrtime(origin); // [seconds, nanoseconds]

  // determine drfit in seconds and nanoseconds
  const seconds = startSecs + offsetSecs;
  const nanoseconds = startNanos + offsetNanos;

  // if nanos excess NANOS_PER_SECOND value.
  if (nanoseconds >= NANOS_PER_SECOND) {
    return [seconds + 1, nanoseconds % NANOS_PER_SECOND];
  }
  return [seconds, nanoseconds];
}
```

packages/opencensus-core/src/common/time-util.ts
```diff
@@ -0,0 +1,40 @@
...
+}
+
+const seconds = (epochMilli: number): number => {
+  return Number(BigInt(epochMilli).divide(MILLIS_PER_SECOND).toString());
+};
+
+const nanos = (epochMilli: number): number => {
+  const mos = epochMilli - seconds(epochMilli) * MILLIS_PER_SECOND;
+  return Number(BigInt(mos).times(NANOS_PER_MILLI).toString());
+};
```
draffensperger: Optional nit: would it be cleaner to make these regular `function` declarations rather than `const` function expressions?

justindsmith:
@mayurkale22 LGTM!

mayurkale22:
I have modified the code to use above mentioned logic, PTAL @draffensperger @justindsmith

packages/opencensus-core/src/common/time-util.ts
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Gets the current timestamp with seconds and nanoseconds.
+ *
+ * @returns {Timestamp} The Timestamp.
+ */
+export function getTimestampWithProcessHRTime(): Timestamp {
+  const [offsetSecs, offsetNanos] = hrtime(origin);  // [seconds, nanoseconds]
+
+  // determine drfit in seconds and nanoseconds
```
draffensperger: Nit: "drift" spelling

packages/opencensus-core/src/common/time-util.ts
```diff
@@ -0,0 +1,47 @@
...
+  const seconds = startSecs + offsetSecs;
+  const nanos = startNanos + offsetNanos;
+
+  // if nanos excess NANOS_PER_SECOND value.
```
draffensperger: Is this comment needed, it seems to express pretty closely what the if statement says?

packages/opencensus-core/test/test-derived-gauge.ts
```diff
@@ -73,15 +68,19 @@ describe('DerivedGauge', () => {
...
-          metric.timeseries, [{
-            labelValues: LABEL_VALUES_200,
-            points: [{value: 2, timestamp: {nanos: 1e7, seconds: 1000}}]
-          }]);
+
+      const [timeseries] = metric.timeseries;
+      const [points] = timeseries.points;
+      assert.deepStrictEqual(timeseries.labelValues, LABEL_VALUES_200);
+      assert.equal(points.value, 2);
+      assert.equal(Math.floor(now / 1e3), points.timestamp.seconds);
```
draffensperger: What would you think about mocking `Date.now` and `process.hrtime` for the test (setting them and unsetting them before/after)? That way if someone sets a breakpoint in the test and some time elapses it will still pass and you can also assert exact values for nanoseconds.

packages/opencensus-core/test/test-metric-registry.ts
```diff
@@ -95,6 +89,7 @@ describe('addInt64Gauge', () => {
     const pointEntry = int64Gauge.getOrCreateTimeSeries(LABEL_VALUES_200);
     pointEntry.add(100);
 
+    now = Date.now();
```
draffensperger: Similar comment here about mocking out the time vs. measuring it in the test

packages/opencensus-core/test/test-metric-registry.ts
```diff
@@ -281,6 +267,7 @@ describe('addDerivedInt64Gauge', () => {
     derivedInt64Gauge.createTimeSeries(LABEL_VALUES_200, map);
     map.set('key1', 'value1');
 
+    now = Date.now();
```
draffensperger: For these tests and others below too

packages/opencensus-core/test/test-time-util.ts
```diff
@@ -0,0 +1,27 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as assert from 'assert';
+import {getTimestampWithProcessHRTime} from '../src/common/time-util';
+
+describe('getTimestampWithProcessHRTime()', () => {
+  it('compare now with getTimestampWithProcessHRTime', () => {
```
draffensperger: What would you think about using time mocking here and adding tests that cover the cases where the nanoseconds both do and don't overflow?

What you could do is move the initialization logic in `common/time-util.ts` above into a function say `setupHrtimeDrift` or similar, then you would add a new exported variable `export TEST_ONLY = {setupHrtimeDrift}` with a comment that is exported to enable testing the drift configuration (and of course you call the `setupHrtimeDrift` in the `time-util.ts` file). 

Then you import the `TEST_ONLY` symbol here in the test and call `setupHrtimeDrift` after you have mocked `process.hrtime` and `Date.now`. Then you redefine them again before calling `getTimestampWithProcessHRTime` for the two cases to test.

If this seems too involved, I'm open to leaving this as-is though.

packages/opencensus-core/test/test-derived-gauge.ts
draffensperger: What if you had something like this:

```
let hrtime = process.hrtime;
let hrtimeOrigin: [number, number] = [0, 0];
let hrtimeRefSeconds = 0;
let hrtimeRefNanos = 0;

/** This is used to enable tests to mock process.hrtime while still allow us to cache it. */
function resetHrtimeFunctionCache() {
   hrtime = process.hrtime;
}

function setHrtimeReference() {
  resetHrtimeFunctionCache();
  hrtimeOrigin = hrtime();
  const refTime = Date.now();
  hrtimeRefSeconds = ...
  hrtimeRefNanos = ...
}

setHrtimeReference();

export TEST_ONLY = {setHrtimeReference};

.....

// In the test

import {TEST_ONLY, ...} from '../common/time-utils.ts';

...

let realHrtimeFn = process.hrtime;
let realNowFn = Date.now;
beforeEach(() => {
   process.hrtime = () => [55, 22];
   Date.now = () => 14500000;
    // Force the clock to recalibrate the time offset with the mocked time
   TEST_ONLY.setHrtimeReference();
   // Simulate that now some time passed ...   
   process.hrtime = () => [66, 33];
   TEST_ONLY.resetHrtimeFunctionCache();
});

afterEach(() => {
   process.hrtime = realHrtimeFn;
   Date.now = realDateNow;
   // Reset the hrtime reference so that it uses a real clock again.
   TEST_ONLY.setHrtimeReference();
});
```

If this feels too involved, I'm OK with skipping it but wanted to put it there as an option. If you go this route, it could be helpful to export these as some kind of time testing helpers to easily set/unset the mocked time.

packages/opencensus-core/src/common/time-util.ts
mayurkale22: Done.

packages/opencensus-core/src/common/time-util.ts
mayurkale22: Make sense, removed.

packages/opencensus-core/test/test-derived-gauge.ts
mayurkale22: @draffensperger Awesome, thanks for the help. I have modified the tests, PTAL.

mayurkale22:
@justindsmith please take another look. Thanks.

PR #264, opened 2019-01-07
Update README, keywords and gauges & HTTP instrumentation quickstart examples.

mayurkale22:
1.  Add ```Gauge``` and ```Derived Gauge``` quickstart examples.
2. Add ```HTTP``` instrumentation quickstart example.
3. Add ```Metrics``` and ```Stats``` keywords in package.json.
4. Add codecov and license badge.

### DIFF
--- a/examples/gauges/derived_gauge_quickstart.js
+++ b/examples/gauges/derived_gauge_quickstart.js
@@ -1,0 +1,48 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is an example shows how to create a Derived Gauge metric.
+ * Derived Gauge metric API is to report instantaneous measurement of an
+ * int64/double value. Gauges can go both up and down. The gauges values
+ * can be negative.
+ */
+
+const { Metrics, MeasureUnit } = require("@opencensus/core");
+
+// a registry is a collection of metric objects.
+const metricRegistry = Metrics.getMetricRegistry();
+
+// application labels - applied to each metric / gauge.
+const labelKeys = [{ key: "Method", description: "desc" }];
+const labelValues = [{ value: "GET" }];
+
+// a new gauge instance - builds a new Int64 derived gauge to be added to the
+// registry.
+const int64Gauge = metricRegistry.addDerivedInt64Gauge(
+  "request/count",
+  "The total number of requests",
+  MeasureUnit.UNIT,
+  labelKeys
+);
+
+// the value of a single point in the TimeSeries is observed from a map
+// collection (SizeAttributeInterface)
+const map = new Map();
+int64Gauge.createTimeSeries(labelValues, map);
+
+// sets value
+map.set('key', 'value');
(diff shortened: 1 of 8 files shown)


examples/gauges/derived_gauge_quickstart.js
```diff
@@ -0,0 +1,48 @@
...
+// a registry is a collection of metric objects.
+const metricRegistry = Metrics.getMetricRegistry();
+
+// application labels - applied to each metric / gauge.
+const labelKeys = [{ key: "Method", description: "desc" }];
+const labelValues = [{ value: "GET" }];
+
+// a new gauge instance - builds a new Int64 derived gauge to be added to the
+// registry.
+const int64Gauge = metricRegistry.addDerivedInt64Gauge(
```
draffensperger: The concept of derived gauges is a bit new to me. I would have normally expected that the request count would be some sort of cumulative metric. Could you explain how the derived gauge relates to the cumulative count? Maybe a short explanation in the code comments would help too with a link to more details?

examples/gauges/derived_gauge_quickstart.js
```diff
@@ -0,0 +1,48 @@
...
+  labelKeys
+);
+
+// the value of a single point in the TimeSeries is observed from a map
+// collection (SizeAttributeInterface)
+const map = new Map();
+int64Gauge.createTimeSeries(labelValues, map);
+
+// sets value
+map.set('key', 'value');
```
draffensperger: Should `'value'` here be something more realistic? How does it relate to the `labelValues` above? I would have guessed that maybe only `'GET'` would be a valid option based on that. Maybe change it to have a couple possible label values and then have it to one of them? Or am I missing something here?

examples/gauges/gauge_quickstart.js
```diff
@@ -0,0 +1,48 @@
...
+  "The total number of requests",
+  MeasureUnit.UNIT,
+  labelKeys
+);
+
+// It is recommended to keep a reference of a point for manual operations.
+const requestCount = int64Gauge.getOrCreateTimeSeries(labelValues);
+
+// adds value -> 1
+requestCount.add(1);
```
draffensperger: Since this example is the same as above up until this point, should it be combined into one file so that someone can see setting a metric label and adding/setting values as well?

packages/opencensus-instrumentation-http/README.md
```diff
@@ -5,6 +5,62 @@ OpenCensus HTTP Instrumentation allows the user to automatically collect trace d
...
+
+  console.log(`Node HTTP listening on ${port}`);
+});
+```
+
+#### Now, set up a HTTP Instrumentation.
+Now that we have our app working, let’s actually add the HTTP Trace instrumentation.
+
+```javascript
+const { plugin } = require('@opencensus/instrumentation-http');
```
OsvaldoRosado: It's not clear to me where this code should go from the text of this readme alone. Before the `index.js` content listed above? After it? Another file?

packages/opencensus-instrumentation-http/README.md
```diff
@@ -5,6 +5,62 @@ OpenCensus HTTP Instrumentation allows the user to automatically collect trace d
...
+
+// 2. Setup Tracer
+const tracer = new CoreTracer();
+tracer.start({samplingRate: 1}); // always sampler
+
+// 3. Enable HTTP plugin
+plugin.enable(http, tracer, version, null);
+```
+
+The instrumented handler creates a trace automatically for each request that is received. We shall need a ```Trace exporter```, allowing us to examine the traces.
```
justindsmith: It also creates a span for any outgoing request that is made. So both incoming and outgoing are captured.

packages/opencensus-instrumentation-http/README.md
justindsmith: I'm also a little concerned that this is not really using the APIs in the "recommended" way (through the `Tracing` singleton). While I know it's to highlight how to enable just this single plugin, I'm not sure what value this example is providing for anyone that might land on this README file.

I think this example would be better moved to a separate location if we need it, and the README should focus on common usage patterns only - how a developer would most generally need to interact with the library. In this case "not much" since it's enabled by default when using the Tracing singleton.

mayurkale22:
Obsolete, I will open new PR to add gauges example.

PR #265, opened 2019-01-08
export Metrics classes and types

mayurkale22:


packages/opencensus-core/src/index.ts
```diff
@@ -24,6 +24,29 @@ export * from './trace/instrumentation/types';
 export * from './trace/propagation/types';
 export * from './exporters/types';
 export * from './common/types';
+export * from './metrics/gauges/types';
+import {Metric, MetricDescriptor, TimeSeries, MetricDescriptorType, LabelKey, LabelValue, Point as TimeSeriesPoint, DistributionValue, BucketOptions, Bucket as DistributionBucket, SummaryValue, Explicit, Exemplar, Timestamp, Snapshot, ValueAtPercentile, MetricProducerManager, MetricProducer} from './metrics/export/types';
+
+export {
```
draffensperger: Would it be equivalent to use `export {Metric, ...} from './metrics/export/types'` rather than importing them and then exporting them? If so, that feels a little more compact to me.

mayurkale22: The only reason I did this way is to resolve ambiguity. The ```Bucket``` and ```Point``` are already exported under stats module.  Both the members are there for metrics module also,  I have added alias for ```Point ->  TimeSeriesPoint``` and ```Bucket -> DistributionBucket``` before exporting it.

draffensperger: OK, makes sense. Would it work to use `export .. from ...` for the things that aren't renamed and then a separate import and export for the things that get renamed?

Really a nit though so feel free to leave as is.

Issue #257, opened 2018-12-28
Rewrite stats exporters to use Metrics instead of Stats (ViewData)


mayurkale22:
The plan is to use [Metric data model ](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto)to transform data from Stats API (Stats data models to Metric data models handled by [MetricUtils](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/stats/metric-utils.ts) class) and Gauge APIs using Metric Producer Manager. 
- [x] Rewrite/update Stackdriver stats exporter.
- [ ] Rewrite/update Prometheus stats exporter.

**Way to translate Stats data model to the Metrics.**
- MetricPoint
    - MetricDescriptor
        - Name <- ViewData.View.name
        - Description <- ViewData.View.description
        - Unit <- ViewData.View.Measure.unit
        - Type <- Cumulative (except LastValue which is GAUGE)
    - TimeSeries
        - Labels <- {ViewData.rows, ViewData.View.columns}
- CumulativePoint
    - Value <- ViewData.AggregationData.value (for DistributionValue the bucket_bounds are defined in the ViewData.View.DistributionAggregation.bucket_bounds)
    - Start_time <- ViewData.start_time
    - End_time <- ViewData.end_time

Follow-up of #253

Equivalent change in Java: census-instrumentation/opencensus-java#1501.

vigneshtdev:
hey i would like to help out with this issue. In opencencus java there is a tags package where the different tagvalue & tagkey can be configured. I am unable to find a similar module in the package. Can you point out where it us?

mayurkale22:
Rewrite/update Prometheus stats exporter. - **On Hold**

Currently, prom-client for node is not compatible with OpenCensus Metrics data model. Basically, prom-client doesn't seem to support custom collectors. I have opened an issue: https://github.com/siimon/prom-client/issues/241 to track this.

PR #266, opened 2019-01-09
Add helper util to convert OC Metrics data models to SD monitoring data models

mayurkale22:
This PR is part of #257, Basically, ```StackdriverStatsExporterUtils``` methods will be used by ```StackdriverStatsExporter``` to translate data from OpenCensus data models to StackDriver monitoring data models. I will send separate PR to include this util in ```StackdriverStatsExporter```.

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
+++ b/packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
@@ -17,272 +17,226 @@
 import {BucketOptions, DistributionBucket, DistributionValue, LabelKey, LabelValue, Metric, MetricDescriptor as OCMetricDescriptor, MetricDescriptorType, TimeSeriesPoint, Timestamp} from '@opencensus/core';
 import * as os from 'os';
 import * as path from 'path';
-import {Distribution, LabelDescriptor, MetricDescriptor, MetricKind, Point, TimeSeries, ValueType} from './types';
+
+import {Distribution, LabelDescriptor, MetricDescriptor, MetricKind, MonitoredResource, Point, TimeSeries, ValueType} from './types';
+
+const OPENCENSUS_TASK = 'opencensus_task';
+const OPENCENSUS_TASK_DESCRIPTION = 'Opencensus task identifier';
+export const OPENCENSUS_TASK_VALUE_DEFAULT = generateDefaultTaskValue();
+
+/** Converts a OpenCensus MetricDescriptor to a StackDriver MetricDescriptor. */
+export function createMetricDescriptorData(
+    metricDescriptor: OCMetricDescriptor, metricPrefix: string,
+    displayNamePrefix: string): MetricDescriptor {
+  return {
+    type: getMetricType(metricDescriptor.name, metricPrefix),
+    description: metricDescriptor.description,
+    displayName: createDisplayName(metricDescriptor.name, displayNamePrefix),
+    metricKind: createMetricKind(metricDescriptor.type),
+    valueType: createValueType(metricDescriptor.type),
+    unit: metricDescriptor.unit,
+    labels: createLabelDescriptor(metricDescriptor.labelKeys)
+  };
+}
 
 /**
- * Util methods to convert OpenCensus Metrics data models to StackDriver
- * monitoring data models.
+ * Converts metric's timeseries to a list of TimeSeries, so that metric can be
+ * uploaded to StackDriver.
  */
-export class StackdriverStatsExporterUtils {
-  static readonly OPENCENSUS_TASK: string = 'opencensus_task';
-  static readonly OPENCENSUS_TASK_DESCRIPTION: string =
-      'Opencensus task identifier';
-  static readonly OPENCENSUS_TASK_VALUE_DEFAULT =
-      StackdriverStatsExporterUtils.generateDefaultTaskValue();
+export function createTimeSeriesList(
+    metric: Metric, monitoredResource: MonitoredResource,
+    metricPrefix: string): TimeSeries[] {
+  const timeSeriesList: TimeSeries[] = [];
 
-  /**
-   * Returns a task label value in the format of 'nodejs-<pid>@<hostname>'.
-   */
-  static generateDefaultTaskValue(): string {
-    const pid = process.pid;
-    const hostname = os.hostname() || 'localhost';
-    return 'nodejs-' + pid + '@' + hostname;
-  }
+  // TODO(mayurkale): Use Resource API here, once available (PR#173)
+  const metricDescriptor = metric.descriptor;
+  const metricKind = createMetricKind(metricDescriptor.type);
+  const valueType = createValueType(metricDescriptor.type);
 
-  /**
-   * Converts a OpenCensus MetricDescriptor to a StackDriver MetricDescriptor.
-   * @param metricDescriptor The OpenCensus MetricDescriptor.
-   */
-  static createMetricDescriptorData(
-      metricDescriptor: OCMetricDescriptor, metricPrefix: string,
-      displayNamePrefix: string): MetricDescriptor {
-    return {
-      type: this.getMetricType(metricDescriptor.name, metricPrefix),
-      description: metricDescriptor.description,
-      displayName:
-          this.createDisplayName(metricDescriptor.name, displayNamePrefix),
-      metricKind: this.createMetricKind(metricDescriptor.type),
-      valueType: this.createValueType(metricDescriptor.type),
-      unit: metricDescriptor.unit,
-      labels: this.createLabelDescriptor(metricDescriptor.labelKeys)
-    } as MetricDescriptor;
-  }
-
-  /**
-   * Converts metric's timeseries to a list of TimeSeries, so that metric can be
-   * uploaded to StackDriver.
-   * @param metric The metric to get TimeSeries information from.
-   */
-  static createTimeSeriesList(
-      metric: Metric,
-      monitoredResource: {type: string, labels: {[key: string]: string}},
-      metricPrefix: string): TimeSeries[] {
-    const timeSeriesList: TimeSeries[] = [];
-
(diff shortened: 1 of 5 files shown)


packages/opencensus-exporter-stackdriver/src/types.ts
```diff
@@ -103,7 +103,7 @@ export interface Distribution {
 }
 
 export interface Point {
-  interval: {endTime: string, startTime: string};
+  interval: {endTime: string, startTime?: string};
```
mayurkale22: ```startTime``` should be optional (Must be present for cumulative metrics only).  More info: https://cloud-dot-devsite.googleplex.com/monitoring/api/ref_v3/rest/v3/TimeInterval

packages/opencensus-exporter-stackdriver/src/types.ts
```diff
@@ -115,8 +115,8 @@ export interface Point {
 
 export interface TimeSeries {
   metric: {type: string; labels: {[key: string]: string};};
-  resource: {type: 'global', labels: {[key: string]: string}};
+  resource: {type: string, labels: {[key: string]: string}};
```
mayurkale22: In the future, we have planned to add more type like ```k8s_container```,  ```gce_instance``` and ```aws_ec2_instance```.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+    // TODO(mayurkale): Use Resource API here, once available (PR#173)
+    const metricDescriptor = metric.descriptor;
+    const metricKind = this.createMetricKind(metricDescriptor.type);
+    const valueType = this.createValueType(metricDescriptor.type);
+
+    for (const timeSeries of metric.timeseries) {
+      timeSeriesList.push({
+        metric: this.createMetric(
+            metricDescriptor, timeSeries.labelValues, metricPrefix),
+        resource: monitoredResource,
```
songy23: Just to double check: it's fine to use the same `monitoredResource` here (i.e no need to convert)?

mayurkale22: Currently no, but once we have monitored resource util in place (#173) we need to do the conversion same as Java's implementation.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+import * as os from 'os';
+import * as path from 'path';
+import {Distribution, LabelDescriptor, MetricDescriptor, MetricKind, Point, TimeSeries, ValueType} from './types';
+
+/**
+ * Util methods to convert OpenCensus Metrics data models to StackDriver
+ * monitoring data models.
+ */
+export class StackdriverStatsExporterUtils {
+  static readonly OPENCENSUS_TASK: string = 'opencensus_task';
```
draffensperger: Why are these constants and methods `static` within a class? While that is needed in Java, IMO it's just extra wrapping for TypeScript since the module itself already acts as a namespace. What would you think about just exporting the functions and constants directly?

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+    return {
+      type: this.getMetricType(metricDescriptor.name, metricPrefix),
+      description: metricDescriptor.description,
+      displayName:
+          this.createDisplayName(metricDescriptor.name, displayNamePrefix),
+      metricKind: this.createMetricKind(metricDescriptor.type),
+      valueType: this.createValueType(metricDescriptor.type),
+      unit: metricDescriptor.unit,
+      labels: this.createLabelDescriptor(metricDescriptor.labelKeys)
+    } as MetricDescriptor;
```
draffensperger: Could you remove the `as MetricDescriptor`? The TS compiler should know the object must match the type based on the method return type.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+      monitoredResource: {type: string, labels: {[key: string]: string}},
+      metricPrefix: string): TimeSeries[] {
+    const timeSeriesList: TimeSeries[] = [];
+
+    // TODO(mayurkale): Use Resource API here, once available (PR#173)
+    const metricDescriptor = metric.descriptor;
+    const metricKind = this.createMetricKind(metricDescriptor.type);
+    const valueType = this.createValueType(metricDescriptor.type);
+
+    for (const timeSeries of metric.timeseries) {
```
draffensperger: Optional: could we use `map` here instead of a for loop + push?

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+    }
+  }
+
+  /**
+   * Constructs a LabelDescriptor from a LabelKey.
+   * @param tag The Tags to get TimeSeries information from.
+   */
+  static createLabelDescriptor(labelKeys: LabelKey[]): LabelDescriptor[] {
+    const labelDescriptorList = labelKeys.map(labelKey => {
+      return {
```
draffensperger: Optional, you can avoid the `return` if you put the object in paratheses, so `labelKeys.map(labelKey => ({ key: ...}))`

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+        key: labelKey.key,
+        valueType: 'STRING',  // Now we only support String type.
+        description: labelKey.description
+      } as LabelDescriptor;
```
draffensperger: Could you use a type annotation on `labelDescriptorList`, i.e. `const labelDesceriptorList: LabelDescriptor[] = ` rather than a type assertion `as LabelDescriptor` here? The type annotation will cause the code to fail to compile if the interface changes so would help detect refactoring bugs.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+  }
+
+  /**
+   * Creates a Metric using the LabelKeys and LabelValues.
+   * @param metricDescriptor The OpenCensus MetricDescriptor.
+   * @param labelValues The OpenCensus LabelValue.
+   */
+  static createMetric(
+      metricDescriptor: OCMetricDescriptor, labelValues: LabelValue[],
+      metricPrefix: string): {type: string; labels: {[key: string]: string};} {
```
draffensperger: The `{type: string; labels: {[key: string]: string};}` type looks fairly complex, what would you think about making it a first class exported type alias?

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+  /**
+   * Converts timeseries's point, so that metric can be uploaded to StackDriver.
+   * @param point The timestamped measurement.
+   * @param startTimeStamp The start timestamp of timeseries.
+   * @param valueType The ValueType of metric.
+   */
+  static createPoint(
+      point: TimeSeriesPoint, startTimeStamp: Timestamp,
+      valueType: ValueType): Point {
+    let value;
```
draffensperger: Optional: you can avoid use of `let` here by creating a separate function like `createValue` or similar that would take in the value type and value and return the value, then just have `const value = createValue(valueType, point)`

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+      value = {int64Value: point.value as number};
+    } else if (valueType === ValueType.DOUBLE) {
+      value = {doubleValue: point.value as number};
+    } else if (valueType === ValueType.DISTRIBUTION) {
+      value = {
+        distributionValue:
+            this.createDistribution(point.value as DistributionValue)
+      };
+    } else {
+      // console.log(`${valueType} is not supported.`);
```
draffensperger: Could you remove this commented code and `else` black?

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,288 @@
...
+    return bucketCounts;
+  }
+
+  private static toISOString(timestamp: Timestamp) {
+    const str = new Date(timestamp.seconds * 1000).toISOString();
+    const nsStr = `${this.padNS(timestamp.nanos)}`.replace(/0+$/, '');
+    return str.replace('000Z', `${nsStr}Z`);
+  }
+
+  private static padNS(ns: number) {
```
draffensperger: Could you make this function name a bit more descriptive? It seems like it's left padding a number, but I wouldn't know that without reading the code.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -0,0 +1,242 @@
...
+  bucketCounts.push(0);
+  buckets.map((bucket: DistributionBucket) => {
+    bucketCounts.push(bucket.count);
+  });
+  return bucketCounts;
+}
+
+/** Returns a task label value in the format of 'nodejs-<pid>@<hostname>'. */
+function generateDefaultTaskValue(): string {
+  console.log('inside generateDefaultTaskValue');
```
justindsmith: Remove console.log

mayurkale22: Done.

PR #271, opened 2019-01-10
Simplify Logger interface, remove silly, make level optional

draffensperger:
By removing `silly` and making the `level` attribute optional (and defaulted to `debug`) then the `console` interface itself will implement `Logger`. See [this TS playground](http://www.typescriptlang.org/play/#src=export%20type%20LogFunction%20%3D%20(message%3A%20any%2C%20...args%3A%20any%5B%5D)%20%3D%3E%20void%3B%0D%0A%0D%0A%2F**%20Defines%20an%20logger%20interface.%20*%2F%0D%0Aexport%20interface%20Logger%20%7B%0D%0A%20%20level%3F%3A%20string%3B%0D%0A%20%20error%3A%20LogFunction%3B%0D%0A%20%20warn%3A%20LogFunction%3B%0D%0A%20%20info%3A%20LogFunction%3B%0D%0A%20%20debug%3A%20LogFunction%3B%20%20%0D%0A%7D%0D%0A%0D%0Aconst%20logger%3A%20Logger%20%3D%20console%3B)

This will be useful for `opencensus-web` where a major goal is small code size for web clients, so being able to just use `console` itself as a logger cuts down on code size.

This does seem like a relatively minor breaking change though.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Add Gauges (`DoubleGauge`, `LongGauge`, `DerivedDoubleGauge`, `DerivedLongGauge`) APIs.
 - Add support for supplying instrumentation configuration via tracing option. Option argument added to instrumentation interface.
 - Add ignoreIncomingPaths and ignoreOutgoingUrls support to the http and https tracing instrumentations.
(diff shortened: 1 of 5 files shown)


draffensperger:
@mayurkale22 I added a note to the changlog, could you take another look and merge if all is good?

Issue #272, opened 2019-01-10
Make `Span` start/end times be high-resolution timestamps


draffensperger:
Currently the `start` and `end` times in the [Span interface](https://github.com/census-instrumentation/opencensus-node/blob/3db8bf551d1c2519809f045592364f15707851c2/packages/opencensus-core/src/trace/model/types.ts#L153) use the `Date` type, which is accurate only to milliseconds.

But it would be nice to have higher timestamp accuracy as sub-millisecond times can be useful to measure fast events or small differences between client and service spans for an RPC. My suggestion would be to migrate to use the [Timestamp](https://github.com/census-instrumentation/opencensus-node/blob/47a672e5f082d0ae981b53cd32000273ac0f593f/packages/opencensus-core/src/metrics/export/types.ts#L314) interface or a `[seconds, nanos]` tuple.

We now have the [getTimestampWithProcessHRTime](https://github.com/census-instrumentation/opencensus-node/blob/3db8bf551d1c2519809f045592364f15707851c2/packages/opencensus-core/src/common/time-util.ts#L49), which would be used to populate the times.

This would be a breaking change, but I think doing it sooner rather than later would reduce the pain of it.

draffensperger:
We may be able to make this less of a breaking change by introducing new field names for `Span` for the high resolution start/end times but keep `start` and `end` as property accessors that return `Date`. Then eventually those old fields could be deprecated and removed. Now sure how much we are concerned about backwards compatibility at this point but that is an option.

PR #273, opened 2019-01-10
Set startTimestamp to creation time of cumulative view

mayurkale22:


mayurkale22:
Same as #272, we need to change 'View' startTime to use high resolutions timestamp (https://github.com/census-instrumentation/opencensus-node/blob/3db8bf551d1c2519809f045592364f15707851c2/packages/opencensus-core/src/common/time-util.ts#L49). I think this is breaking change and I prefer to do in separate PR.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -200,9 +200,10 @@ export class BaseView implements View {
 
   /**
    * Gets view`s metric
+   * @param start The start timestamp
```
draffensperger: Should this say add "in epoch milliseconds"?

mayurkale22: Done,thanks

PR #275, opened 2019-01-11
Stats/Stackdriver: export data using metrics API 

mayurkale22:
This PR is part of #257 

The majority of work has been handled/covered in  ```StackdriverStatsExporterUtils ``` https://github.com/census-instrumentation/opencensus-node/pull/266, this PR is based on that work to export data to backend.

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
+++ b/packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
@@ -14,30 +14,34 @@
  * limitations under the License.
  */
 
-import {AggregationType, DistributionData, logger, Logger, Measure, Measurement, MeasureType, MeasureUnit, StatsEventListener, Tags, View} from '@opencensus/core';
+import {logger, Logger, Measurement, Metric, MetricDescriptor as OCMetricDescriptor, MetricProducerManager, Metrics, StatsEventListener, View} from '@opencensus/core';
 import {auth, JWT} from 'google-auth-library';
 import {google} from 'googleapis';
-import * as path from 'path';
 
-import {Distribution, LabelDescriptor, MetricDescriptor, MetricKind, Point, StackdriverExporterOptions, TimeSeries, ValueType} from './types';
+import {createMetricDescriptorData, createTimeSeriesList} from './stackdriver-stats-utils';
+import {StackdriverExporterOptions, TimeSeries} from './types';
 
 google.options({headers: {'x-opencensus-outgoing-request': 0x1}});
 const monitoring = google.monitoring('v3');
-
-const RECORD_SEPARATOR = String.fromCharCode(30);
-const UNIT_SEPARATOR = String.fromCharCode(31);
+const GOOGLEAPIS_SCOPE = 'https://www.googleapis.com/auth/cloud-platform';
 
 /** Format and sends Stats to Stackdriver */
 export class StackdriverStatsExporter implements StatsEventListener {
   private period: number;
   private projectId: string;
   private metricPrefix: string;
-  private viewToUpload:
-      {[key: string]: {view: View; tags: {[key: string]: Tags;}}};
+  private displayNamePrefix: string;
+  // tslint:disable:no-any
+  private onMetricUploadError: any;
   private timer: NodeJS.Timer;
+  static readonly DEFAULT_DISPLAY_NAME_PREFIX: string = 'OpenCensus';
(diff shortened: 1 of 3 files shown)


packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
```diff
@@ -14,30 +14,34 @@
...
 /** Format and sends Stats to Stackdriver */
 export class StackdriverStatsExporter implements StatsEventListener {
   private period: number;
   private projectId: string;
   private metricPrefix: string;
-  private viewToUpload:
-      {[key: string]: {view: View; tags: {[key: string]: Tags;}}};
+  private displayNamePrefix: string;
+  // tslint:disable:no-any
+  private onMetricUploadError: any;
```
draffensperger: It seems like a possible value of this is a function. Could you make the type something like `(err: Error) => void|{}`, which would allow it to either a function that takes an error, or just any arbitrary object? That way there is some amount of documentation about what role it plays.

packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
```diff
@@ -47,82 +51,99 @@ export class StackdriverStatsExporter implements StatsEventListener {
...
-      const v = this.viewToUpload[name];
-      for (const tagsKey of Object.keys(v.tags)) {
-        timeSeries.push(this.createTimeSeriesData(v.view, v.tags[tagsKey]));
-      }
-    }
-    this.viewToUpload = {};
+    const resourceLabels = {project_id: this.projectId};
+    const monitoredResource = {type: 'global', labels: resourceLabels};
+
+    metricsList.map(metric => {
```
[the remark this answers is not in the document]
mayurkale22: Done

packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
```diff
@@ -47,82 +51,99 @@ export class StackdriverStatsExporter implements StatsEventListener {
...
+      const timeSeriesList =
+          createTimeSeriesList(metric, monitoredResource, this.metricPrefix);
+      timeSeriesList.forEach(ts => {
```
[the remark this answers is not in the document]
mayurkale22: Done

packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
mayurkale22: Agreed, done.

packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
```diff
@@ -47,82 +50,96 @@ export class StackdriverStatsExporter implements StatsEventListener {
...
-        auth: authClient
-      };
+  async export() {
+    const metricsList: Metric[] = [];
+    const metricProducerManager: MetricProducerManager =
+        Metrics.getMetricProducerManager();
+    for (const metricProducer of metricProducerManager.getAllMetricProducer()) {
+      for (const metric of metricProducer.getMetrics()) {
+        const isRegistered =
+            await this.registerMetricDescriptor(metric.descriptor);
```
draffensperger: Would it be worth running these `registerMetricDescriptor` calls in parallel? Not sure how useful it would be but they don't seem dependent on each other in this loop.

It seems like if you ran them all in parallel with something like `Promise.all` then you could aggregate their results together into the `metricsList` list once they all complete.

mayurkale22: The idea is to register each metricDescriptor once (and ignore metricDescriptor that are already registered). If we run  ```registerMetricDescriptor``` in parallel, there are chances will make more external calls than usual. WDYT?

draffensperger: OK. I suppose registering metric descriptors may not happen that often, so it's probably OK to leave as is. Maybe just something we can think of if anybody has a performance issue with it that we could try to compress it to a `Set` and run in parallel, but I like leaving it alone for now.

mayurkale22: This is certainly a good point, I have added TODO comment for now to track this in future.

mayurkale22:
I will merge this PR later today, if no other comments.

PR #277, opened 2019-01-11
Nit: Export for tests and Singleton instance for BaseMetricProducerMa…

mayurkale22:
…nager


packages/opencensus-core/src/metrics/export/metric-producer-manager.ts
```diff
@@ -66,4 +74,4 @@ class BaseMetricProducerManager implements MetricProducerManager {
   }
 }
 
-export const metricProducerManagerInstance = new BaseMetricProducerManager();
```
mayurkale22: This will fix compiler error: ```error TS4025: Exported variable 'metricProducerManagerInstance' has or is using private name 'BaseMetricProducerManager'.```

Issue #276, opened 2019-01-11
Decide how to inject MetricProducerForStats.


mayurkale22:
https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/stats/stats.ts#L40-L46

```MetricProducerForStats``` is MetricProducer for stats that can be registered for exporting using ```MetricProducerManager```. If we don't add or register this stats producer into ```MetricProducerManager```, stats data will not be exported to backend/exporter. 
```
// It should be something like below
const metricProducer: MetricProducer = new MetricProducerForStats(statsObject);
Metrics.getMetricProducerManager().add(metricProducer);
```

At this moment, two approaches are there in my mind.
1. Inside ```Stats``` ctor, add above code (statsObject -> this). Everything is working fine, but with this every stats object creation will add new entry to ```MetricProducerManager```.
2. Make singleton ```Stats``` class (similar to ```Tracing```) and inject singleton instance in below code. This would be a breaking change, but this is what we have done in Java.

### Background

**Stats vs Metrics**

- The ```Stats``` Library allows users to record raw measurements or data without any aggregations. It also allows users to define views (that will produce metrics, when they are applied across a set of measurements) to configure how data are aggregated and broken down (using tag keys). In short, Stats is an API for recording raw measurements. 

- The ```Metrics``` ([metrics.proto](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto)) is an API that works with aggregated data. We have provided Gauge API to report already aggregated data.

**MetricProducerManager**
At a high level, we have added ```MetricProducerManager``` to Keep a set of MetricProducer that is used by exporters to determine the metrics that need to be exported. In Node, we have two source of MetricProducer -> ```Stats``` and ```MetricRegistry (Gauges)```.

![image](https://user-images.githubusercontent.com/6525361/51008271-ffa91d80-14ff-11e9-8d3c-43d181c10a41.png)

_Note: Codahale and Micrometer is applicable for OpenCensus Java library only._

mayurkale22:
@draffensperger @justindsmith @OsvaldoRosado Let me know your thoughts on this.

draffensperger:
[...]

PR #283, opened 2019-01-14
WIP: Export global singleton stats and register MetricProducerForStats

mayurkale22:
This PR is based on #276 - The ```new Stats()``` has been deprecated on Stats class. The global singleton ```stats``` object should be used instead.


mayurkale22:
Duplicate, handled by #291

PR #289, opened 2019-01-16
Minor changes to Stackdriver example

DazWilkin:
Two changes (one needs to be fixed):

-- tagKeys used when creating the view ought match those used when recording stats. The code now creates tagKeys of `["method", "status"];`

-- views should be created and then registered but this appears to not be required with this implementation. The code now creates the views without assigning these to unused constants. If the library is amended to require the `createView(X)` followed by `registeredView(X)`, then these constants may be reverted. Or the result `createView` may be applied directly to `stats.registerView(stats.createView(...)`.

### DIFF
--- a/examples/stats/exporter/stackdriver.js
+++ b/examples/stats/exporter/stackdriver.js
@@ -62,35 +62,35 @@ const stream = fs.createReadStream("./test.txt");
 // Create an interface to read and process our file line by line
 const lineReader = readline.createInterface({ input: stream });
 
-const tagKey = "method";
+const tagKeys = ["method", "status"];
 
-// Register the view.
-const latencyView = stats.createView(
+// Create the view.
+stats.createView(
   "demo/latency",
   mLatencyMs,
   AggregationType.DISTRIBUTION,
-  [tagKey],
+  tagKeys,
   "The distribution of the repl latencies",
   // Latency in buckets:
   // [>=0ms, >=25ms, >=50ms, >=75ms, >=100ms, >=200ms, >=400ms, >=600ms, >=800ms, >=1s, >=2s, >=4s, >=6s]
   [0, 25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000]
 );
 
-// Register the view.
-const lineCountView = stats.createView(
+// Create the view.
+stats.createView(
   "demo/lines_in",
   mLineLengths,
   AggregationType.COUNT,
-  [tagKey],
+  tagKeys,
   "The number of lines from standard input"
 );
 
-// Register the view.
-const lineLengthView = stats.createView(
+// Create the view.
+stats.createView(
   "demo/line_lengths",
   mLineLengths,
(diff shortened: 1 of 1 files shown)


examples/stats/exporter/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Run it
+
+```
+node ./prometheus.js
+```
+
+You can observe the Prometheus Metrics Exporter that is created by the sample:
+```
+http://localhost:9465/metrics
```
mayurkale22: ```s/9465/9464```, we have mentioned 9464 port in example.

examples/stats/exporter/prometheus.js
```diff
@@ -14,4 +14,172 @@
...
+  startNanoseconds = endNanoseconds;
+});
+
+/**
+ * The default export interval is 60 seconds.
+ * The thread with must live for at least the interval past any
+ * metrics that must be collected, or some risk being lost if they are recorded
+ * after the last export.
+ */
+setTimeout(function() {
```
mayurkale22: I think we should remove this block. This is not applicable for Prometheus exporter.

examples/stats/exporter/README.md
```diff
@@ -0,0 +1,89 @@
...
+--role=roles/monitoring.metricWriter
+
+gcloud iam service-accounts keys create ${FILE} \
+--iam-account=${ACCOUNT}@${PROJECT}.iam.gserviceaccount.com \
+--project=${PROJECT}
+```
+
+### Run it
+
+```
```
mayurkale22: Do you think we should install cmd here?
```
npm install @opencensus/core
npm install @opencensus/exporter-stackdriver
```

examples/stats/exporter/prometheus.js
DazWilkin: Fixed: 643941cdf7d80e5064e00bdf23cbccae117a519f

examples/stats/exporter/README.md
DazWilkin: Fixed by adding a general `package.json` and instructions: 713b4d2fcac33d5a72fdd466f4729402f3acf668

Added Dockerfiles and instructions too.

mayurkale22:
@DazWilkin thanks for the PR. Somehow build is failing since last 2 days on node6.

examples/stats/exporter/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=6.0"
+  },
+  "files": [],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "jshint": "^2.9.7"
```
mayurkale22: Can we use ```gts``` here? This is what we use everywhere..

DazWilkin: Fixed: 42f43e11c81d025057cfc57c5217aec9095d2aff

Issue #287, opened 2019-01-16
No need to `stats.registerView(view)`


DazWilkin:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
```
0.0.8
```
### What version of Node are you using?
```
node --version
v10.13.0
```
### What did you do?
If possible, provide a recipe for reproducing the error.

Attempting to run `examples/stackdriver.js`

### What did you expect to see?

Metrics should not be exported unless they are captured by Views.

In the example, although the views are created, they're not registered, yet the sample works.

This works but it ought not:
```
// Register the view.
const latencyView = stats.createView(
  "demo/latency",
  mLatencyMs,
  AggregationType.DISTRIBUTION,
  [tagKey],
  "The distribution of the repl latencies",
  // Latency in buckets:
  // [>=0ms, >=25ms, >=50ms, >=75ms, >=100ms, >=200ms, >=400ms, >=600ms, >=800ms, >=1s, >=2s, >=4s, >=6s]
  [0, 25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000]
);
```

### What did you see instead?

The code should (IIUC) look like this:

```
// Create & Register the view.
const latencyView = stats.createView(
  "demo/latency",
  mLatencyMs,
  AggregationType.DISTRIBUTION,
  [tagKey],
  "The distribution of the repl latencies",
  // Latency in buckets:
  // [>=0ms, >=25ms, >=50ms, >=75ms, >=100ms, >=200ms, >=400ms, >=600ms, >=800ms, >=1s, >=2s, >=4s, >=6s]
  [0, 25, 50, 75, 100, 200, 400, 600, 800, 1000, 2000, 4000, 6000]
);
stats.registerView(latencyView);
```

### Additional context

With the existing code, after the `stats.createView`, `stats.registeredViews` creates a view (but I think it should not [yet]) called `repl/latency`, the view is named after the measure.

mayurkale22:
Currently, ```createView()``` method is combined with ```registerView()``` (internal call). I have opened an issue to deperecate this behavior. As you mentioned, ideally user should call ```registerView()``` explicitly to listen to new measurements in its measure.

mayurkale22:
The fix will be available in the new version (0.0.9), closing this one. @DazWilkin thanks for the issue.

Issue #270, opened 2019-01-09
Add instrumentation for gRPC on propagating tags and recording gRPC stats 


mayurkale22:
Provide instrumentation for gRPC Client and Server to record [gRPC stats](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md).  Interceptors should also propagate Tags.

mayurkale22:
**Specs are available here:** https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/gRPC.md

**Equivalent counterpart in Java**: https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics

**Equivalent counterpart in Go**: https://github.com/census-instrumentation/opencensus-go/blob/2269ed7aa5777d1157c9674a6079207a0f32b68a/plugin/ocgrpc/stats_common.go, https://github.com/census-instrumentation/opencensus-go/blob/2269ed7aa5777d1157c9674a6079207a0f32b68a/plugin/ocgrpc/server_metrics.go, https://github.com/census-instrumentation/opencensus-go/blob/2269ed7aa5777d1157c9674a6079207a0f32b68a/plugin/ocgrpc/client_metrics.go

vigneshtdev:
should i modify opencensus-intrumentation-grpc or create a new package?

vigneshtdev:
what is equivalent of `Stats.getViewManager()`(https://github.com/census-instrumentation/opencensus-java/blob/master/contrib/grpc_metrics/src/main/java/io/opencensus/contrib/grpc/metrics/RpcViews.java#L168) in node package?

mayurkale22:
```ViewManager``` in Java is for registering views for collecting stats. In Node, you can do that using ```Stats``` API itself. (example: ```stats.registerView(view)```).

vigneshtdev:
but there is no way to get views in node package

mayurkale22:
[...]

Issue #269, opened 2019-01-09
Add instrumentation for HTTP on propagating tags and recording HTTP stats


mayurkale22:
Provide instrumentation for HTTP Client and Server to record [HTTP stats](https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md).  Interceptors should also propagate Tags.

PR #291, opened 2019-01-17
 Make Stats singleton (globalStats), separate registerView from createView

### LINKED ISSUE #290: Remove registerView hook from createView API
This is related to #287

mayurkale22:
This PR is based on #276.

This PR contains two breaking changes:
1. The ```new Stats()``` has been deprecated on Stats class. The global singleton ```globalStats``` object should be used instead.
2. ```registerView()``` hook has been removed from ```createView()``` API. This is related to https://github.com/census-instrumentation/opencensus-node/issues/287.

Fixes #290 #276

### DIFF
--- a/packages/opencensus-core/src/index.ts
+++ b/packages/opencensus-core/src/index.ts
@@ -66,7 +66,6 @@ export * from './exporters/console-exporter';
 // STATS CLASSES
 
 // classes
-export * from './stats/stats';
 export * from './stats/view';
 export * from './stats/recorder';
 export * from './stats/bucket-boundaries';
@@ -94,3 +93,10 @@ export * from './metrics/metric-registry';
 // GAUGES CLASSES
 export * from './metrics/gauges/derived-gauge';
 export * from './metrics/gauges/gauge';
+
+
+// Stats singleton instance
+import {BaseStats} from './stats/stats';
+import {Stats} from './stats/types';
+const globalStats: Stats = BaseStats.instance;
+export {globalStats};
--- a/packages/opencensus-core/src/stats/stats.ts
+++ b/packages/opencensus-core/src/stats/stats.ts
@@ -17,7 +17,6 @@
 import * as defaultLogger from '../common/console-logger';
 import * as loggerTypes from '../common/types';
 import {StatsEventListener} from '../exporters/types';
-import {MetricProducer} from '../metrics/export/types';
 import {Metric} from '../metrics/export/types';
 import {Metrics} from '../metrics/metrics';
 
@@ -44,7 +43,7 @@ export class BaseStats implements Stats {
 
     // Create a new MetricProducerForStats and register it to
     // MetricProducerManager when Stats is initialized.
-    const metricProducer: MetricProducer = new MetricProducerForStats(this);
+    const metricProducer = new MetricProducerForStats(this);
     Metrics.getMetricProducerManager().add(metricProducer);
   }
 
@@ -54,8 +53,7 @@ export class BaseStats implements Stats {
   }
 
   /**
-   * Registers a view to listen to new measurements in its measure. Prefer using
-   * the method createView() that creates an already registered view.
+   * Registers a view to listen to new measurements in its measure.
    * @param view The view to be registered
    */
   registerView(view: View): void {
--- a/packages/opencensus-core/src/stats/stats.ts
+++ b/packages/opencensus-core/src/stats/stats.ts
@@ -17,18 +17,23 @@
 import * as defaultLogger from '../common/console-logger';
 import * as loggerTypes from '../common/types';
 import {StatsEventListener} from '../exporters/types';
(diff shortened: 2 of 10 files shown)


packages/opencensus-core/src/index.ts
```diff
@@ -90,3 +89,10 @@ export * from './metrics/metric-registry';
 // GAUGES CLASSES
 export * from './metrics/gauges/derived-gauge';
 export * from './metrics/gauges/gauge';
+
+
+// Stats singleton instance
+import {BaseStats} from './stats/stats';
+import {Stats} from './stats/types';
+const globalStats: Stats = BaseStats.instance;
```
draffensperger: Is the `: Stats` type specifier needed here?

Those are useful for enforcing the type assignment to a structural type like `const x: Y = {...}`, but since `BaseStats.instance` already has a type I would think it isn't needed.

packages/opencensus-core/src/stats/stats.ts
```diff
@@ -37,21 +42,23 @@ export class Stats {
...
     this.logger = logger.logger();
 
-    // TODO (mayurkale): Decide how to inject MetricProducerForStats.
-    // It should be something like below, but looks like not the right place.
-
     // Create a new MetricProducerForStats and register it to
     // MetricProducerManager when Stats is initialized.
-    // const metricProducer: MetricProducer = new MetricProducerForStats(this);
-    // Metrics.getMetricProducerManager().add(metricProducer);
+    const metricProducer: MetricProducer = new MetricProducerForStats(this);
```
draffensperger: Similar here on not needing the `: MetricProducer` type annotation since it's implied.

packages/opencensus-core/src/stats/stats.ts
```diff
@@ -37,21 +42,23 @@ export class Stats {
...
+    return this.singletonInstance || (this.singletonInstance = new this());
   }
 
   /**
    * Registers a view to listen to new measurements in its measure. Prefer using
    * the method createView() that creates an already registered view.
    * @param view The view to be registered
    */
-  registerView(view: View) {
+  registerView(view: View): void {
```
draffensperger: Nit: do we need to include `: void` here? (Same comment applies to the other methods below)

packages/opencensus-core/src/stats/types.ts
```diff
@@ -14,8 +14,77 @@
...
+   * aggregation type
+   */
+  createView(
+      name: string, measure: Measure, aggregation: AggregationType,
+      tagKeys: string[], description: string,
+      bucketBoundaries?: number[]): View;
+
+  /**
+   * Registers a view to listen to new measurements in its measure. Prefer using
+   * the method createView() that creates an already registered view.
```
draffensperger: Is this comment about creating an already registered view still true given the change above to remove the `registerView` call in the `createView` implementation above?

packages/opencensus-core/src/stats/types.ts
```diff
@@ -14,8 +14,77 @@
...
+   * @param view The view to be registered
+   */
+  registerView(view: View): void;
```
draffensperger: Do we need `: void` here? (Same comment applies to `registerExporter` below).

mayurkale22:
> What's the general motivation for the breaking change?

1. To make things consistent across all libraries before GA release.
2. ```globalStats``` follows the pattern of ```Tracing```.
3. registerView should not be part of createView, currently this is not a big problem. But we planned to add default stats support for HTTP (#269) and gRPC (#270) plugin. At that time, if you don't have separate API for createView and registerView, library will start capturing and recording data without users attention.

packages/opencensus-core/src/stats/stats.ts
draffensperger: OK, I think that's fine.

draffensperger:
Thanks for the explanation. Is the plan to go directly to GA or will there be an intermediate Beta release?

mayurkale22:
> Thanks for the explanation. Is the plan to go directly to GA or will there be an intermediate Beta release?

I am not 100% sure about this. All the work related to GA Release / Feature Parity work effort available here : https://github.com/census-instrumentation/opencensus-node/issues?q=is%3Aopen+is%3Aissue+label%3AGA-Effort

packages/opencensus-core/src/stats/stats.ts
mayurkale22: done.

packages/opencensus-core/src/stats/types.ts
mayurkale22: Good catch, removed obsolete comment.

packages/opencensus-core/src/index.ts
mayurkale22: @draffensperger Are you ok with my comment or you have better approach to handle this?

Issue #268, opened 2019-01-09
Stats: apply restriction on TagKey and TagValue


mayurkale22:
The TagKey name must meet the following requirements:

1. Must contain only printable ASCII (codes between 32 and 126 inclusive)
2. Must have length greater than zero and less than 256.
3. Must not be empty.

The TagValue name must meet the following requirements:

1. It MUST contain only printable ASCII (codes between 32 and 126)

Specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md#tagkey

vigneshtdev:
hey i would like to help with this issue. In opencensus java it had a package called tag where different tagvalue & tagkey can be configured. Where should I look for it in this package?

mayurkale22:
@vigneshtdev Thanks for your interest in the project, all contributions are welcomed and highly appreciated. Currently we dont have separate package for [Tags](https://github.com/census-instrumentation/opencensus-specs/tree/master/tags) api. I think we should create new one same as [Java](https://github.com/census-instrumentation/opencensus-java/tree/master/api/src/main/java/io/opencensus/tags) and [Python](https://github.com/census-instrumentation/opencensus-python/tree/master/opencensus/tags). WDYT?

vigneshtdev:
I think that it is the correct way to go, as it will make implementation across all languages more consistent. However it would require major refactoring in both src & tst

vigneshtdev:
Or else if such refactoring in unnecessary we can modify the existing invalidTags method. Currenlty it only check if the tagkey & tagvalue have printable characters.

songy23:
Fixed in #280.

PR #295, opened 2019-01-17
[core] Add Tags API

mayurkale22:
This PR is based on specs : https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md#tagkey

### DIFF
--- a/packages/opencensus-core/src/tags/tag-map.ts
+++ b/packages/opencensus-core/src/tags/tag-map.ts
@@ -22,13 +22,9 @@ export class TagMap {
   // A map mapping TagKey to to its respective TagValue.
   private readonly registeredTags: Map<TagKey, TagValue> = new Map();
 
-  constructor() {}
-
   /**
    * Inserts a key and value in the map if the map does not already contain the
    * key.
-   * @param tagKey The tag key.
-   * @param tagValue The tag value.
    */
   insert(tagKey: TagKey, tagValue: TagValue): void {
     if (!isValidTagKey(tagKey)) {
@@ -44,21 +40,12 @@ export class TagMap {
     }
   }
 
-  /**
-   * Deletes a tag from the map if the key is in the map.
-   * @param tagKey The tag key.
-   */
+  /** Deletes a tag from the map if the key is in the map. */
   delete(tagKey: TagKey): void {
-    if (this.registeredTags.has(tagKey)) {
-      this.registeredTags.delete(tagKey);
-    }
+    this.registeredTags.delete(tagKey);
   }
 
-  /**
-   * Updates the map by updating the value of a key.
-   * @param tagKey A tag key to be updated.
-   * @param tagValue The value to update the key to in the map.
-   */
+  /** Updates the map by updating the value of a key. */
   update(tagKey: TagKey, tagValue: TagValue): void {
     if (this.registeredTags.has(tagKey)) {
       if (!isValidTagValue(tagValue)) {
--- a/packages/opencensus-core/src/tags/tag-map.ts
+++ b/packages/opencensus-core/src/tags/tag-map.ts
@@ -22,11 +22,8 @@ export class TagMap {
   // A map mapping TagKey to to its respective TagValue.
   private readonly registeredTags: Map<TagKey, TagValue> = new Map();
 
-  /**
-   * Inserts a key and value in the map if the map does not already contain the
-   * key.
-   */
-  insert(tagKey: TagKey, tagValue: TagValue): void {
+  /** Adds the key/value pair regardless of whether the key is present. */
+  set(tagKey: TagKey, tagValue: TagValue): void {
     if (!isValidTagKey(tagKey)) {
       throw Error(`Invalid TagKey name: ${tagKey.name}`);
     }
@@ -34,25 +31,12 @@ export class TagMap {
     if (!isValidTagValue(tagValue)) {
       throw Error(`Invalid TagValue: ${tagValue.value}`);
     }
-
-    if (!this.registeredTags.has(tagKey)) {
-      this.registeredTags.set(tagKey, tagValue);
-    }
+    this.registeredTags.set(tagKey, tagValue);
   }
 
   /** Deletes a tag from the map if the key is in the map. */
   delete(tagKey: TagKey): void {
     this.registeredTags.delete(tagKey);
-  }
-
-  /** Updates the map by updating the value of a key. */
-  update(tagKey: TagKey, tagValue: TagValue): void {
-    if (this.registeredTags.has(tagKey)) {
-      if (!isValidTagValue(tagValue)) {
-        throw Error(`Invalid TagValue: ${tagValue.value}`);
-      }
-      this.registeredTags.set(tagKey, tagValue);
-    }
   }
 
   /** Gets the tags map. */
--- a/packages/opencensus-core/src/tags/tag-map.ts
+++ b/packages/opencensus-core/src/tags/tag-map.ts
@@ -1,0 +1,75 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 the "License";
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


packages/opencensus-core/src/tags/tag-map.ts
```diff
@@ -0,0 +1,75 @@
...
+
+import {TagKey, TagValue} from './types';
+import {isValidTagKey, isValidTagValue} from './validation';
+
+/** TagMap is maps of TagKey -> TagValue */
+export class TagMap {
+  // A map mapping TagKey to to its respective TagValue.
+  private readonly registeredTags: Map<TagKey, TagValue> = new Map();
+
+  constructor() {}
```
draffensperger: Can this empty constructor be removed?

packages/opencensus-core/src/tags/tag-map.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  /**
+   * Inserts a key and value in the map if the map does not already contain the
+   * key.
+   * @param tagKey The tag key.
```
draffensperger: These `@param` comments seem to be fully redundant with the parameter names. Can they be removed?

This comment applies to `insert`, `delete` and `update` below too.

packages/opencensus-core/src/tags/tag-map.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.registeredTags.set(tagKey, tagValue);
+    }
+  }
+
+  /**
+   * Deletes a tag from the map if the key is in the map.
+   * @param tagKey The tag key.
+   */
+  delete(tagKey: TagKey): void {
+    if (this.registeredTags.has(tagKey)) {
```
draffensperger: Do you need call to `has`? `Map.delete` just returns `false` if the key was not deleted but otherwise is a no-op.

packages/opencensus-core/src/tags/tag-map.ts
```diff
@@ -0,0 +1,75 @@
...
+   * Updates the map by updating the value of a key.
+   * @param tagKey A tag key to be updated.
+   * @param tagValue The value to update the key to in the map.
+   */
+  update(tagKey: TagKey, tagValue: TagValue): void {
+    if (this.registeredTags.has(tagKey)) {
+      if (!isValidTagValue(tagValue)) {
+        throw Error(`Invalid TagValue: ${tagValue.value}`);
+      }
+      this.registeredTags.set(tagKey, tagValue);
```
draffensperger: Do you need to check for a valid tag value here?

packages/opencensus-core/src/tags/tag-map.ts
```diff
@@ -0,0 +1,75 @@
...
+      this.registeredTags.delete(tagKey);
+    }
+  }
+
+  /**
+   * Updates the map by updating the value of a key.
+   * @param tagKey A tag key to be updated.
+   * @param tagValue The value to update the key to in the map.
+   */
+  update(tagKey: TagKey, tagValue: TagValue): void {
```
draffensperger: Why have separate `insert` and `update` functions and not just a `set` function that does insert and update?

That would enable combining both functions and you also would not need the `if (this.registeredTags.has(tagKey)) {` check.

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -0,0 +1,56 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {TagKey, TagValue} from './types';
+
+// String that has only printable characters
+const invalidString = /[^\u0020-\u007e]/;
```
draffensperger: Nit: what would you think about calling this `nonPrintableCharsRegex`? If you did that could you remove the comment? I find the `String` in the comment and name a bit confusing.

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -0,0 +1,56 @@
...
+
+// Max Length of a TagKey
+const MAX_LENGTH = 255;
```
draffensperger: Nit: could this be `TAG_KEY_MAX_LENGTH` and then remove the comment?

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Determines whether the given String is a valid tag key.
+ *
+ * @param tagKey the name to be validated.
+ * @return whether the name is valid.
+ */
+export function isValidTagKey(tagKey: TagKey): boolean {
+  if (tagKey === null || typeof tagKey === 'undefined') {
```
draffensperger: Could this be compacted to `if (!tagKey) return false;`? Given that `tagKey` needs to be an object with a valid `name`, all falsely values should be invalid.

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -0,0 +1,56 @@
...
+  return name && isLegalCharacters(name) && name.length <= MAX_LENGTH;
+}
+
+/**
+ * Determines whether the given String is a valid tag value.
+ *
+ * @param tagValue the name to be validated.
+ * @return whether the name is valid.
+ */
+
```
draffensperger: Nit: should this extra line be removed?

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -0,0 +1,56 @@
...
+export function isValidTagValue(tagValue: TagValue): boolean {
+  if (tagValue === null || typeof tagValue === 'undefined') {
```
draffensperger: Similar question here on whether it makes sense to use falsiness as a check here.

packages/opencensus-core/test/test-tag-map.ts
```diff
@@ -0,0 +1,99 @@
...
+
+    it('should not insert duplicate tagkey and tagvalue', () => {
+      tagMap.insert(key1, value1);
+      const tags = tagMap.tags;
+      assert.equal(tags.size, 1);
+      tagMap.insert(key1, value1);
+      assert.equal(tags.size, 1);
+    });
+  });
+  describe('update()', () => {
```
draffensperger: Nit, should this be `delete()`?

packages/opencensus-core/src/tags/tag-map.ts
songy23: +1 to this. In Java we use this approach in [`TagContextBuilder.put()`](https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/tags/TagContextBuilderImpl.java#L40).

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -0,0 +1,42 @@
...
+      tagKey.name.length <= TAG_KEY_MAX_LENGTH;
+}
+
+/** Determines whether the given String is a valid tag value. */
+export function isValidTagValue(tagValue: TagValue): boolean {
+  if (!tagValue) {
+    return false;
+  }
+  return isPrintableString(tagValue.value) &&
+      tagValue.value.length <= TAG_KEY_MAX_LENGTH;
```
songy23: `TagValue` doesn't have a length constraint (https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md#tagvalue).

mayurkale22: I have compared with [Java](https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/tags/TagValue.java#L77) and [Python](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/tags/validation.py#L42), I see we have enforced length constaint on ```TagValue```. Let me know if you still want me to remove it.

packages/opencensus-core/src/tags/tag-map.ts
mayurkale22: Done, thanks for the reference.

packages/opencensus-core/src/tags/validation.ts
songy23: I see. We can leave the constraint for now, it's easy to relax it in the future.

mayurkale22:
@vigneshtdev https://github.com/census-instrumentation/opencensus-node/issues/268#issuecomment-453720884 FYI, we have decided to add separate package for Tags API.

mayurkale22:
@draffensperger thanks for the reviews, I have fixed all the mentioned comments.

PR #299, opened 2019-01-18
Add `opencensus-resource-util` package

mayurkale22:
Add util to automatically detect monitored resources based the environment that the application is running in, according to [OpenCensus MonitoredResource specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/utils/MonitoredResource.md). This util should be independent of specific exporters, since it's supposed to be share between Stackdriver Stats exporter and Stackdriver Trace exporter.

Counterpart in Java: https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/monitored_resource_util.

This PR contains:
1. Initial package skeleton.
2. resource label constants.
3. export core resource types and class from core package.

### DIFF
--- a/packages/opencensus-resource-util/src/resource-labels.ts
+++ b/packages/opencensus-resource-util/src/resource-labels.ts
@@ -15,65 +15,57 @@
  */
 
 /** Constants for collecting resource information. */
-// tslint:disable-next-line:variable-name
-export const AWSResourceKeyConstants = {
-  /** AWS key that represents a type of the resource. */
-  AWS_EC2_INSTANCE_TYPE: 'aws.com/ec2/instance',
 
-  /** AWS key that represents a region for the VM. */
-  AWS_REGION_KEY: 'aws.com/ec2/region',
+/** AWS key that represents a type of the resource. */
+export const AWS_EC2_INSTANCE_TYPE = 'aws.com/ec2/instance';
 
-  /** AWS key that represents the AWS account number for the VM. */
-  AWS_ACCOUNT_KEY: 'aws.com/ec2/account_id',
+/** AWS key that represents a region for the VM. */
+export const AWS_REGION_KEY = 'aws.com/ec2/region';
 
-  /** AWS key that represents the VM instance identifier assigned by AWS. */
-  AWS_INSTANCE_ID_KEY: 'aws.com/ec2/instance_id',
+/** AWS key that represents the AWS account number for the VM. */
+export const AWS_ACCOUNT_KEY = 'aws.com/ec2/account_id';
 
-  /** AWS key that represents a prefix for region value. */
-  AWS_REGION_VALUE_PREFIX: 'aws:'
-};
+/** AWS key that represents the VM instance identifier assigned by AWS. */
+export const AWS_INSTANCE_ID_KEY = 'aws.com/ec2/instance_id';
 
-// tslint:disable-next-line:variable-name
-export const GCPResourceKeyConstants = {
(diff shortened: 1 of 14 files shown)


packages/opencensus-resource-util/src/resource-labels.ts
```diff
@@ -0,0 +1,79 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/** Constants for collecting resource information. */
+// tslint:disable-next-line:variable-name
```
draffensperger: Why not name this `AwsResourceKeyConstants` following the variable naming guide?

That rule is ultimately based on the Google JS style guide I believe: https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined

packages/opencensus-resource-util/src/resource-labels.ts
```diff
@@ -0,0 +1,79 @@
...
+  AWS_ACCOUNT_KEY: 'aws.com/ec2/account_id',
+
+  /** AWS key that represents the VM instance identifier assigned by AWS. */
+  AWS_INSTANCE_ID_KEY: 'aws.com/ec2/instance_id',
+
+  /** AWS key that represents a prefix for region value. */
+  AWS_REGION_VALUE_PREFIX: 'aws:'
+};
+
+// tslint:disable-next-line:variable-name
```
draffensperger: Similar, here, can this be called `GcpResourceKeyConstants`?

packages/opencensus-resource-util/src/resource-labels.ts
```diff
@@ -0,0 +1,79 @@
...
+   * GCE.
+   */
+  GCP_INSTANCE_ID_KEY: 'cloud.google.com/gce/instance_id',
+
+  /** GCP GCE key that represents the GCE zone in which the VM is running. */
+  GCP_ZONE_KEY: 'cloud.google.com/gce/zone'
+};
+
+// tslint:disable-next-line:variable-name
+export const K8SResourceKeyConstants = {
```
draffensperger: Can this be `K8sResourceKeyConstants` so that we won't need the variable name lint disabling?

packages/opencensus-resource-util/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "pretty": true,
+    "module": "commonjs",
+    "target": "es6",
+    "strictNullChecks": false,
```
draffensperger: Why set `strictNullChecks` to `false`? I would think `true` would give us stronger type guarantees and more safety as the new library is developed.

packages/opencensus-resource-util/src/resource-labels.ts
mayurkale22: Agreed, I wanted to group labels together based on types (GCP, AWS, K8s). Removed namespace https://github.com/census-instrumentation/opencensus-node/pull/299/commits/d994ca9cc34e554b24585c7eed7ab79ef184466a

packages/opencensus-resource-util/tsconfig.json
mayurkale22: SGTM, done.

Issue #301, opened 2019-01-21
Jaeger exporter require wrapper does not play nicely with Webpack


tstirrat15:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
Master branch.

### What version of Node are you using?
Node 8

### What did you do?
Our project bundles both frontend and backend code using [Webpack](https://webpack.js.org/). This means that while [this code](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-jaeger/src/jaeger-driver/index.ts#L22) is fine on Node, Webpack has no idea what to do with it, because it uses an internal `require` wrapper that doesn't work in this context.

### What did you expect to see?
That this code can be bundled just fine using webpack.

### What did you see instead?
I get a runtime error:

```
path.js:28
    throw new TypeError('Path must be a string. Received ' + inspect(path));
    ^

TypeError: Path must be a string. Received 1122
    at assertPath (path.js:28:11)
    at Object.dirname (path.js:1346:5)
    at Object.<anonymous> (/home/me/project/build/server.js:72478:24)
    at __webpack_require__ (/home/me/project/build/server.js:27:30)
    at Object.module.exports.Object.defineProperty.value (/home/me/project/build/server.js:72297:25)
    at __webpack_require__ (/home/me/project/build/server.js:27:30)
    at Object.<anonymous> (/home/me/project/build/server.js:43500:10)
    at __webpack_require__ (/home/me/project/build/server.js:27:30)
    at Module.<anonymous> (/home/me/project/build/server.js:67332:85)
    at Module.module.exports.m (/home/me/project/build/server.js:67452:30)
```

tstirrat15:
This was closed by #302.

PR #302, opened 2019-01-21
Jaeger exporter: remove require wrapper and make requires direct

tstirrat15:
## Motivation
See #301.

It seems like this was done for developer convenience, and this function is only used in this file. While I don't have a problem with that in principle, it breaks things for Webpack.

## Changes
Remove the require wrapper and make the requires direct

## Testing
See that the build still builds correctly.


tstirrat15:
I signed it!

mayurkale22:
Please rebase and fix the build

tstirrat15:
@mayurkale22 ready for rereview.

PR #307, opened 2019-01-24
Update stats, zpages, prometheus packages with new Tags API

### LINKED ISSUE #305: Update Stats, other packages with new Tags API
Use newly added tags api (https://github.com/census-instrumentation/opencensus-node/tree/master/packages/opencensus-core/src/tags) in stats.

Change 
```tagKeys: string[]```  -> ```tagKeys: TagKey[]``` and interface 
```Tags``` -> ```TagMap```.

### PR DESCRIPTION
**Pull request by mayurkale22**
Fixes #305 

This PR contains breaking API changes, highlighted in the CHANGELOG with old and new code.

### DIFF
--- a/packages/opencensus-core/src/stats/recorder.ts
+++ b/packages/opencensus-core/src/stats/recorder.ts
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+import {TagKey, TagValue} from '../tags/types';
 import {AggregationData, AggregationType, CountData, DistributionData, LastValueData, Measurement, MeasureType, SumData} from './types';
 
 export class Recorder {
@@ -38,6 +39,20 @@ export class Recorder {
       default:
         return this.addToLastValue(aggregationData, value);
     }
+  }
+
+  /** Gets the tag values from tags and columns */
+  static getTagValues(tags: Map<TagKey, TagValue>, columns: TagKey[]):
+      TagValue[] {
+    const tagValues: TagValue[] = [];
+
+    // ignore not found key values.
+    columns.forEach((tagKey) => {
+      if (tags.has(tagKey)) {
+        tagValues.push(tags.get(tagKey));
+      }
+    });
+    return tagValues;
   }
 
   private static addToDistribution(
--- a/packages/opencensus-core/src/stats/recorder.ts
+++ b/packages/opencensus-core/src/stats/recorder.ts
@@ -17,6 +17,8 @@
 import {TagKey, TagValue} from '../tags/types';
 import {AggregationData, AggregationType, CountData, DistributionData, LastValueData, Measurement, MeasureType, SumData} from './types';
 
+const UNKNOWN_TAG_VALUE: TagValue = null;
+
 export class Recorder {
   static addMeasurement(
       aggregationData: AggregationData,
@@ -44,15 +46,10 @@ export class Recorder {
   /** Gets the tag values from tags and columns */
   static getTagValues(tags: Map<TagKey, TagValue>, columns: TagKey[]):
       TagValue[] {
-    const tagValues: TagValue[] = [];
-
-    // ignore not found key values.
-    columns.forEach((tagKey) => {
-      if (tags.has(tagKey)) {
-        tagValues.push(tags.get(tagKey));
-      }
-    });
-    return tagValues;
+    return columns.map(
+        (tagKey) =>
+            (tags.get(tagKey) ||
+             /** replace not found key values by null. */ UNKNOWN_TAG_VALUE));
   }
 
   private static addToDistribution(
--- a/packages/opencensus-core/src/stats/view.ts
+++ b/packages/opencensus-core/src/stats/view.ts
@@ -118,7 +118,6 @@ export class BaseView implements View {
   recordMeasurement(measurement: Measurement, tags: TagMap) {
     const tagValues = Recorder.getTagValues(tags.tags, this.columns);
     const encodedTags = this.encodeTagValues(tagValues);
-
     if (!this.tagValueAggregationMap[encodedTags]) {
       this.tagValueAggregationMap[encodedTags] =
           this.createAggregationData(tagValues);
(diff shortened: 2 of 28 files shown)

### CODE REVIEW

File: packages/opencensus-core/src/stats/recorder.ts [hunk 1]
```diff
@@ -40,6 +41,20 @@ export class Recorder {
     }
   }
 
+  /** Gets the tag values from tags and columns */
+  static getTagValues(tags: Map<TagKey, TagValue>, columns: TagKey[]):
+      TagValue[] {
+    const tagValues: TagValue[] = [];
+
+    // ignore not found key values.
```
songy23: For tag keys that are not found, we should use `null` or `undefined` as the tag value. See https://github.com/census-instrumentation/opencensus-java/blob/5a97afdefdb2d9f4213264d91fa67b646adbf5d9/impl_core/src/main/java/io/opencensus/implcore/stats/RecordUtils.java#L86-L87.

The tag values returned from this method should have the same size as tag keys.

packages/opencensus-core/src/stats/recorder.ts
```diff
@@ -40,6 +41,20 @@ export class Recorder {
     }
   }
 
+  /** Gets the tag values from tags and columns */
+  static getTagValues(tags: Map<TagKey, TagValue>, columns: TagKey[]):
```
draffensperger: Why is this a `static` method of the class rather than an exported function in the file? Is that because the other similar functions are static methods of the class?

By default in TypeScript, I would prefer exported functions in a module rather than static methods of a class.

packages/opencensus-core/src/stats/view.ts
```diff
@@ -294,7 +246,22 @@ export class BaseView implements View {
...
    */
-  getSnapshot(tags: Tags): AggregationData {
-    return this.rows[this.encodeTags(tags)];
+  getSnapshot(tagValues: TagValue[]): AggregationData {
+    return this.tagValueAggregationMap[this.encodeTagValues(tagValues)];
+  }
+
+  /** Determines whether the given TagKeys are valid. */
+  private validateTagKeys(tagKeys: TagKey[]): TagKey[] {
+    const tagKeysCopy = Object.assign([], tagKeys);
```
draffensperger: Why does this need to be copied?

packages/opencensus-core/src/tags/validation.ts
```diff
@@ -30,7 +30,7 @@ export function isValidTagKey(tagKey: TagKey): boolean {
 
 /** Determines whether the given String is a valid tag value. */
 export function isValidTagValue(tagValue: TagValue): boolean {
-  if (!tagValue) {
+  if (!tagValue || tagValue.value === null || tagValue.value === undefined) {
```
draffensperger: Why add ` || tagValue.value === null || tagValue.value === undefined`? Because both `null` and `undefined` are falsey, wouldn't those be captured within the `!tagValue`?

packages/opencensus-exporter-prometheus/src/prometheus-stats.ts
```diff
@@ -209,8 +220,11 @@ export class PrometheusStatsExporter implements StatsEventListener {
    * @param view View used to get the DistributionData
    * @param tags Tags used to get the DistributionData
    */
-  private getBoundaries(view: View, tags: Tags): number[] {
-    const data = view.getSnapshot(tags) as DistributionData;
+  private getBoundaries(view: View, labels: labelValues): number[] {
+    const tagValues = Object.keys(labels).map((label) => {
+      return {value: String(labels[label])};
```
draffensperger: Optional: similar comment here on not needing to use a `return` if you put parens around the object

packages/opencensus-exporter-prometheus/src/prometheus-stats.ts
```diff
@@ -209,8 +220,11 @@ export class PrometheusStatsExporter implements StatsEventListener {
...
+    });
+    const data = view.getSnapshot(tagValues) as DistributionData;
```
draffensperger: Why is the `as DistributionData` needed?

packages/opencensus-core/src/stats/recorder.ts
mayurkale22: > By default in TypeScript, I would prefer exported functions in a module rather than static methods of a class.

I agreed, but currently all other methods are static in the recorder class. I wanted to keep the same notion.

packages/opencensus-core/src/stats/recorder.ts
mayurkale22: This project has ```strictNullChecks``` set to false in tsconfig, hence didn't add null for missing tag values.

packages/opencensus-core/src/stats/view.ts
mayurkale22: This is to make a shallow copy of tagKeys to protect internal functionality, in case any update on  original data.

packages/opencensus-core/src/stats/recorder.ts
songy23: The caveat here is the tag values returned must match the order of tag keys. If we skip any tag values then they became mismatched.

One example: a view is defined to break down latencies by method and caller. The tag keys are `[method, caller]` in order. If a `record` is called with only tag `caller:c1`, this method will return tag values `[c1]`. Then `c1` will be matched with `method`.

packages/opencensus-core/src/tags/validation.ts
mayurkale22: ```!tagValue``` is not applicable to its property (``value```). Yes, both```null``` and ```undefined``` are falsey for value property.

packages/opencensus-exporter-prometheus/src/prometheus-stats.ts
mayurkale22: Only DistributionData has buckets property + ```getSnapshot``` returns ```AggregationData``` which are type of ```SumData | CountData | LastValueData | DistributionData;```

packages/opencensus-core/src/stats/view.ts
draffensperger: Oh, I see. I missed that `tagKeysCopy` is returned by the function.

PR #309, opened 2019-01-25
Pin mongodb dependency

mayurkale22:


mayurkale22:
Node 6 build started failing ever since [mongodb](https://www.npmjs.com/package/mongodb/v/3.1.13) published new version 3.1.11 (10 days before). The error was -> ```no original function _find to wrap```, looks like they have deprecated ```_find_``` cursor method from ```mongodb-core (3.1.10)``` which comes with ```mongodb >= 3.1.11```. For now, I have pinned mongodb dependency to previous stable version (3.1.10).

I think next action item should be to replace ```_find``` cursor method with ```next``` and fix the test cases accordingly.

Issue #285, opened 2019-01-16
Add TraceParams Limits


mayurkale22:
Add functionality to hold global trace parameters (like Attributes, Annotations etc.). Limits will be enforced on each span in order to prevent unbounded memory increase for long-running spans.

Specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md#traceparams

Counterpart in Java: https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/config/TraceParams.java

vigneshtdev:
I can help with this one.

mayurkale22:
@vigneshtdev Awesome, please go ahead and send the PR. 

Counterpart in Go : https://github.com/census-instrumentation/opencensus-go/pull/1015

vigneshtdev:
Meanwhile can you look at the gist I have posted for grpc instrumentation

mayurkale22:
> Meanwhile can you look at the gist I have posted for grpc instrumentation

Sure, I will do that today.

vigneshtdev:
hey in case trace parameters in tracer configuration are not specified should we default to max values or mark those fields are 0?

mayurkale22:
Closed in #311, the next part is to enforce these limits #316

PR #311, opened 2019-01-28
Add trace params

vigneshtdev:
Addresses #285. 
Unit Tests included as well.


packages/opencensus-core/src/trace/config/types.ts
```diff
@@ -70,4 +72,28 @@ export interface TracingConfig {
   logger?: Logger;
 }
 
+/** Specify the deafault Tracer Paramaters */
```
draffensperger: Can this commented code be removed?

packages/opencensus-core/src/trace/config/types.ts
```diff
@@ -37,6 +37,8 @@ export interface TracerConfig {
   logger?: Logger;
   /** A propagation instance */
   propagation?: Propagation;
+  /** Trace Parameters */
+  traceParameters?: TraceParameters;
```
draffensperger: To me, `traceParameters` feels a bit generic, and it seems like all these params are related to size. What would you think about calling this something like `traceSizeParameters`, or if that's too long, `traceSizeConfig` or something similar?

packages/opencensus-core/src/trace/config/types.ts
```diff
@@ -70,4 +72,28 @@ export interface TracingConfig {
...
+// export class TracerParams {
+
+//   static readonly DEFAULT_SPAN_MAX_NUM_ATTRIBUTES?: number = 32;
+//   static readonly DEFAULT_SPAN_MAX_NUM_ANNOTATIONS?: number = 32;
+//   static readonly DEFAULT_SPAN_MAX_NUM_MESSAGE_EVENTS?: number = 128;
+//   static readonly DEFAULT_SPAN_MAX_NUM_LINKS?: number = 32;
+
+// }
+
+export interface TraceParameters {
```
draffensperger: Could you add a top level comment for this interface as a whole?

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -46,7 +46,14 @@ export class CoreTracer implements types.Tracer {
   sampler: samplerTypes.Sampler;
   /** A configuration for starting the tracer */
   logger: loggerTypes.Logger = logger.logger();
-
+  /** maxAnnotationEvents is max number of annotation events per span  */
+  numberOfAnnontationEventsPerSpan?: number;
```
draffensperger: Rather than storing the individual values here, could it just store something like `traceParameters: TraceParameters = DEFAULT_TRACE_PARAMETERS`?

Then if we add one more parameter later on, we don't need to add it two places.

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -81,6 +88,28 @@ export class CoreTracer implements types.Tracer {
...
+      this.numberOfAttributesPerSpan =
+          TraceParamasBuilder.getNumberOfAttributesPerSpan(
+              config.traceParameters);
+      this.numberOfMessageEventsPerSpan =
+          TraceParamasBuilder.getNumberOfMessageEventsPerSpan(
+              config.traceParameters);
+      this.numberOfLinksPerSpan =
+          TraceParamasBuilder.getNumberOfLinksPerSpan(config.traceParameters);
+    } else {
+      this.numberOfAnnontationEventsPerSpan =
```
draffensperger: Could we remove this `else` block if we assign a default value to the parameters above?

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -81,6 +88,28 @@ export class CoreTracer implements types.Tracer {
     this.config = config;
     this.logger = this.config.logger || logger.logger();
     this.sampler = SamplerBuilder.getSampler(config.samplingRate);
+    if (config.traceParameters != null) {
+      this.numberOfAnnontationEventsPerSpan =
```
draffensperger: What would you think about integrating the default values right in line here rather than abstracting it into a builder class?

It could just be something like:
```
this.traceParameters.numberOfAnnontationEventsPerSpan = 
    config.traceParameters.numberOfAnnontationEventsPerSpan ||
        DEFAULT_NUMBER_OF_ANNOTATION_EVENTS_PER_SPAN
```

and similar for the other parameters.

I personally think the builder pattern, while it makes sense for other languages is not needed in TypeScript.

draffensperger:
Thanks for the contribution, great to have this moving ahead!! I left a number of nit-picky comments - feel free to explain if you disagree on any of these.

packages/opencensus-core/src/trace/config/types.ts
vigneshtdev: Renamed from `TraceParameters` to `TraceParams`

packages/opencensus-core/src/trace/model/tracer.ts
vigneshtdev: I have a question, do we assume that user configures the trace with parameters having values more than their limit?

draffensperger: Ah, I missed the logic of the builder pattern below that capped those values at their limits. If other languages have that cap, we should preserve it for Node I'd think.

In that case, I would recommend making these be top-level attributes (like you have) but be made `readonly` to prevent changes to them once they are initialized.

packages/opencensus-core/src/trace/config/types.ts
draffensperger: I saw this was marked as resolved - is the commented out code below intentional? (Or maybe just a commit not pushed yet?)

packages/opencensus-core/test/test-sampler.ts
```diff
@@ -100,4 +109,48 @@ describe('Sampler', () => {
...
+  describe('getNumberOfLinksPerSpan', () => {
+    it('should return max number of links per span', () => {
+      assert.equal(TraceParamasBuilder.getNumberOfLinksPerSpan(undefined), 32);
+    });
+    it('should return 7', () => {
+      assert.equal(
+          TraceParamasBuilder.getNumberOfLinksPerSpan(traceParameters), 7);
+    });
+  });
 });
```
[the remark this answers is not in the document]
vigneshtdev: Done.

packages/opencensus-core/src/trace/config/types.ts
```diff
@@ -70,4 +72,28 @@ export interface TracingConfig {
...
+  /**
+   * numberOfAnnontationEventsPerSpan is max number of annotation events per
+   * span
+   */
+  numberOfAnnontationEventsPerSpan?: number;
+  /** numberOfMessageEventsPerSpan is number of message events per span */
```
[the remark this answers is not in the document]
vigneshtdev: Sorry could a elaborate a bit more.

packages/opencensus-core/src/trace/sampler/sampler.ts
```diff
@@ -97,3 +99,46 @@ export class SamplerBuilder {
     return new ProbabilitySampler(probability);
   }
 }
+
+/** Builder Class of TraceParams */
+export class TraceParamasBuilder {
+  private static readonly MAX_NUMBER_OF_ANNOTATION_EVENTS_PER_SPAN = 32;
+  private static readonly MAX_NUMBER_OF_MESSAGE_EVENTS_PER_SPAN = 128;
+  private static readonly MAX_NUMBER_OF_ATTRIBUTES_PER_SPAN = 32;
```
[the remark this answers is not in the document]
vigneshtdev: Done

packages/opencensus-core/src/trace/config/types.ts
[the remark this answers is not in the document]
vigneshtdev: in the interface definition, it should just be `number of` right?

packages/opencensus-core/src/trace/sampler/sampler.ts
```diff
@@ -97,3 +99,40 @@ export class SamplerBuilder {
...
 }
+
+/** Builder Class of TraceParams */
+export class TraceParamsBuilder {
+  private static readonly MAX_NUMBER_OF_ANNOTATION_EVENTS_PER_SPAN = 32;
+  private static readonly MAX_NUMBER_OF_MESSAGE_EVENTS_PER_SPAN = 128;
+  private static readonly MAX_NUMBER_OF_ATTRIBUTES_PER_SPAN = 32;
+  private static readonly MAX_NUMBER_OF_LINKS_PER_SPAN = 32;
+
+  static getNumberOfAnnotationEventsPerSpan(traceParameters: TraceParams):
```
draffensperger: What would you think about making these functions be top level `function` exports rather than static members of a class? You could still namespace them when they are used if needed based on how they are imported.

vigneshtdev: all four functions are a part of TraceParams, right? so exporting them individually out of sampler.ts doesn't seem right. Instead, we are exporting them under TraceParams class. Just like in the specs document https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md#traceparams.

vigneshtdev: Also since we are providing default values, in case user doesn't provide them, so are those functions really worth exporting?

draffensperger: Yeah, I would be OK exporting these under a containing class in this case.

And having separate sets of constants for the defaults and maximums would be OK too I think. I'll modify my comment above to just ask that we make them top level constants for the defaults instead of properties.

packages/opencensus-core/test/test-sampler.ts
```diff
@@ -100,4 +109,31 @@ describe('Sampler', () => {
          });
        });
   });
-});
\ No newline at end of file
+  describe('getNumberOfAnnotationEventsPerSpan', () => {
+    it('should return  12', () => {
```
draffensperger: Could you make one of these test that values over the maximum get set to the maximum?

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -42,11 +44,27 @@ export class CoreTracer implements types.Tracer {
   private endedTraces: types.RootSpan[] = [];
   /** Bit to represent whether trace is sampled or not. */
   private readonly IS_SAMPLED = 0x1;
+  /** Default Limit for Annotations per span */
+  private readonly DEFAULT_SPAN_MAX_NUM_ANNOTATIONS = 32;
```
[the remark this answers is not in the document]
mayurkale22: +1

packages/opencensus-core/src/trace/model/tracer.ts
mayurkale22: I agree with @draffensperger, builder pattern is not required here.

vigneshtdev: https://github.com/census-instrumentation/opencensus-node/pull/311#discussion_r252286902

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -42,11 +44,27 @@ export class CoreTracer implements types.Tracer {
...
+    numberOfLinksPerSpan: this.DEFAULT_SPAN_MAX_NUM_LINKS,
+    numberOfMessageEventsPerSpan: this.DEFAULT_SPAN_MAX_NUM_MESSAGE_EVENTS
+  };
   /** A sampler used to make sample decisions */
   sampler: samplerTypes.Sampler;
   /** A configuration for starting the tracer */
   logger: loggerTypes.Logger = logger.logger();
-
+  /** A configuration object for trace parameters */
+  traceParams: TraceParams = this.DEFAULT_TRACE_PARAMS;
```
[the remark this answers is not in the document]
vigneshtdev: done.

mayurkale22:
@vigneshtdev is this ready to merge?

packages/opencensus-core/src/trace/config/types.ts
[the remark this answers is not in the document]
rghetia: it is a maximum number allowed. So it is a max value.

vigneshtdev:
> @vigneshtdev is this ready to merge?

Yeah you can go ahead and merge it

packages/opencensus-core/test/test-sampler.ts
mayurkale22: @vigneshtdev any update here?

vigneshtdev: It's been added already. https://github.com/census-instrumentation/opencensus-node/blob/3d8e965612d55a98c53e4947ac174af7e4db5c13/packages/opencensus-core/test/test-tracer.ts#L255

PR #312, opened 2019-01-28
Auto detect GCE, GKE and AWS EC2 resources

### LINKED ISSUE #38: Add utils to automatically detect monitored resources.
Add utils to automatically detect monitored resources based the environment that the application is running in, according to [OpenCensus MonitoredResource specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/utils/MonitoredResource.md). This util should be independent of specific exporters, since it's supposed to be share between Stackdriver Stats exporter and Stackdriver Trace exporter.

Counterpart in Java: https://github.com/census-instrumentation/opencensus-java/tree/master/contrib/monitored_resource_util.

mayurkale22:
This PR is to automatically detect monitored resources based the environment that the application is running in, according to [OpenCensus MonitoredResource specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/utils/MonitoredResource.md). This util should be independent of specific exporters, since it's supposed to be shared between the Stackdriver Stats exporter and Stackdriver Trace exporter.

I will send another PR to integrate this detected resources (```detectResource```) to the Stackdriver exporter.

Fixes #38.


packages/opencensus-resource-util/src/index.ts
```diff
@@ -13,3 +13,31 @@
...
+  const resourceFromEnv = CoreResource.createFromEnvironmentVariables();
+  resources.push(resourceFromEnv);
+
+  const resourceType = await getResourceType();
+  if (resourceType === ResourceType.GCP_GKE_CONTAINER) {
+    resources.push(await getKubernetesEngineResource());
+  } else if (resourceType === ResourceType.GCP_GCE_INSTANCE) {
+    resources.push(await getComputerEngineResource());
+  }
+  if (resourceType === ResourceType.AWS_EC2_INSTANCE) {
```
draffensperger: For consistency, should this have an `else` before the `if` similar to the `resourceType === ResourceType.GCP_GCE_INSTANCE` above?

packages/opencensus-resource-util/src/index.ts
```diff
@@ -13,3 +13,31 @@
...
+    resources.push(await getAwsEC2Resource());
+  }
```
draffensperger: What would you think about adding an exhuaustiveness check here? See https://basarat.gitbooks.io/typescript/docs/types/discriminated-unions.html

Would it make sense for this to be a `switch`?

packages/opencensus-resource-util/src/resource-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+export enum ResourceType {
+  GCP_GKE_CONTAINER = 'k8s_container',
+  GCP_GCE_INSTANCE = 'gce_instance',
+  AWS_EC2_INSTANCE = 'aws_ec2_instance',
+  NONE = 'NONE'
+}
+
+/** Determine the compute environment in which the code is running. */
+export async function getResourceType(): Promise<ResourceType> {
+  if (!resourceType) {
```
draffensperger: Optional nit: what would you think about making this first line to be `if (resourceType) return resourceType;`, which would allow if statements below to be less nested.

packages/opencensus-resource-util/src/resource-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/** Gets project id from GCP project metadata. */
+async function getProjectId() {
+  try {
+    return await gcpMetadata.project('project-id');
+  } catch {
+    /* ignore */
+  }
+  return '';
```
draffensperger: What would you think about moving this `return ''` up into the `catch` block? Then you could potentially remove the `/* ignore */` comment since the error handling just causes an empty string to return.

packages/opencensus-resource-util/src/resource-utils.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/** Gets instance id from GCP instance metadata. */
+async function getInstanceId() {
+  try {
+    return await gcpMetadata.instance('id');
+  } catch {
+    /* ignore */
+  }
+  return '';
```
draffensperger: Similar comment here on moving the `return ''` up into the `catch`

packages/opencensus-resource-util/src/index.ts
```diff
@@ -13,3 +13,31 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
```
draffensperger: Optional (may just be my opinion): I tend to like to have `index.ts` just export symbols and then have more clearly named files that do the heavy lifting. E.g. calling this one `detect-resource.ts` or similar and same for its test.

mayurkale22: Make Sense, Done.

packages/opencensus-resource-util/src/index.ts
```diff
@@ -13,3 +13,31 @@
...
+import {getAwsEC2Resource, getComputerEngineResource, getKubernetesEngineResource, getResourceType, ResourceType} from './resource-utils';
+
+/**
+ * Returns a Resource. Detector sequentially runs resource detection from
+ * environment variables, K8S, GCE and AWS.
+ */
+export async function detectResource(): Promise<Resource> {
+  const resources: Resource[] = [];
+  const resourceFromEnv = CoreResource.createFromEnvironmentVariables();
+  resources.push(resourceFromEnv);
```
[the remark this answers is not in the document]
mayurkale22: sure.

packages/opencensus-resource-util/src/index.ts
mayurkale22: Initial ```if...else``` block is to pick GKE or GCE, since one instance cannot be both a GCE instance and a GKE instance

packages/opencensus-resource-util/src/resource-utils.ts
mayurkale22: Good call. I was thinking about it, but missed somehow. done.

PR #314, opened 2019-01-30
Stats/Stackdriver: Use resource util to generate MonitoredResource

mayurkale22:
Fixes https://github.com/census-instrumentation/opencensus-node/issues/130

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
+++ b/packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
@@ -14,7 +14,8 @@
  * limitations under the License.
  */
 
-import {BucketOptions, DistributionBucket, DistributionValue, LabelKey, LabelValue, Metric, MetricDescriptor as OCMetricDescriptor, MetricDescriptorType, TimeSeriesPoint, Timestamp} from '@opencensus/core';
+import {BucketOptions, DistributionBucket, DistributionValue, LabelKey, Labels, LabelValue, Metric, MetricDescriptor as OCMetricDescriptor, MetricDescriptorType, TimeSeriesPoint, Timestamp} from '@opencensus/core';
+import * as resource from '@opencensus/resource-util';
 import * as os from 'os';
 import * as path from 'path';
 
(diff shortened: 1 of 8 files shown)


packages/opencensus-resource-util/src/resource-utils.ts
```diff
@@ -158,7 +158,8 @@ async function getProjectId() {
 /** Gets instance id from GCP instance metadata. */
 async function getInstanceId() {
   try {
-    return await gcpMetadata.instance('id');
+    const id = await gcpMetadata.instance('id');
```
mayurkale22: This is numerical ID, need to convert to string before assigning to label value.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -23,6 +24,55 @@ import {Distribution, LabelDescriptor, MetricDescriptor, MetricKind, MonitoredRe
...
+const GLOBAL = 'global';
+
+/* Return a self-configured StackDriver monitored resource. */
+export async function getDefaultResource(projectId: string):
+    Promise<MonitoredResource> {
+  let type;
+  const labels: Labels = {project_id: projectId};
+  let mappings: Labels = {};
+  const autoDetectedResource = await resource.detectResource();
+  switch (autoDetectedResource.type) {
```
draffensperger: Optional: if you make a helper function that implements the logic of this `switch` then you can make an assignment like `const [type, mappings] = getTypeAndMappings(autoDetectedResource.type)`. This would avoid the `let` initializations.

packages/opencensus-exporter-stackdriver/src/stackdriver-stats-utils.ts
```diff
@@ -237,6 +287,37 @@ function leftZeroPad(ns: number) {
   return `${pad}${str}`;
 }
 
+function getGcpResourceLabelsMappings() {
```
draffensperger: What's the motivation for having these be functions and not just assign the values to the constants above?

IMO, having the constant right next to its value would make it clearer and more compact, but up to you on this.

packages/opencensus-exporter-stackdriver/test/test-stackdriver-stats-utils.ts
```diff
@@ -413,4 +415,75 @@ describe('Stackdriver Stats Exporter Utils', () => {
...
+      const {type, labels} = monitoredResource;
+
+      assert.equal(type, 'global');
+      assert.deepEqual(labels, {project_id: 'my-project-id'});
+    });
+
+    it('should return a k8s MonitoredResource', async () => {
+      process.env.OC_RESOURCE_TYPE = 'k8s.io/container';
+      process.env.OC_RESOURCE_LABELS =
+          'k8s.io/pod/name=pod-xyz-123,k8s.io/container/name=c1,k8s.io/namespace/name=default,cloud.google.com/gce/zone=zone1';
```
draffensperger: Optional nit: for long strings like this, I often like to break them up with the `+` operator to keep the line under 80 chars.

PR #318, opened 2019-01-30
add timeout for AWS MetadataAccessor

mayurkale22:


packages/opencensus-resource-util/src/resource-utils.ts
```diff
@@ -142,7 +147,10 @@ async function awsMetadataAccessor<T>(): Promise<T> {
         }
       });
     });
-    req.on('error', (err) => reject(err));
+    req.on('error', (err) => {
+      clearTimeout(timeoutId);
```
draffensperger: I'm assuming you don't need to clear the timeout on success as well since the promise would already be resolved, but just wanted to state that out loud in case it's wrong.

mayurkale22: I did clear out the timeout when we get response from endpoint. Line: https://github.com/census-instrumentation/opencensus-node/pull/318/files/1b40f56cd06cdbda2ed712b32e3eb1e7bffd0984#diff-e008cd5b71782eb32c1f53b01df244a4R131

draffensperger: Thanks! Must have somehow missed that.

PR #323, opened 2019-02-01
Add message event size fields

draffensperger:
This will be useful for the `opencensus-web` project since message sizes are often known in the browser performance timing API and I'd like to make it include message events per the [OpenCensus HTTP specs](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md).


draffensperger:
@mayurkale22 could you merge this if all looks good?

Issue #117, opened 2018-08-31
Can't propogate trace context with http instrumentation?


yantang-msft:
I'm playing with this library and see how the correlation context is propagated through http request.
Here is the snippet
``` node
const tracing = require('@opencensus/nodejs');
const ConsoleExporter = require('@opencensus/core');
const propagation = require('@opencensus/propagation-tracecontext');

const traceContext = new propagation.TraceContextFormat();

const tracer = tracing.start({propagation: traceContext}).tracer;
const exporter = new ConsoleExporter.ConsoleExporter({
    bufferSize: 1,
    bufferTimeout: 1,
    logger: console
});
tracing.registerExporter(exporter);

const http = require('http');

http.get('http://www.google.com', (resp) => {
    let data = '';
    resp.on('data', (chunk) => {
        data += chunk;
    });

    resp.on('end', () => {
        console.log(`Response end, data length is ${data.length}`);
    });
}).on("error", (err) => {
    console.log("Error: " + err.message);
});

const stdin = process.openStdin();
stdin.addListener("data", () => {});
```

If I don't add the propagation context, the http instrumentation would successfully create a span for the outgoing request. But if I add the propagation context, it gives me this error. Did I missed anything?
```
Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ClientRequest.setHeader (_http_outgoing.js:469:11)
    at Object.setHeader (D:\OpenCensus\node\node_modules\@opencensus\instrumentation-http\build\src\http.js:182:29)
    at TraceContextFormat.inject (D:\OpenCensus\node\node_modules\@opencensus\propagation-tracecontext\build\src\tracecontext-format.js:109:20)
    at D:\OpenCensus\node\node_modules\@opencensus\instrumentation-http\build\src\http.js:187:29
    at contextManager.runAndReturn (D:\OpenCensus\node\node_modules\@opencensus\core\build\src\trace\model\tracer.js:111:20)
    at AsyncHooksNamespace.runAndReturn (D:\OpenCensus\node\node_modules\@opencensus\core\build\src\internal\cls-ah.js:51:21)
    at CoreTracer.startRootSpan (D:\OpenCensus\node\node_modules\@opencensus\core\build\src\trace\model\tracer.js:84:36)
    at Object.outgoingRequest [as get] (D:\OpenCensus\node\node_modules\@opencensus\instrumentation-http\build\src\http.js:159:42)
[...]

PR #324, opened 2019-02-03
fixes http.get patches for header propagation #117

vmarchaud:
I got the same issue as #117 while integrating the tracing system into my APM.
I did go over https://github.com/googleapis/cloud-trace-nodejs/ to checkout their implementation and found a details that could explain the problem: https://github.com/googleapis/cloud-trace-nodejs/blob/master/src/plugins/plugin-http.ts#L198

The test are passing for one good reason: they are all `nocked`, so the actual http code never run. I added a test that make a GET request to `google.fr`, it might be better to spawn a local server to avoid instability due to network.

### DIFF
--- a/packages/opencensus-instrumentation-http/src/http.ts
+++ b/packages/opencensus-instrumentation-http/src/http.ts
@@ -24,7 +24,7 @@ import * as uuid from 'uuid';
 import {HttpPluginConfig, IgnoreMatcher} from './types';
 
 
-export type HttpGetCallback = (res: httpModule.IncomingMessage) => void
+export type HttpGetCallback = (res: httpModule.IncomingMessage) => void;
 export type HttpModule = typeof httpModule;
 export type RequestFunction = typeof httpModule.request;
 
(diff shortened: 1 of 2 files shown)


vmarchaud:
Not related but i believe the problem is the same with `https.get`, might need another PR

vmarchaud:
@draffensperger @mayurkale22 Sorry for the ping but do the PR need something specific to be reviewed and merged ?
I wanted to contribute to the node implementation since i'm using it for my APM but the time it take to make a change merged is quite long.

draffensperger:
Hi @[vmarchaud](/vmarchaud), thanks for the contribution and sorry for the slow response! I suspect this is just a procedural issue where both @mayurkale22 missed this PR since it had others tagged on it. Mayur, is it possible to change the default reviewers for new PRs?

This looks good to me, though Mayur will need to do the final review and merge.

mayurkale22:
Hey @vmarchaud, sorry for the late response. PR looks good for me. @draffensperger I think it is possible using ```CODEOWNERS``` file. https://help.github.com/articles/about-code-owners/

PR #328, opened 2019-02-05
Tracing: Add code and message to Span -> Status

### LINKED ISSUE #327: Tracing: Add message to Status to Span.
Current ```Span -> status``` is represented as only a number. 

Expected:
```
export interface Status {
  // The status code. This is optional field. It is safe to assume 0 (OK) when not set.
 code: number = 0;
  // A developer-facing error message, which should be in English.
 message: string;
}
```

See https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto#L316-L323

Counterpart in Python: https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/status.py

mayurkale22:
Fixes #327

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - Add ignoreIncomingPaths and ignoreOutgoingUrls support to the http and https tracing instrumentations.
 - Add ```opencensus-resource-util``` to auto detect AWS, GCE and Kubernetes(K8S) monitored resource, based on the environment where the application is running.
 - Add optional `uncompressedSize` and `compressedSize` fields to `MessageEvent` interface.
+- Add a ```setStatus``` method in the Span.
 
  **This release has multiple breaking changes. Please test your code accordingly after upgrading.**
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to this project will be documented in this file.
 - Modify `Logger` interface: `level` made optional, `silly` removed.
 - The ```new Stats()``` has been deprecated on Stats class. The global singleton ```globalStats``` object should be used instead. Also, ```registerView()``` is separated out from ```createView()```.
 - Use ```TagKey```, ```TagValue``` and ```TagMap``` to create the tag keys, tag values.
+- The `status` field on `Span` is no longer a number, use `CanonicalCode` instead.
 
 ##### Old code
 ```js
--- a/packages/opencensus-core/src/trace/model/span-base.ts
+++ b/packages/opencensus-core/src/trace/model/span-base.ts
@@ -52,7 +52,7 @@ export abstract class SpanBase implements types.Span {
   /** Kind of span. */
   kind: string = null;
   /** A final status for this span */
-  status: number;
+  status: types.Status = {code: 0};
   /** set isRootSpan  */
   abstract get isRootSpan(): boolean;
 
@@ -181,6 +181,18 @@ export abstract class SpanBase implements types.Span {
       'id': id,
       'timestamp': timestamp ? timestamp : Date.now(),
     } as types.MessageEvent);
+  }
+
+  /**
+   * Sets a status to the span.
+   * @param code The status code.
+   * @param message optional A developer-facing error message.
+   */
+  setStatus(code: number, message?: string) {
+    this.status.code = code;
+    if (message) {
+      this.status.message = message;
+    }
   }
 
   /** Starts the span. */
--- a/packages/opencensus-core/src/trace/model/span-base.ts
+++ b/packages/opencensus-core/src/trace/model/span-base.ts
@@ -18,6 +18,9 @@ import {Clock} from '../../internal/clock';
 import {randomSpanId} from '../../internal/util';
 import * as types from './types';
 
+const STATUS_OK = {
+  code: types.CanonicalCode.OK
+};
 
 /** Defines a base model for spans. */
 export abstract class SpanBase implements types.Span {
@@ -52,7 +55,7 @@ export abstract class SpanBase implements types.Span {
   /** Kind of span. */
   kind: string = null;
   /** A final status for this span */
-  status: types.Status = {code: 0};
+  status: types.Status = STATUS_OK;
   /** set isRootSpan  */
   abstract get isRootSpan(): boolean;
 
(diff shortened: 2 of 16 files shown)


packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -183,6 +183,18 @@ export abstract class SpanBase implements types.Span {
...
   }
 
+  /**
+   * Sets a status to the span.
+   * @param code The status code.
+   * @param message optional A developer-facing error message.
+   */
+  setStatus(code: number, message?: string) {
+    this.status.code = code;
+    if (message) {
```
draffensperger: If someone sets a new code but no message, would this check leave the old message assigned to status?

Could this function just be `this.status = {code, message}`, since we the `message` field is optional?

packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -30,6 +30,17 @@ export interface Attributes {
   [attributeKey: string]: string|number|boolean;
 }
 
+/**
+ * The status of a Span by providing a standard code in conjunction
+ * with an optional descriptive message.
+ */
+export interface Status {
+  /** The status code. */
```
draffensperger: Could you add in the comment that this status code should correspond with a specific set of OpenCensus trace status codes and that HTTP status codes should be appropriately converted per the specs at  https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#mapping-from-http-status-codes-to-trace-status-codes?

packages/opencensus-core/test/test-span.ts
```diff
@@ -258,4 +258,35 @@ describe('Span', () => {
       assert.ok(instanceOfLink(span.messageEvents[0]));
     });
   });
+
+  /**
+   * Should set a status
+   */
+  describe('addStatus()', () => {
```
draffensperger: Nit: `'setStatus'`

packages/opencensus-core/test/test-span.ts
```diff
@@ -258,4 +258,35 @@ describe('Span', () => {
...
+      assert.equal(span.status.code, 0);
+      assert.equal(span.status.message, null);
+    });
+
+    it('should set an error status', () => {
+      const rootSpan = new RootSpan(tracer);
+      rootSpan.start();
+      const span = new Span(rootSpan);
+      span.start();
+      span.setStatus(400, 'This is an error');
```
draffensperger: Per comment about about the canonical values, could this be a number like  `span.SetSTatus('3', 'Invalid argument')` ?

packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -30,6 +30,17 @@ export interface Attributes {
...
+  code: number;
+  /** A developer-facing error message. */
+  message?: string;
+}
+
```
draffensperger: What would you think about adding an enum with the list of OpenCensus status codes from the table referenced above?

draffensperger: To be clear I'm not suggesting the type of the `code` field be the enum (that might limit clients of the library if a new status is added but they can't upgrade yet, and for `opencensus-web` I want to keep `@opencensus/core` just as a dev dependency). But just having the enum provides documentation for what the possible canonical status codes are.

mayurkale22: I checked in Java implementation, we are using enum for ```code``` -> https://github.com/census-instrumentation/opencensus-java/blob/master/api/src/main/java/io/opencensus/trace/Status.java#L384. This is also matching with [status spec](https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/HTTP.md#status).

draffensperger: OK, I'd be fine with using an enum for it.

packages/opencensus-core/src/trace/model/span-base.ts
mayurkale22: Good catch! Done.

packages/opencensus-core/test/test-span.ts
mayurkale22: Done

packages/opencensus-core/src/trace/model/types.ts
mayurkale22: Done as per https://github.com/census-instrumentation/opencensus-node/pull/328#discussion_r253985285.

packages/opencensus-core/test/test-span.ts
```diff
@@ -258,4 +258,35 @@ describe('Span', () => {
       assert.ok(instanceOfLink(span.messageEvents[0]));
     });
   });
+
+  /**
```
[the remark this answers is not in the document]
mayurkale22: Done.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ All notable changes to this project will be documented in this file.
 - Add ignoreIncomingPaths and ignoreOutgoingUrls support to the http and https tracing instrumentations.
 - Add ```opencensus-resource-util``` to auto detect AWS, GCE and Kubernetes(K8S) monitored resource, based on the environment where the application is running.
 - Add optional `uncompressedSize` and `compressedSize` fields to `MessageEvent` interface.
+- Add a ```setStatus``` method in the Span.
 
  **This release has multiple breaking changes. Please test your code accordingly after upgrading.**
```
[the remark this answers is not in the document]
mayurkale22: Done

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -219,15 +219,15 @@ export class GrpcPlugin extends BasePlugin {
         // tslint:disable-next-line:no-any
         value: any, trailer: grpcTypes.Metadata, flags: grpcTypes.writeFlags) {
       if (err) {
-        rootSpan.status = GrpcPlugin.convertGrpcStatusToSpanStatus(err.code);
```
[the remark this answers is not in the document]
mayurkale22: Already did it here: https://github.com/census-instrumentation/opencensus-node/pull/328/files#diff-a5285736aa52f4a852028199a7016379R404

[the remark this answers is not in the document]
draffensperger: Thanks! I think I got lost in the changes there and did not look closely enough.

mayurkale22:
Thanks @draffensperger for the reviews.

Issue #316, opened 2019-01-30
Enforce TraceParams Limits


mayurkale22:
After finishing #311, Please use that functionality (global trace parameters) to enforce limits on attributes, annotations, links and message events (better to handle each in separate PR).

Also, update ```droppedAnnotationsCount```, ```droppedMessageEventsCount```, ```droppedAttributesCount``` and ```droppedLinksCount``` accordingly. 

Counterpart in Java: https://github.com/census-instrumentation/opencensus-java/blob/master/impl_core/src/main/java/io/opencensus/implcore/trace/RecordEventsSpanImpl.java#L520

Counterpart in Go: https://github.com/census-instrumentation/opencensus-go/pull/1020

vigneshtdev:
I will try to finish off #270 first.

mayurkale22:
@vigneshtdev just wanted to keep you in this loop. If you don't mind, I will implement this (later this week). Let me know. Thanks again!

PR #330, opened 2019-02-06
Enforce attribute limit

mayurkale22:
This PR will address the first part of #316

### DIFF
--- a/packages/opencensus-core/src/trace/model/span-base.ts
+++ b/packages/opencensus-core/src/trace/model/span-base.ts
@@ -40,7 +40,7 @@ export abstract class SpanBase implements types.Span {
   /** An object to log information to */
   logger: Logger;
   /** A set of attributes, each in the format [KEY]:[VALUE] */
-  attributes: types.Attributes = {attributeMap: {}, droppedAttributesCount: 0};
+  attributes: types.Attributes = {};
   /** A text annotation with a set of attributes. */
   annotations: types.Annotation[] = [];
   /** An event describing a message sent/received between Spans */
@@ -61,6 +61,9 @@ export abstract class SpanBase implements types.Span {
   abstract get isRootSpan(): boolean;
   /** Trace Parameters */
   activeTraceParams: configTypes.TraceParams;
+
+  /** The number of dropped attributes. */
+  droppedAttributesCount = 0;
 
   /** Constructs a new SpanBaseModel instance. */
   constructor() {
@@ -139,14 +142,17 @@ export abstract class SpanBase implements types.Span {
    * @param value The result of an operation.
    */
   addAttribute(key: string, value: string|number|boolean) {
-    if (Object.keys(this.attributes.attributeMap).length >=
-        this.activeTraceParams.numberOfAttributesPerSpan) {
-      this.attributes.droppedAttributesCount++;
-      const attributeKeyToDelete =
-          Object.keys(this.attributes.attributeMap).shift();
-      delete this.attributes.attributeMap[attributeKeyToDelete];
+    if (this.attributes[key]) {
+      delete this.attributes[key];
     }
-    this.attributes.attributeMap[key] = value;
+
+    if (Object.keys(this.attributes).length >=
+        this.activeTraceParams.numberOfAttributesPerSpan) {
+      this.droppedAttributesCount++;
+      const attributeKeyToDelete = Object.keys(this.attributes).shift();
+      delete this.attributes[attributeKeyToDelete];
+    }
+    this.attributes[key] = value;
   }
 
   /**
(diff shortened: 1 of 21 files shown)


packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -133,7 +137,17 @@ export abstract class SpanBase implements types.Span {
    * @param value The result of an operation.
    */
   addAttribute(key: string, value: string|number|boolean) {
-    this.attributes[key] = value;
+    this.totalRecordedAttributes++;
+    if (Object.keys(this.attributes.attributeMap).length >=
+        this.activeTraceParams.numberOfAttributesPerSpan) {
+      const attributeKeyToDelete =
+          Object.keys(this.attributes.attributeMap).shift();
+      delete this.attributes.attributeMap[attributeKeyToDelete];
```
draffensperger: Why delete an existing attribute rather than just dropping the new attribute? Is that the specified way to do it?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -133,7 +137,17 @@ export abstract class SpanBase implements types.Span {
...
+    }
+
+    this.attributes.attributeMap[key] = value;
+    this.attributes.droppedAttributesCount = this.totalRecordedAttributes -
```
draffensperger: Would it work to just increment `this.attributes.droppedAttributesCount` in the above `if` block?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -55,6 +55,10 @@ export abstract class SpanBase implements types.Span {
   status: number;
   /** set isRootSpan  */
   abstract get isRootSpan(): boolean;
+  /** Trace Parameters */
+  activeTraceParams: configTypes.TraceParams;
+
+  private totalRecordedAttributes = 0;
```
draffensperger: If we increment `droppedAttributeCount` do we need this field?

packages/opencensus-core/src/trace/model/span-base.ts
mayurkale22: Make Sense, Done. 👍

packages/opencensus-core/src/trace/model/span-base.ts
mayurkale22: This implementation is based on the spec here: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/TraceConfig.md#limits

> When limits are exceeded, implementations should by default preserve the most recently added values and drop the oldest values.

draffensperger: Thanks, makes sense.

rghetia: does shift() guarantee that it returns the oldest?

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -331,13 +334,16 @@ describe('OpenCensus Agent Exporter', () => {
       rootSpan.setStatus(CanonicalCode.OK);
 
       // Attribute
+      rootSpan.addAttribute('my_first_attribute', 'foo');
+      rootSpan.addAttribute('my_second_attribute', 'foo2');
       rootSpan.addAttribute('my_attribute_string', 'bar2');
       rootSpan.addAttribute('my_attribute_number', 456);
       rootSpan.addAttribute('my_attribute_boolean', false);
```
rghetia: Change the test to add same key again such that it is no longer oldest and hence should not be deleted.

packages/opencensus-core/src/trace/model/span-base.ts
mayurkale22: As per my knowledge, integer properties are sorted, others (non-integer) are listed in the creation order. ```shift``` -> Removes the first element from an array. @draffensperger correct me, if I am wrong.

draffensperger: I checked on the order by running the following JS in a `node` console:
```
x = {}
x.b = 2
x.a = 1
Object.keys(x)  // gives ['b', 'a'] as expected

y = {}
y.a = 1
y.b = 2
Object.keys(y)  // gives ['a', 'b'] as expected
```

That said, what would you think about adding a unit test to validate this "removes the last attribute" behavior?

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
mayurkale22: Done, PTAL

packages/opencensus-core/src/trace/model/span-base.ts
draffensperger: OK, sweet!

PR #331, opened 2019-02-06
Enforce Links limit

mayurkale22:
This PR will address the second part (enforce limits on links) of #316

### DIFF
--- a/packages/opencensus-exporter-ocagent/test/test-ocagent.ts
+++ b/packages/opencensus-exporter-ocagent/test/test-ocagent.ts
@@ -175,7 +175,7 @@ describe('OpenCensus Agent Exporter', () => {
     tracing = nodeTracing.start({
       exporter: ocAgentExporter,
       samplingRate: INITIAL_SAMPLER_PROBABILITY,
-      traceParams: {numberOfAttributesPerSpan: 4}
+      traceParams: {numberOfAttributesPerSpan: 4, numberOfLinksPerSpan: 3}
     });
   });
 
@@ -354,6 +354,8 @@ describe('OpenCensus Agent Exporter', () => {
       rootSpan.addMessageEvent(null as any, 'ffff', timeStamp);
 
       // Links
+      rootSpan.addLink('aaaaa', 'aaa', 'CHILD_LINKED_SPAN');
+      rootSpan.addLink('bbbbb', 'bbbbb', 'CHILD_LINKED_SPAN');
       rootSpan.addLink('ffff', 'ffff', 'CHILD_LINKED_SPAN', {
         'child_link_attribute_string': 'foo1',
         'child_link_attribute_number': 123,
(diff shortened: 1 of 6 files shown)


packages/opencensus-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -517,4 +519,175 @@ describe('OpenCensus Agent Exporter', () => {
...
+                           value: 'stringValue',
+                           stringValue: {value: 'foo1', truncatedByteCount: 0}
+                         },
+                         child_link_attribute_number:
+                             {value: 'intValue', intValue: '123'},
+                         child_link_attribute_boolean:
+                             {value: 'boolValue', boolValue: true}
+                       }
+                     }
+                   },
```
rghetia: you have checked the drop counts for links earlier but it would nice to check that oldest links is removed. you can either add this is as part of this or where you test the dropLinkCount.

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -517,4 +519,175 @@ describe('OpenCensus Agent Exporter', () => {
...
+         // Attribute
+         rootSpan.addAttribute('my_first_attribute', 'foo');
+         rootSpan.addAttribute('my_second_attribute', 'foo2');
+
+         // Annotation
+         rootSpan.addAnnotation(
+             'my_annotation',
+             {myString: 'bar', myNumber: 123, myBoolean: true});
+
+         // Metric Event
```
draffensperger: Nit: "Message event"

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
```diff
@@ -517,4 +519,175 @@ describe('OpenCensus Agent Exporter', () => {
...
+         const timeStamp = 123456789;
+         rootSpan.addMessageEvent('MessageEventTypeSent', 'ffff', timeStamp);
```
draffensperger: Curious, why is this 'MessageEventTypeSent' and not just 'SENT'?

This would be for a follow up PR, but what would you think about adding enums for span kind, message event type and link type?

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
mayurkale22: Done https://github.com/census-instrumentation/opencensus-node/pull/331/commits/34ea4cf501e1838a1201b856e93ae277c31e182a

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
mayurkale22: This is legacy code, I am not sure rational behind this. Created an issue to track this -> https://github.com/census-instrumentation/opencensus-node/issues/333. Thanks 👍

packages/opencensus-exporter-ocagent/test/test-ocagent.ts
mayurkale22: Good Point! We already have asset check here https://github.com/census-instrumentation/opencensus-node/pull/331/files#diff-0ab7878c00c0a6393dd66fd59e796156R482 for the links. Basically, it doesn't contain a dropped link object.

rghetia: as per our conversation, this is already taken care of in previous test 'should adapt a span correctly' on line 320.
so ignore the comment.

Issue #105, opened 2018-08-24
Cannot create nested Child Spans


thesandlord:
Currently, there is no way to create nested spans:

```
Root Span
      Child Span
             Grandchild Span
```

In the [type definition](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/model/types.ts#L276), startChildSpan has the following signature:

```TypeScript
startChildSpan(name?: string, type?: string, parentSpanId?: string): Span;
```

However, if you look at the [actual function implementation](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/model/tracer.ts#L225):

```TypeScript
/**
   * Starts a span.
   * @param name The span name.
   * @param type The span type.
   * @param parentSpanId The parent span ID.
   */
  startChildSpan(name?: string, type?: string): types.Span {
    let newSpan: types.Span = null;
    if (!this.currentRootSpan) {
      this.logger.debug(
          'no current trace found - must start a new root span first');
    } else {
      newSpan = this.currentRootSpan.startChildSpan(name, type);
    }
    return newSpan;
  }
```

Even though the `@param parentSpanId` is present in the JavaDoc tags, it isn't actually used in the actual function!

This means that the parentSpanId is *always* set to the `rootSpanId`, which means nesting spans is impossible.

The simple workaround is to just override the `parentSpanId` attribute of the grandchild span to the child span id. I've tested this and it seems to work fine.

i.e:

```JavaScript
const childSpan = tracing.tracer.startChildSpan('Child Span')

const grandChildSpan = tracing.tracer.startChildSpan('Child Span')
grandChildSpan.parentSpanId = childSpan.id
```

I think the the code above could be fixed with the following:
```TypeScript
/**
   * Starts a span.
   * @param name The span name.
   * @param type The span type.
   * @param parentSpanId The parent span ID.
   */
  startChildSpan(name?: string, type?: string, parentSpanId?: string): types.Span {
    let newSpan: types.Span = null;
    if (!this.currentRootSpan) {
      this.logger.debug(
          'no current trace found - must start a new root span first');
    } else {
      newSpan = this.currentRootSpan.startChildSpan(name, type);
    }
    if (parentSpanId) {
      newSpan.parentSpanId = parentSpanId;
    }
    return newSpan;
  }
```

On another note, there really should be the option of passing in an object to the `createChildSpan` function instead of three arguments.

hvent90:
[...]

PR #332, opened 2019-02-07
 Add support for child span count

### LINKED ISSUE #306: Tracing: Add child span count to Span.
See https://github.com/census-instrumentation/opencensus-proto/blob/6d79e46232e026136ae013489ef611ca721cb839/src/opencensus/proto/trace/v1/trace.proto#L302.

mayurkale22:
Fixes #306  
This is based on spec: See https://github.com/census-instrumentation/opencensus-proto/blob/6d79e46232e026136ae013489ef611ca721cb839/src/opencensus/proto/trace/v1/trace.proto#L302.

### DIFF
--- a/packages/opencensus-core/test/test-root-span.ts
+++ b/packages/opencensus-core/test/test-root-span.ts
@@ -68,6 +68,21 @@ describe('RootSpan', () => {
     });
   });
 
+  describe('get numberOfChildren()', () => {
+    it('should get numberOfChildren from rootspan instance', () => {
+      const root = new RootSpan(tracer);
+      root.start();
+      assert.equal(root.numberOfChildren, 0);
+      root.startChildSpan('spanName', types.SpanKind.UNSPECIFIED);
+      assert.equal(root.numberOfChildren, 1);
+
+      for (let i = 0; i < 10; i++) {
+        root.startChildSpan('spanName' + i, types.SpanKind.UNSPECIFIED);
+      }
+      assert.equal(root.numberOfChildren, 11);
+    });
+  });
+
   /**
    * Should get trace id from rootspan instance
    */
@@ -170,6 +185,7 @@ describe('RootSpan', () => {
       for (const span of root.spans) {
         assert.ok(span.ended);
       }
+      assert.equal(root.numberOfChildren, 1);
     });
   });
 
@@ -208,6 +224,7 @@ describe('RootSpan', () => {
 
       assert.ok(rootSpan.annotations.length > 0);
       assert.equal(rootSpan.droppedAnnotationsCount, 0);
+      assert.equal(rootSpan.numberOfChildren, 0);
       assert.ok(instanceOfAnnotation(rootSpan.annotations[0]));
     });
   });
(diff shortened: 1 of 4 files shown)


packages/opencensus-core/test/test-root-span.ts
```diff
@@ -169,6 +184,7 @@ describe('RootSpan', () => {
       for (const span of root.spans) {
         assert.ok(span.ended);
       }
+      assert.equal(root.numberOfChildren, 1);
```
rghetia: there is no need to check for child count in other test case unless the test case affects the logic.

packages/opencensus-core/test/test-root-span.ts
```diff
@@ -206,6 +222,7 @@ describe('RootSpan', () => {
       rootSpan.addAnnotation('description test', {} as Attributes, Date.now());
 
       assert.ok(rootSpan.annotations.length > 0);
+      assert.equal(rootSpan.numberOfChildren, 0);
```
rghetia: ditto.

packages/opencensus-core/test/test-root-span.ts
mayurkale22: Sure, Done.

mayurkale22:
> What would you think about making the child count a core property of the `Span` type? That would then enable situations where there is a larger operation root span with sub-operations as child spans and those child spans then make RPC calls.

AFAIK there isn't a notion of grandchild spans in node (all child spans are direct children of a root span). https://github.com/census-instrumentation/opencensus-node/issues/105

draffensperger:
> AFAIK there isn't a notion of grandchild spans in node (all child spans are direct children of a root span). #105

Is that by design based on the specs / other language implementations? Or is that more of a specific limitation of OpenCensus Node at this time?

mayurkale22:
@draffensperger allow me to rephrase my previous sentence https://github.com/census-instrumentation/opencensus-node/pull/332#issuecomment-462457456.

AFAIK in terms of single process or server, there will be only two types of requests (incoming and outgoing). Root spans corresponds to incoming requests and child span for outgoing. There is only in and out (no grand out). So, if there are multiple micro services running we connect all these spans through common traceId (which is propagated through context). 

> What would you think about making the child count a core property of the `Span` type?

I think it makes sense to have child count property of the ```RootSpan``` type, as there is no way to create child from child (nested child).

draffensperger:
I agree that for RPCs you typically only have in and out, but if we allow for other types of spans, I think we can get the grandchild case. I'm thinking of custom intermediate spans that represent larger conceptual things like "process middleware" or "load recommendations" and similar, that each may have multiple RPCs or other custom spans in them. As a user views a trace, they could then see which high level operation in the request is slow and then drill down to the subcalls of that operation. This of course requires that the user sets up those custom spans, but ideally I would think we would let them do that.

draffensperger:
I think the work I'm suggesting is captured well in a previous issue that I just commented on: https://github.com/census-instrumentation/opencensus-node/issues/105#issuecomment-468304948

So I'm fine with merging this and then adding support for nested child spans at a later time.

mayurkale22:
> I think the work I'm suggesting is captured well in a previous issue that I just commented on: [#105 (comment)](https://github.com/census-instrumentation/opencensus-node/issues/105#issuecomment-468304948)
> 
> So I'm fine with merging this and then adding support for nested child spans at a later time.

Ok cool, thanks.

PR #334, opened 2019-02-07
Use enum for MessageEvent Type, Link Type and SpanKind

### LINKED ISSUE #333: Tracing: Add enum for MessageEvent Type, Link Type and Span Kind
Consider replacing ```string``` type with strict enum type for ```MessageEvent```, ```Link``` and ```SpanKind``` (similar to other libraries). Specs: https://github.com/census-instrumentation/opencensus-proto/blob/6d79e46232e026136ae013489ef611ca721cb839/src/opencensus/proto/trace/v1/trace.proto#L186-L193

Example:

SpanKind
```
export enum SpanKind {
  UNSPECIFIED = 0,
  SERVER = 1,
  CLIENT = 2,
}
```

LinkType
```
export enum LinkType {
  UNSPECIFIED = 0,
  CHILD_LINKED_SPAN = 1,
  PARENT_LINKED_SPAN = 2,
}
```
MessageEventType
```
export enum MessageEventType {
  UNSPECIFIED = 0,
  SENT = 1,
  RECEIVED = 2,
}
```

mayurkale22:
Fixes #333

### DIFF
--- a/packages/opencensus-core/src/trace/model/types.ts
+++ b/packages/opencensus-core/src/trace/model/types.ts
@@ -191,7 +191,7 @@ export interface MessageEvent {
   /** A timestamp for the event. */
   timestamp: number;
   /** Indicates whether the message was sent or received. */
-  type: string;
+  type: MessageEventType;
   /** An identifier for the MessageEvent's message. */
   id: string;
   /** The number of uncompressed bytes sent or received. */
@@ -213,7 +213,7 @@ export interface Link {
   /** The span ID for a span within a trace. */
   spanId: string;
   /** The relationship of the current span relative to the linked. */
-  type: string;
+  type: LinkType;
   /** A set of attributes on the link. */
   attributes: Attributes;
 }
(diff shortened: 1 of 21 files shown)


packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -249,6 +249,51 @@ export interface SpanEventListener {
   onEndSpan(span: RootSpan): void;
 }
 
+/** An event describing a message sent/received between Spans. */
+export enum MessageEventType {
+  /** Unknown event type. */
+  UNSPECIFIED = 0,
```
draffensperger: I do like the numeric enums here so on balance I recommend keeping them this way (nice correspondence with the protos too). But on the other hand, use of a string enum could enable the breaking change to be less significant since the strings 'SENT' and 'RECEIVED' would still be valid. Just mentioning it because it's in my mind, but recommend keeping it this way!

mayurkale22: I compared with other lang. implementation, we use numeric enums for these (```Kind```, ```Link``` and ```MessageEvent```) types. I am trying to get most of the breaking change in this and next release before going to Beta.

draffensperger: Makes sense. The sooner the breaking changes come the less the pain in the future.

PR #338, opened 2019-02-11
Exporter/Stackdriver: Use Trace SDKs (v2 API)

### LINKED ISSUE #322: Use Trace SDKs (v2 API)
Currently, OC-Node uses Trace SDKs (v1 API) -> which doesn't support ```Annotations```, ```Links```, ```Message Events```, ```Time Events``` etc. 

The plan is to migrate to Trace SDKs (v2 API) asap. All of the other OC languages are on V2 API.

https://github.com/googleapis/google-api-nodejs-client/tree/master/src/apis/cloudtrace

mayurkale22:
Fixes #322 and #39

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
+++ b/packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
@@ -1,0 +1,170 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as coreTypes from '@opencensus/core';
+import * as types from './types';
+
+const AGENT_LABEL_KEY = 'g.co/agent';
+const AGENT_LABEL_VALUE_STRING = `opencensus-node [${coreTypes.version}]`;
+const AGENT_LABEL_VALUE = createAttributeValue(AGENT_LABEL_VALUE_STRING);
+
+/**
+ * Creates StackDriver Links from OpenCensus Link.
+ * @param links coreTypes.Link[]
+ * @param droppedLinksCount number
+ * @returns types.Links
+ */
+export function createLinks(
+    links: coreTypes.Link[], droppedLinksCount: number): types.Links {
+  return {link: links.map((link) => createLink(link)), droppedLinksCount};
+}
+
+/**
+ * Creates StackDriver Attributes from OpenCensus Attributes.
+ * @param attributes coreTypes.Attributes
+ * @param resourceLabels Record<string, types.AttributeValue>
+ * @param droppedAttributesCount number
+ * @returns types.Attributes
+ */
+export function createAttributes(
+    attributes: coreTypes.Attributes,
+    resourceLabels: Record<string, types.AttributeValue>,
+    droppedAttributesCount: number): types.Attributes {
+  const attributesBuilder =
+      createAttributesBuilder(attributes, droppedAttributesCount);
+  attributesBuilder.attributeMap[AGENT_LABEL_KEY] = AGENT_LABEL_VALUE;
+  attributesBuilder.attributeMap =
+      Object.assign({}, attributesBuilder.attributeMap, resourceLabels);
+  return attributesBuilder;
+}
+
+/**
+ * Creates StackDriver TimeEvents from OpenCensus Annotation and MessageEvent.
+ * @param annotationTimedEvents coreTypes.Annotation[]
+ * @param messageEventTimedEvents coreTypes.MessageEvent[]
+ * @param droppedAnnotationsCount number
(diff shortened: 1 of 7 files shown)


packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+ * @returns types.TimeEvents
+ */
+export function createTimeEvents(
+    annotationTimedEvents: coreTypes.Annotation[],
+    messageEventTimedEvents: coreTypes.MessageEvent[],
+    droppedAnnotationsCount: number,
+    droppedMessageEventsCount: number): types.TimeEvents {
+  const timeEvents: types.TimeEvent[] = [];
+  if (annotationTimedEvents) {
+    annotationTimedEvents.forEach(annotation => {
```
draffensperger: Could this be a `map` rather than `forEach` + `push`?

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
```diff
@@ -0,0 +1,170 @@
...
+  const type = createLinkType(link.type);
+  const attributes = createAttributesBuilder(link.attributes, 0);
+  return {traceId, spanId, type, attributes};
+}
+
+function createAttributeValue(value: string|number|
+                              boolean): types.AttributeValue {
+  switch (typeof value) {
+    case 'number':
+      return {intValue: String(value)};
```
draffensperger: Could you make this be `doubleValue`? The `number` type in JS is a double so we would be losing accuracy for floating point values.

I would think if something is a BigInt and fits in 64 bits it should be made to be an `intValue`, which it looks like [Node 10 supports](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V10.md#2018-10-10-version-10120-current-targos)

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
```diff
@@ -70,51 +72,66 @@ export class StackdriverTraceExporter implements Exporter {
...
-   * @param traces
+   * Sends new spans to new or existing traces in the Stackdriver format to the
+   * service.
+   * @param spans
    */
-  private sendTrace(traces: TracesWithCredentials) {
+  private batchWriteSpans(spans: SpansWithCredentials) {
     return new Promise((resolve, reject) => {
-      cloudTrace.projects.patchTraces(traces, (err: Error) => {
+      cloudTrace.projects.traces.batchWrite(spans, (err: Error) => {
```
draffensperger: Any reason we are using HTTP and not gRPC? (If the answer is just "that's what we did before", could you create an issue to track moving to gRPC for this?)

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
```diff
@@ -70,51 +72,66 @@ export class StackdriverTraceExporter implements Exporter {
...
+      endTime: span.endTime.toISOString(),
+      attributes: createAttributes(
+          span.attributes, resourceLabels, span.droppedAttributesCount),
+      timeEvents: createTimeEvents(
+          span.annotations, span.messageEvents, span.droppedAnnotationsCount,
+          span.droppedMessageEventsCount),
+      links: createLinks(span.links, span.droppedLinksCount),
+      status: {code: span.status.code},
+      sameProcessAsParentSpan: !span.remoteParent,
+      childSpanCount: null,
```
songy23: nit: leave a TODO here?

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
mayurkale22: v2 API doesn't support ```doubleValue``` type for ```AttributeValue``` -> https://github.com/googleapis/google-api-nodejs-client/blob/master/src/apis/cloudtrace/v2.ts#L153-L166

draffensperger: Hmm, OK. Then here are a couple options:

* Just always translate `number` to its string representation. This is probably my suggestion. Maybe add a TODO to change this once the API supports it. The OpenCensus trace proto has `doubleValue`.
* Check whether the number fits in a 64-bit int (less than the max 64-bit int and no fractional part) and only then write it as an `intValue` and otherwise use `String`.

I think `intValue` uses `String` type because `number` can't represent full accuracy 64-bit ints.

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
mayurkale22: Good point, added TODO here.

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
mayurkale22: Done.

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace.ts
mayurkale22: done.

packages/opencensus-exporter-stackdriver/src/stackdriver-cloudtrace-utils.ts
mayurkale22: Added TODO here to consider ```doubleValue``` type once available with cloudTrace V2 API. As per our offline discussion (+ your suggestion), will always translate ```number``` to its string representation for ```AttributeValue``` type.

PR #344, opened 2019-02-13
Add message sizes to `addMessageEvent` function

draffensperger:
Message size parameters were added in https://github.com/census-instrumentation/opencensus-node/pull/323 and this now adds them to the `addMessageEvent` function on span as optional parameters.

This also adds a comment to the `id` field of the `MessageEvent` interface to explain that it is a hexadecimal string that should increment as new messages are sent.

cc/ @odeke-em

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Add `compressedSize` and `uncompressedSize` params to `Span.addMessageEvent`
 
 ## 0.0.9 - 2019-02-12
 - Add Metrics API.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
-- Add `compressedSize` and `uncompressedSize` params to `Span.addMessageEvent`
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## Unreleased
 
+- Add `compressedSize` and `uncompressedSize` params to `Span.addMessageEvent`
```
mayurkale22: Nit: Add optional ?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -209,18 +209,25 @@ export abstract class SpanBase implements types.Span {
    * @param type The type of message event.
    * @param id An identifier for the message event.
    * @param timestamp A time in milliseconds. Defaults to Date.now()
+   * @param compressedSize The number of uncompressed bytes sent or received
```
mayurkale22: typo: uncompressed bytes -> compressed bytes

draffensperger: Fixed, thanks!

CHANGELOG.md
draffensperger: Done.

draffensperger:
Yes, sure, I can add the exporter logic in a new PR. Probably need it for Stackdriver and agent.

PR #345, opened 2019-02-15
 exporter/Zipkin: Add missing fields (Annotation,Status,MessageEvent etc)

### LINKED ISSUE #183: exporter/Zipkin: missing fields
Reported offline by @vimalraj-a
>tags and annotations are not supported in nodejs zipkin-exporter. I can help by implementing this if not someone doing it. @Authors Let me know the contribution procedure

I am filing this issue to have an actionable item but also to note that it seems like as per https://github.com/census-instrumentation/opencensus-node/blob/bf1f90c65c13861e11e6a0a137935675cae17fce/packages/opencensus-exporter-zipkin/src/zipkin.ts#L142-L152

the following are missing:
* remoteEndpoint
* status
* time events (annotations, messageEvents)
* "tags" which are from OpenCensus Attributes and Status

If this is a pain to add or needs a time investment, I think we already have the OpenCensus Agent/Service https://github.com/census-instrumentation/opencensus-service and the Agent-Node.js-Exporter which has proper Zipkin exporter translation. This is precisely one of the problems that the OpenCensus Agent/Service solves so that no need to implement exporters in every language under the sun when we could use them once :)

mayurkale22:
… etc)

Fixes #183 

screen shot 2019-02-14 at 11 14 02 pm

### DIFF
--- a/packages/opencensus-exporter-zipkin/src/zipkin.ts
+++ b/packages/opencensus-exporter-zipkin/src/zipkin.ts
@@ -40,7 +40,7 @@ interface TranslatedSpan {
   id: string;
   parentId?: string;
   kind: string;
-  timestamp: number;
+  timestamp: number;  // in microseconds
   duration: number;
   debug: boolean;
   shared: boolean;
@@ -50,7 +50,7 @@ interface TranslatedSpan {
 }
 
 interface Annotation {
-  timestamp?: number;
+  timestamp?: number;  // in microseconds
   value?: string;
 }
 
@@ -177,6 +177,7 @@ export class ZipkinTraceExporter implements Exporter {
     return spanTraslated;
   }
 
+  /** Converts OpenCensus Attributes ans Status to Zipkin Tags format. */
   private createTags(
       attributes: coreTypes.Attributes, status: coreTypes.Status) {
     const tags: {[key: string]: string} = {};
@@ -190,6 +191,10 @@ export class ZipkinTraceExporter implements Exporter {
     return tags;
   }
 
+  /**
+   * Converts OpenCensus Annotation and MessageEvent to Zipkin Annotations
+   * format.
+   */
   private createAnnotations(
       annotationTimedEvents: coreTypes.Annotation[],
       messageEventTimedEvents: coreTypes.MessageEvent[]) {
--- a/packages/opencensus-exporter-zipkin/src/zipkin.ts
+++ b/packages/opencensus-exporter-zipkin/src/zipkin.ts
@@ -14,11 +14,20 @@
  * limitations under the License.
  */
 
-import {Exporter, ExporterBuffer, ExporterConfig, RootSpan, Span} from '@opencensus/core';
+import * as coreTypes from '@opencensus/core';
+import {Exporter, ExporterBuffer, ExporterConfig, RootSpan, Span, SpanKind} from '@opencensus/core';
 import {logger, Logger} from '@opencensus/core';
-import {prototype} from 'events';
 import * as http from 'http';
 import * as url from 'url';
+
+const STATUS_CODE = 'census.status_code';
+const STATUS_DESCRIPTION = 'census.status_description';
+
+const messageEventTypeTranslation: {[k: number]: string} = {
+  0: 'UNSPECIFIED',
+  1: 'SENT',
+  2: 'RECEIVED'
+};
 
 export interface ZipkinExporterOptions extends ExporterConfig {
   url?: string;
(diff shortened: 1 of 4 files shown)


packages/opencensus-exporter-zipkin/src/zipkin.ts
```diff
@@ -14,12 +14,21 @@
...
+import {Exporter, ExporterBuffer, ExporterConfig, RootSpan, Span, SpanKind} from '@opencensus/core';
 import {logger, Logger} from '@opencensus/core';
-import {prototype} from 'events';
 import * as http from 'http';
 import * as url from 'url';
 
+const STATUS_CODE = 'census.status_code';
+const STATUS_DESCRIPTION = 'census.status_description';
+
+const messageEventTypeTranslation: {[k: number]: string} = {
```
draffensperger: Nit: could you name this MESSAGE_EVENT_TYPE_TRANSLATION?

packages/opencensus-exporter-zipkin/src/zipkin.ts
```diff
@@ -138,23 +154,68 @@ export class ZipkinTraceExporter implements Exporter {
...
-  private translateSpan(span: Span|RootSpan): TranslatedSpan {
+  translateSpan(span: Span|RootSpan): TranslatedSpan {
     const spanTraslated = {
       traceId: span.traceId,
       name: span.name,
       id: span.id,
-      parentId: span.parentSpanId,
-      kind: 'SERVER',
+      // Zipkin API for span kind only accept
+      // (CLIENT|SERVER|PRODUCER|CONSUMER)
```
draffensperger: What do PRODUCER and CONSUMER mean?

packages/opencensus-exporter-zipkin/src/zipkin.ts
```diff
@@ -138,23 +154,68 @@ export class ZipkinTraceExporter implements Exporter {
...
+   * format.
+   */
+  private createAnnotations(
+      annotationTimedEvents: coreTypes.Annotation[],
+      messageEventTimedEvents: coreTypes.MessageEvent[]) {
+    let annotations: Annotation[] = [];
+    if (annotationTimedEvents) {
+      annotations =
+          annotationTimedEvents.map((annotation) => ({
+                                      timestamp: annotation.timestamp * 1000,
```
draffensperger: Nit: could you make a constant called `MICROS_PER_MILLI` or similar to use instead of `1000` directly here and the places below to show that the unit conversion is?

packages/opencensus-exporter-zipkin/test/test-zipkin.ts
```diff
@@ -109,6 +100,73 @@ describe('Zipkin Exporter', function() {
...
+
+        span.addAnnotation('processing', {}, 1550213104708);
+        span.addMessageEvent(MessageEventType.SENT, '1', 1550213104708);
+        span.addMessageEvent(MessageEventType.RECEIVED, '2', 1550213104708);
+        span.addMessageEvent(MessageEventType.UNSPECIFIED, '3', 1550213104708);
+        span.addAnnotation('done', {}, 1550213104708);
+        span.end();
+        rootSpan.end();
+
+        const spanTraslated = exporter.translateSpan(rootSpan);
```
draffensperger: Nit: `spanTranslated` not `spanTraslated` (missing an `n`)

packages/opencensus-exporter-zipkin/src/zipkin.ts
mayurkale22: I think these are used in case of messaging calls (ex: Kafka). This indicates that the request is sent to a broker instead of directly to a server and visa versa. Link to old discussion thread https://github.com/census-instrumentation/opencensus-specs/issues/54

packages/opencensus-exporter-zipkin/src/zipkin.ts
mayurkale22: Done.

packages/opencensus-exporter-zipkin/src/zipkin.ts
mayurkale22: Done.

packages/opencensus-exporter-zipkin/test/test-zipkin.ts
mayurkale22: Done

packages/opencensus-exporter-zipkin/src/zipkin.ts
draffensperger: Makes sense, then, yeah, seems that SERVER is the best we can do assuming it's required.

PR #349, opened 2019-02-19
Add option of passing in an object(SpanOptions) to the startChildSpan

mayurkale22:
This will make things consistent with ```startRootSpan``` function. Currently,  ```startChildSpan``` function takes three arguments instead of an object.

To avoid breaking changes, I have introduced three favors of ```startChildSpan``` overloaded function.


mayurkale22:
@draffensperger Thanks for the reviews

PR #352, opened 2019-02-19
gRPC Trace: correct span name format

mayurkale22:
This change is as per specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/gRPC.md#spans

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -18,11 +18,11 @@ import {BasePlugin, CanonicalCode, HeaderGetter, HeaderSetter, PluginInternalFil
 import {EventEmitter} from 'events';
 import * as grpcTypes from 'grpc';
 import * as lodash from 'lodash';
-import * as path from 'path';
-import * as semver from 'semver';
 import * as shimmer from 'shimmer';
 
 const findIndex = lodash.findIndex;
+export const SENT_PREFIX = 'Sent';
+export const RECV_PREFIX = 'Recv';
 
 //
 // Types definitions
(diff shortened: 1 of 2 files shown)


draffensperger:
Should this logic go in the Stackdriver exporter instead of the gRPC instrumentation itself?

This is how I read the specs:

- *Name the spans by package, service, method and set span kind*: "Span name is formatted as (also known as full gRPC method name): $package.$service/$method", and "Outgoing requests should be a span kind of CLIENT and incoming requests should be a span kind of SERVER."
- *For backends (like Stackdriver Trace v2) that lack span kind, use a "Sent." or "Recv." prefix to incidate it*:  "For backends that does not support Span.Kind, the exported span names can be prefixed with Sent. (for CLIENT kind) and Recv. (for SERVER kind)."

mayurkale22:
@draffensperger AFAIK, we append these prefixes (```Sent.``` and ```Recv.```) when gRPC call starts, for the client and server span name(s). Counterpart in [Java](https://github.com/grpc/grpc-java/blob/9a38dea91c6194f7a882e2be1c2846567e9cdcda/core/src/main/java/io/grpc/internal/CensusTracingModule.java#L415) and [Python](https://github.com/census-instrumentation/opencensus-python/blob/master/opencensus/trace/ext/grpc/client_interceptor.py#L252).

@songy23 correct me if I am wrong.

draffensperger:
Interesting. The go library does it in the exporter, see https://github.com/census-ecosystem/opencensus-go-exporter-stackdriver/blob/d1e19f5c23e955d7b65ff59e935007a2edce045e/trace_proto.go#L55

Based on the specs, I think the Go library is doing it correctly and the Java/Python ones should be changed eventually.

@bogdandrutu - should span names be set with `Sent.` and `Recv.` when the span is created or when it is exported to Stackdriver?

mayurkale22:
obsolete, closing this.

Issue #348, opened 2019-02-18
feat: enable strictNullChecks on this repo.


mayurkale22:
```strictNullChecks``` will help us to make the code safer (referencing nulls or undefined values), more robust and reduce some boilerplate.

- [x] opencensus-core
- [x] opencensus-nodejs
- [x] opencensus-instrumentation-http
- [x] opencensus-instrumentation-grpc
- [x] opencensus-propagation-tracecontext
- [x] opencensus-exporter-zipkin
- [x] opencensus-exporter-jaeger
- [x] opencensus-instrumentation-redis
- [x] opencensus-instrumentation-ioredis
- [x] opencensus-exporter-ocagent
- [x] opencensus-resource-util
- [x] opencensus-propagation-binaryformat
- [x] opencensus-propagation-b3
- [x] opencensus-propagation-stackdriver
- [x] opencensus-propagation-jaeger
- [x] opencensus-instrumentation-https
- [x] opencensus-instrumentation-http2
- [x] opencensus-instrumentation-mongodb
- [x] opencensus-exporter-stackdriver
- [x] opencensus-exporter-instana
- [x] opencensus-exporter-prometheus
- [x] opencensus-exporter-zpages

PR #354, opened 2019-02-20
Use binary format for grpc plugin

mayurkale22:
Based on: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/gRPC.md#propagation

This PR contains below things
1. Use default binary format instead of B3 format.
2. Enforce ```strictNullChecks``` on grpc repo. (part of #348)
3. Enforce ```noUnusedLocals``` on grpc repo.
4. Update ```index.ts``` for ```[opencensus-propagation-binaryformat]``` package.

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -479,7 +479,7 @@ export class GrpcPlugin extends BasePlugin {
    * Returns a span context on a Metadata object if it exists and is
    * well-formed, or null otherwise.
    * @param metadata The Metadata object from which span context should be
-   * retrieved.
+   *     retrieved.
    */
   static getSpanContext(metadata: grpcTypes.Metadata): SpanContext|null {
     const metadataValue = metadata.getMap()[GRPC_TRACE_KEY] as Buffer;
@@ -498,7 +498,7 @@ export class GrpcPlugin extends BasePlugin {
   /**
    * Set span context on a Metadata object if it exists.
    * @param metadata The Metadata object to which a span context should be
-   * added.
+   *     added.
    * @param spanContext The span context.
    */
   static setSpanContext(metadata: grpcTypes.Metadata, spanContext: SpanContext):
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -22,7 +22,7 @@ import * as lodash from 'lodash';
 import * as shimmer from 'shimmer';
 
 /** The metadata key under which span context is stored as a binary value. */
-const GRPC_TRACE_KEY = 'grpc-trace-bin';
+export const GRPC_TRACE_KEY = 'grpc-trace-bin';
 const findIndex = lodash.findIndex;
 
 //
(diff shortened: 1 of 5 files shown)


packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -444,7 +428,7 @@ export class GrpcPlugin extends BasePlugin {
    *  https://github.com/GoogleCloudPlatform/cloud-trace-nodejs/blob/src/plugins/plugin-grpc.ts#L96)
    */
   // tslint:disable-next-line:no-any
-  private getMetadata(original: GrpcClientFunc, args: any[], span: Span):
+  private getMetadata(original: GrpcClientFunc, args: any[]):
```
draffensperger: Rather than using `any` here, could this just be an empty object `{}`?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -490,6 +474,40 @@ export class GrpcPlugin extends BasePlugin {
   static convertGrpcStatusToSpanStatus(statusCode: grpcTypes.status): number {
     return statusCode;
   }
+
+  /**
+   * Returns a span context on a Metadata object if it exists and is
+   * well-formed, or null otherwise.
+   * @param metadata The Metadata object from which span context should be
+   * retrieved.
```
draffensperger: Nit: could you indent "retrieved" by 4 spaces to show it is a continuation of the `@param` JSDoc tag?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -490,6 +474,40 @@ export class GrpcPlugin extends BasePlugin {
...
+    if (!spanContext) {
+      return null;
+    }
+    return spanContext;
+  }
+
+  /**
+   * Set span context on a Metadata object if it exists.
+   * @param metadata The Metadata object to which a span context should be
+   * added.
```
draffensperger: Same here on the indent

packages/opencensus-instrumentation-grpc/tsconfig.json
```diff
@@ -7,7 +7,8 @@
     "pretty": true,
     "module": "commonjs",
     "target": "es6",
-    "strictNullChecks": false
+    "strictNullChecks": true,
+    "noUnusedLocals": true
```
draffensperger: Nice!! :+1:

mayurkale22:
> Could you add a test for this new header assignment logic?

Good call, done b2e8835a49cb116597696a4df88fae4d6fc703f9

packages/opencensus-instrumentation-grpc/src/grpc.ts
draffensperger: Ping on this: @mayurkale22 is it possible to use `{}` here instead of `any` and remove the tslint disable above the line? (This was already here so if you don't want to do it now that's OK, but figure it's nice to clean up if we can while we're in here).

mayurkale22:
> is it possible to use `{}` here instead of `any` and remove the tslint disable above the line? (This was already here so if you don't want to do it now that's OK, but figure it's nice to clean up if we can while we're in here

This is done in many places in gRPC plugin, I would prefer to address this in separate PR. WDYT?

draffensperger:
Yes, that's fine to leave the `any` there for now, makes sense!

PR #355, opened 2019-02-21
Add gRPC integration guide

mayurkale22:
Node version of -> https://opencensus.io/guides/grpc/go/


examples/grpc/capitalize_client.js
```diff
@@ -0,0 +1,95 @@
...
+
+const path = require('path');
+const grpc = require('grpc');
+const protoLoader = require('@grpc/proto-loader');
+const tracing = require('@opencensus/nodejs');
+const { plugin } = require('@opencensus/instrumentation-grpc');
+const { StackdriverTraceExporter } =
+    require('@opencensus/exporter-stackdriver');
+
+let tracer;
```
draffensperger: Optional: what if you made `setupOpencensusAndExporters` return the tracer so that you could assign it with `const`. Maybe call it something like `getSetupTracer` or something similar.

examples/grpc/capitalize_client.js
```diff
@@ -0,0 +1,95 @@
...
+
+const PROTO_PATH = path.join(__dirname, 'protos/defs.proto');
+const PROTO_OPTIONS = { keepCase: true, enums: String, defaults: true, oneofs: true };
+const definition = protoLoader.loadSync(PROTO_PATH, PROTO_OPTIONS);
+const rpcProto = grpc.loadPackageDefinition(definition).rpc;
+
+function main () {
+  const client = new rpcProto.Fetch('localhost:50051',
+    grpc.credentials.createInsecure());
+  let data;
```
draffensperger: Optional: could this just be `const data = process.argv[2] || 'opencensus'`? Not sure if that's actually better but feels more compact.

examples/grpc/capitalize_client.js
```diff
@@ -0,0 +1,95 @@
...
+        console.log('could not get grpc response');
+        return;
+      }
+      console.log('< ', response.data.toString('utf8'));
+      rootSpan.end();
+    });
+  });
+
+  setTimeout(() => {
+    console.log('done.');
```
draffensperger: What's the role of this message / is there a better way to check for doneness than waiting one minute?

examples/grpc/capitalize_client.js
```diff
@@ -0,0 +1,95 @@
...
+}
+
+function setupOpencensusAndExporters () {
+  // Enable OpenCensus exporters to export traces to Stackdriver CloudTrace.
+  // Exporters use Application Default Credentials (ADCs) to authenticate.
+  // See https://developers.google.com/identity/protocols/application-default-credentials
+  // for more details.
+  // Expects ADCs to be provided through the environment as ${GOOGLE_APPLICATION_CREDENTIALS}
+  // A Stackdriver workspace is required and provided through the environment as ${GOOGLE_PROJECT_ID}
+  const projectId = process.env.GOOGLE_PROJECT_ID;
```
draffensperger: Would it be worth specifying above that this is intended to run on the Google Cloud Platform? Or is that already implied somewhere?

examples/grpc/capitalize_server.js
```diff
@@ -0,0 +1,88 @@
...
+
+const path = require('path');
+const grpc = require('grpc');
+const protoLoader = require('@grpc/proto-loader');
+const tracing = require('@opencensus/nodejs');
+const { plugin } = require('@opencensus/instrumentation-grpc');
+const { StackdriverTraceExporter } =
+    require('@opencensus/exporter-stackdriver');
+
+let tracer;
```
draffensperger: Similar optional comment here on assigning via `const` with the return value of the setup function.

examples/grpc/README.md
```diff
@@ -2,7 +2,7 @@
 
 Our service takes in a payload containing bytes and capitalizes them.
 
-Using OpenCensus Node, we can collect traces of our system and export them to the backend of our choice, to give observability to our distributed systems.
+Using OpenCensus Node, we can collect traces of our system and export them to the backend of our choice(we are using Stackdriver for this example), to give observability to our distributed systems.
```
draffensperger: Nit: add space

examples/grpc/capitalize_client.js
mayurkale22: The minimum reporting period for Stackdriver is 1 minute. The thread with the ```StackdriverStatsExporter``` must live for at least the interval past any metrics that must be collected, or some risk being lost if they are recorded after the last export. We planned to add a ```flush``` method instead of doing this.

I will enable ```StackdriverStatsExporter``` here, once the gRPC stats are available #270

examples/grpc/capitalize_server.js
mayurkale22: Done.

examples/grpc/capitalize_client.js
draffensperger: OK, interesting. Can you add a comment to explain this?

mayurkale22: Done.

examples/grpc/README.md
mayurkale22: Done

PR #357, opened 2019-02-23
Add support for recording gRPC stats

### LINKED ISSUE #270: Add instrumentation for gRPC on propagating tags and recording gRPC stats

mayurkale22:
This PR is to get the ball rolling, Unit tests are still pending. Reviews are welcome at this moment.
Fixes #270 

```client-metrics.ts``` and ```server-metrics.ts```  are updated with minor nits.

Update: Tests are included now.

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -14,12 +14,16 @@
  * limitations under the License.
  */
 
-import {BasePlugin, CanonicalCode, PluginInternalFiles, RootSpan, Span, SpanContext, SpanKind, TraceOptions} from '@opencensus/core';
+import {BasePlugin, CanonicalCode, PluginInternalFiles, RootSpan, Span, SpanContext, SpanKind, TagMap, TraceOptions} from '@opencensus/core';
 import {deserializeSpanContext, serializeSpanContext} from '@opencensus/propagation-binaryformat';
 import {EventEmitter} from 'events';
 import * as grpcTypes from 'grpc';
 import * as lodash from 'lodash';
 import * as shimmer from 'shimmer';
+import * as clientMetrics from './grpc-stats/client-metrics';
+import * as serverMetrics from './grpc-stats/server-metrics';
+
+const sizeof = require('object-sizeof');
 
 /** The metadata key under which span context is stored as a binary value. */
 export const GRPC_TRACE_KEY = 'grpc-trace-bin';
@@ -44,7 +48,9 @@ type ServerCall = typeof grpcTypes.ServerUnaryCall|
 
 type ServerCallWithMeta = ServerCall&{
   metadata: grpcTypes.Metadata;
-  status: Status
+  status: Status;
+  // tslint:disable-next-line:no-any
+  request?: any
 }
 &EventEmitter;
 
@@ -203,6 +209,7 @@ export class GrpcPlugin extends BasePlugin {
       plugin: GrpcPlugin, rootSpan: RootSpan, call: ServerCallWithMeta,
       callback: SendUnaryDataCallback,
       original: grpcTypes.handleCall<RequestType, ResponseType>, self: {}) {
+    const startTime = Date.now();
     function patchedCallback(
         err: grpcTypes.ServiceError,
         // tslint:disable-next-line:no-any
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: packages/opencensus-instrumentation-grpc/src/grpc.ts [hunk 1]
```diff
@@ -14,12 +14,16 @@
...
+import {BasePlugin, CanonicalCode, PluginInternalFiles, RootSpan, Span, SpanContext, SpanKind, TagMap, TraceOptions} from '@opencensus/core';
 import {deserializeSpanContext, serializeSpanContext} from '@opencensus/propagation-binaryformat';
 import {EventEmitter} from 'events';
 import * as grpcTypes from 'grpc';
 import * as lodash from 'lodash';
 import * as shimmer from 'shimmer';
+import * as clientMetrics from './grpc-stats/client-metrics';
+import * as serverMetrics from './grpc-stats/server-metrics';
+
+const sizeof = require('object-sizeof');
```
draffensperger: Is this possible to do as a TypeScript `import`?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -44,7 +48,9 @@ type ServerCall = typeof grpcTypes.ServerUnaryCall|
 
 type ServerCallWithMeta = ServerCall&{
   metadata: grpcTypes.Metadata;
-  status: Status
+  status: Status;
+  // tslint:disable-next-line:no-any
+  request?: any
```
draffensperger: Could this be `unknown` or the empty object type `{}` instead?

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -507,6 +531,64 @@ export class GrpcPlugin extends BasePlugin {
       metadata.set(GRPC_TRACE_KEY, serializedSpanContext);
     }
   }
+
+  /** Method to record stats for client and server. */
+  static recordStats(
+      // tslint:disable-next-line:no-any
+      kind: SpanKind, tags: TagMap, argsOrValue: any, reqOrRes: any,
```
draffensperger: Could you use the empty object type `{}` in place of `any` here? (That should work because you are not actually referencing anything about the type, just transparently passing it into `sizeof`)

packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -223,6 +230,14 @@ export class GrpcPlugin extends BasePlugin {
             GrpcPlugin.ATTRIBUTE_GRPC_STATUS_CODE,
             grpcTypes.status.OK.toString());
       }
+
+      // record stats
+      const tags = new TagMap();
+      tags.set(serverMetrics.GRPC_SERVER_METHOD, {value: rootSpan.name});
```
draffensperger: As an aside: I think this use case is a good reason to not add `Sent.` or `Recv.` to the gRPC span names until they are ready to be exported.

packages/opencensus-instrumentation-grpc/src/grpc.ts
mayurkale22: ```object-sizeof``` package is missing typings. I tried to use https://github.com/miktam/sizeof/blob/master/index.d.ts but didn;t work.

packages/opencensus-instrumentation-grpc/src/grpc.ts
mayurkale22: SGTM, done

packages/opencensus-instrumentation-grpc/src/grpc.ts
mayurkale22: hmm, makes sense. Looks like need to revise the Java's implementation.

packages/opencensus-instrumentation-grpc/src/grpc.ts
draffensperger: OK, makes sense.

packages/opencensus-instrumentation-grpc/test/test-grpc.ts
```diff
@@ -14,19 +14,44 @@
...
+import * as clientMetrics from '../src/grpc-stats/client-metrics';
+import * as serverMetrics from '../src/grpc-stats/server-metrics';
+import {registerAllGrpcViews} from '../src/grpc-stats/stats-common';
 
 const PROTO_PATH = __dirname + '/fixtures/grpc-instrumentation-test.proto';
 const grpcPort = 50051;
 const MAX_ERROR_STATUS = grpcModule.status.UNAUTHENTICATED;
 const log = logger.logger();
 
+
```
draffensperger: Nit: is this extra blank line intentional?

PR #358, opened 2019-02-24
Fix mongodb instrumentation not ending span correctly

vmarchaud:
From my tests all the mongodb queries had the wrong duration so i inspected the code and found out that the span were never ended and automatically truncated when the root span is ended.

So i rewrote a little bit the patchs so they could correctly be ended and added the verification in the tests

cc @draffensperger @mayurkale22

### DIFF
--- a/packages/opencensus-instrumentation-mongodb/src/mongodb.ts
+++ b/packages/opencensus-instrumentation-mongodb/src/mongodb.ts
@@ -15,9 +15,7 @@
  */
 
 import {BasePlugin, Func, Span, SpanKind} from '@opencensus/core';
-import {logger, Logger} from '@opencensus/core';
 import * as mongodb from 'mongodb';
-import * as semver from 'semver';
 import * as shimmer from 'shimmer';
 
 export type MongoDB = typeof mongodb;
@@ -26,7 +24,6 @@ export type MongoDB = typeof mongodb;
 export class MongoDBPlugin extends BasePlugin {
   private readonly SERVER_FNS = ['insert', 'update', 'remove', 'auth'];
   private readonly CURSOR_FNS_FIRST = ['_find', '_getmore'];
-  private readonly SPAN_MONGODB_QUERY_TYPE = 'db.mongodb.query';
 
   /** Constructs a new MongoDBPlugin instance. */
   constructor(moduleName: string) {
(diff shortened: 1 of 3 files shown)


packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -136,11 +150,11 @@ export class MongoDBPlugin extends BasePlugin {
         if (plugin.tracer.currentRootSpan && arguments.length > 0 &&
             typeof resultHandler === 'function') {
           const span = plugin.tracer.startChildSpan(
-              {name: `${this.ns}.cursor`, kind: SpanKind.SERVER});
+              `${this.ns}.cursor`, SpanKind.SERVER);
```
mayurkale22: Can we use the object for ```startChildSpan```function instead of passing two/three arguments?

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
```diff
@@ -77,10 +77,10 @@ export class MongoDBPlugin extends BasePlugin {
     return (original: Func<mongodb.Server>) => {
       return function(
                  // tslint:disable-next-line:no-any
-                 this: mongodb.Server, ns: string, command: any,
-                 // tslint:disable-next-line:no-any
-                 ...args: any[]): mongodb.Server {
-        let resultHandler = args[args.length - 1];
+                 this: mongodb.Server, ns: string, command: any, options: any,
```
draffensperger: Could the type of `options` be `Function|{}` instead of `any`?

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
```diff
@@ -74,6 +74,10 @@ function assertSpan(
     expectedKind: SpanKind) {
   assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 1);
   assert.strictEqual(rootSpanVerifier.endedRootSpans[0].spans.length, 1);
+  // @ts-ignore
```
draffensperger: What dos `@ts-ignore` do here?

vmarchaud: Because `span.truncated` is a private field, but i need it to verify that the span are correctly ended. The easiest way was to just ignore the error.

packages/opencensus-instrumentation-mongodb/src/mongodb.ts
vmarchaud: We can't with the current API, it has changed with the version 0.0.9

vmarchaud:
Test failures are related to Stackdriver so i believe we can ignore them for this PR

packages/opencensus-instrumentation-mongodb/test/test-mongodb.ts
draffensperger: OK. Could you add a comment to explain that?

vmarchaud: Done

mayurkale22:
> Test failures are related to Stackdriver so i believe we can ignore them for this PR

I am working on fixing the build.

mayurkale22:
Can you please rebase your branch with the master, hopefully it will fix build.

vmarchaud:
I believe the PR can be merged now ?

vmarchaud:
Ping @mayurkale22

mayurkale22:
@vmarchaud Thanks for the contributions, PR looks good. If no other comments from @draffensperger I will merge this PR later today.

PR #360, opened 2019-02-26
Add Measure, TagKey and View constants for recording HTTP stats

mayurkale22:
Specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/HTTP.md

This is part of #269


packages/opencensus-instrumentation-http/src/http-metrics.ts
```diff
@@ -0,0 +1,239 @@
...
+/**
+ * {@link TagKey} for the server-side logical route, a pattern that matched the
+ * URL, of a handler that processed the request.
+ */
+export const HTTP_SERVER_ROUTE = {
+  name: 'http_server_route'
+};
+
+const SIZE_DISTRIBUTION: number[] = [
+  0, 1024, 2048, 4096, 16384, 65536, 262144, 1048576, 4194304, 16777216,
```
songy23: Not sure if we still want to include the leading 0s?

packages/opencensus-instrumentation-http/src/http-metrics.ts
```diff
@@ -0,0 +1,239 @@
...
+
+/** Register all default views. */
+export function registerAllViews(globalStats: Stats) {
+  registerAllClientViews(globalStats);
+  registerAllServerViews(globalStats);
+}
+
+/** Register all default client views. */
+export function registerAllClientViews(globalStats: Stats) {
+  for (const CLIENT_VIEW of HTTP_BASIC_CLIENT_VIEWS) {
```
draffensperger: Optional nit: should we use `ALL_CAPS` for `const` declarations that are local to a function (here and for `registerAllServerViews` below)?

mayurkale22: Done

packages/opencensus-instrumentation-http/src/http-metrics.ts
mayurkale22: I would keep it now, until we officially remove this from the specs. 
Context (for the record) -> we are already [dropping negative](https://github.com/census-instrumentation/opencensus-node/blob/2df1bc4293dbfe43081bed8e98cea551f6602f5c/packages/opencensus-core/src/stats/bucket-boundaries.ts#L65) and zero bucket bounds in top level API, so it doesn't make sense to include 0 at first place. Also [specs](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/metrics/v1/metrics.proto#L223) says "**The values must be strictly increasing and > 0**".

rghetia: Lets remove it. We can open another PR to update the specs. For now we have two conflicting specs (one for metrics proto and one for http/grpc stats). Since we don't have any useful data in zero bucket might as well remove it.

mayurkale22: Ok, done.

PR #363, opened 2019-02-26
Fix stackdriver test

mayurkale22:


draffensperger:
Do we know why this broke?

mayurkale22:
Looks like some internal auth library changed/updated, didn't spend long time in investigation. I hope that's ok. :)

draffensperger:
Ah, I see. Maybe as a follow-up, should that test be checking for a 403 response code rather than a string in the response?

mayurkale22:
> Ah, I see. Maybe as a follow-up, should that test be checking for a 403 response code rather than a string in the response?

Added check for response code. It's actually 401 (Unauthorized). :)

Issue #174, opened 2018-11-02
Node package folder location assumptions are broken


dnduffy:
The code in package `@opencensus/exporter-ocagent` has hardcoded relative paths with the incorrect number of back references (..) when looking for `@opencensus/core/package.json`.
To get around this first problem you can run `npm install --legacy-bundling` (but you should not need to do this).

Next it will fail to find proto on ocagent.js:86 because the directory `../../../src/protos/` does not exist (nor does it exist with only two back references).  This folder exists in the repository but is not packaged into the NPM package.
The workaround for this is to copy the folder from the repo into `node_modules/@opencensus/exporter-ocagent/src/protos`

justindsmith:
@dnduffy Thanks for the submitted issue! Sorry it took a bit of time to get around to it. I've just pushed a fix that will get picked up in the next release.

PR #366, opened 2019-02-27
Fix the exporter-ocagent

justindsmith:
The `exporter-ocagent` had two critical issues which left the exporter unusable:
1. The required grpc proto files were not being referenced correctly (#174)
2. The `ConfigStream` behavior (where the ocagent can send configuration changes down to the client) was not connecting / receiving correctly from the ocagent, causing errors and spamming logs

The following changes have been applied to fix the above issues:
1. We now package the proto files for both OpenCensus and the google common proto set into the final `@opencensus/exporter-ocagent` package. The changes mostly reside in the `package.json` configuration.
2. We have removed all references to the `ConfigStream` behavior. This is a breaking change, but the ocagent exporter was not in a usable form previously, so I don't expect there to be any users directly impacted.

A few other changes have been included as well:
* The default ocagent port has been changed to `55678` to match the default OC Agent/Collector port.
* A slight reconnect delay has been added to avoid rapid reconnect attempts.
* Updated `grpc` dependency version to `1.18.0`.
* Readme / Docs update
* Changelog update

These changes have all been tested locally against a real local ocagent instance and been confirmed to be working.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Add support for ```tags```, ```status``` and ```annotation``` in Zipkin exporter.
 - Add support for Binary propagation format.
 - Add support for object(```SpanOptions```) as an argument for ```startChildSpan``` function, similar to ```startRootSpan```.
+- Add proto files to exporter-ocagent package. Fixes issue #174. 
 
 ## 0.0.9 - 2019-02-12
 - Add Metrics API.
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,9 @@ All notable changes to this project will be documented in this file.
 - Add support for ```tags```, ```status``` and ```annotation``` in Zipkin exporter.
 - Add support for Binary propagation format.
 - Add support for object(```SpanOptions```) as an argument for ```startChildSpan``` function, similar to ```startRootSpan```.
+- Add proto files to exporter-ocagent package. Fixes issue #174.
```
draffensperger: Nit, should `issue #174` be a link? Or is that done automatically?

justindsmith: Good point, likely not done automatically. I'll add as link.

justindsmith:
> LGTM. If possible, please include `exporter-ocagent` example in examples folder?

I took a quick look and didn't see any tracing specific examples. I don't think having an exporter specific example is useful without the larger tracing context, so I'm going to hold off on adding an example, but I'll add a tracking issue to add a tracing-specific example, as I think that would be useful.

PR #369, opened 2019-02-27
Add HTTP instrumentation guide

### LINKED ISSUE #361: Add HTTP instrumentation example
Similar to gRPC -> https://github.com/census-instrumentation/opencensus-node/tree/master/examples/grpc

mayurkale22:
Node version of -> https://opencensus.io/guides/http/go/net_http/

Fixes #361


examples/http/client.js
```diff
@@ -0,0 +1,69 @@
...
+const tracing = require('@opencensus/nodejs');
+const { plugin } = require('@opencensus/instrumentation-http');
+const { ZipkinTraceExporter } = require('@opencensus/exporter-zipkin');
+const { TraceContextFormat } = require('@opencensus/propagation-tracecontext');
+
+const tracer = setupTracerAndExporters();
+
+/** A function which makes requests and handles response. */
+function makeRequest () {
+  tracer.startRootSpan({ name: 'octutorialsClient.makeRequest' }, rootSpan => {
```
draffensperger: On first reading, it looked to me like the `tracer.startRootSpan` was starting a span for the outgoing request, but really it's a parent span for it.

Could you add a comment that explains what the root span is here and that it will get a child span for the HTTP request?

examples/http/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "http-example",
```
draffensperger: Is the package won't be published to NPM, should it have a name? Or is there another way to communicate that a package isn't published?

examples/http/server.js
```diff
@@ -0,0 +1,84 @@
...
+const http = require('http');
+const tracing = require('@opencensus/nodejs');
+const { plugin } = require('@opencensus/instrumentation-http');
+const { ZipkinTraceExporter } = require('@opencensus/exporter-zipkin');
+const { TraceContextFormat } = require('@opencensus/propagation-tracecontext');
+
+const tracer = setupTracerAndExporters();
+
+/** Starts a HTTP server that receives requests on sample server port. */
+function start (port) {
```
draffensperger: Maybe call this `startServer` so it's clearer at first glance? Nit: also extra space before parens

mayurkale22: Sure, done.

examples/http/package.json
mayurkale22: Yes, this example package won't be published to NPM.

examples/http/client.js
mayurkale22: Added, PTAL

draffensperger: :+1:

PR #370, opened 2019-02-28
Add support for recording HTTP stats

mayurkale22:
This PR provides instrumentation for HTTP Client and Server to record HTTP stats. Part of https://github.com/census-instrumentation/opencensus-node/issues/269

These changes have all been tested locally against prometheus exporter and been confirmed to be working.

screen shot 2019-02-27 at 11 11 41 pm

I will work on recording stats for send/received bytes in seperate PR.

### DIFF
--- a/packages/opencensus-instrumentation-http/src/http.ts
+++ b/packages/opencensus-instrumentation-http/src/http.ts
@@ -15,7 +15,7 @@
  */
 
 import {BasePlugin, CanonicalCode, Func, HeaderGetter, HeaderSetter, MessageEventType, Span, SpanKind, TagMap, TraceOptions} from '@opencensus/core';
-import * as httpModule from 'http';
+import {ClientRequest, ClientResponse, IncomingMessage, request, RequestOptions, ServerResponse} from 'http';
 import * as semver from 'semver';
 import * as shimmer from 'shimmer';
 import * as url from 'url';
@@ -23,12 +23,10 @@ import * as uuid from 'uuid';
 import * as stats from './http-stats';
 import {IgnoreMatcher} from './types';
 
-export type HttpGetCallback = (res: httpModule.IncomingMessage) => void;
-export type HttpModule = typeof httpModule;
-export type RequestFunction = typeof httpModule.request;
+export type HttpGetCallback = (res: IncomingMessage) => void;
+export type RequestFunction = typeof request;
 
-// tslint:disable-next-line:no-any
-function isOpenCensusRequest(options: any) {
+function isOpenCensusRequest(options: RequestOptions) {
   return options && options.headers &&
       !!options.headers['x-opencensus-outgoing-request'];
 }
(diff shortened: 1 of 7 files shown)


packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -14,18 +14,25 @@
...
-import {HttpPluginConfig, IgnoreMatcher} from './types';
+import * as stats from './http-stats';
+import {IgnoreMatcher} from './types';
 
 export type HttpGetCallback = (res: httpModule.IncomingMessage) => void;
 export type HttpModule = typeof httpModule;
 export type RequestFunction = typeof httpModule.request;
 
+// tslint:disable-next-line:no-any
+function isOpenCensusRequest(options: any) {
```
draffensperger: What would you think about making a simple interface for this rather than passing in `any` here? Seems like it just has an optional `headers` attribute that is a map of string to string.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -341,10 +353,10 @@ export class HttpPlugin extends BasePlugin {
    * @param options The arguments to the original function.
    */
   private getMakeRequestTraceFunction(
-      // tslint:disable-next-line:no-any
       request: httpModule.ClientRequest, options: httpModule.RequestOptions,
       plugin: HttpPlugin): Func<httpModule.ClientRequest> {
     return (span: Span): httpModule.ClientRequest => {
+      const startTime = Date.now();
```
draffensperger: Could this be a high-accuracy timestamp? I believe some Node versions support `performance.now()`, which would be an easy swap in, or else you could use `process.hrtime` and do some conversion / use a utility.

packages/opencensus-instrumentation-http/src/index.ts
```diff
@@ -15,3 +15,6 @@
  */
 
 export * from './http';
+import {registerAllClientViews, registerAllServerViews, registerAllViews} from './http-stats';
+
+export {registerAllClientViews, registerAllServerViews, registerAllViews};
```
draffensperger: Could this just be `export {registerAllClientViews, registerAllServerViews, registerAllViews} from './http-stats';`?

packages/opencensus-instrumentation-http/src/http.ts
mayurkale22: Changed to ```RequestOptions```, instead of ```any```.

packages/opencensus-instrumentation-http/src/index.ts
mayurkale22: Sure, Done.

packages/opencensus-instrumentation-http/src/http.ts
mayurkale22: For stats recoding, we want to record ```ms``` level granularity. Looks like Date.now() gives us that. WDYT?

draffensperger: Yes, that's fine. Didn't realize the stats were only at `ms` granularity.

PR #386, opened 2019-03-04
pass stats instance to plugin loader

mayurkale22:
Allow a way to pass ```stats``` instance to ```PluginLoader``` class, this will help users to collect stats for HTTP and GRPC plugin with automatic tracing. 

Also, I have updated the HTTP instrumentation example to capture traces without manually enabling the plugin.  I think this is the recommended and straightforward way for users to enable tracing on supported plugins/frameworks.

### DIFF
--- a/examples/http/package.json
+++ b/examples/http/package.json
@@ -21,7 +21,6 @@
   },
   "dependencies": {
     "@opencensus/exporter-zipkin": "^0.0.9",
-    "@opencensus/instrumentation-http": "^0.0.9",
     "@opencensus/nodejs": "^0.0.9",
     "@opencensus/propagation-tracecontext": "^0.0.9",
     "http": "*"
--- a/packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
+++ b/packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-import {Logger, Plugin, PluginConfig, PluginNames, Tracer} from '@opencensus/core';
+import {Logger, Plugin, PluginConfig, PluginNames, Stats, Tracer} from '@opencensus/core';
 import * as fs from 'fs';
 import * as path from 'path';
 import * as hook from 'require-in-the-middle';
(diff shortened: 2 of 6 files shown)


examples/http/package.json
```diff
@@ -21,7 +21,6 @@
   },
   "dependencies": {
     "@opencensus/exporter-zipkin": "^0.0.9",
-    "@opencensus/instrumentation-http": "^0.0.9",
```
draffensperger: Out of curiosity, why is this no longer needed?

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -49,9 +51,10 @@ export class PluginLoader {
    * Constructs a new PluginLoader instance.
    * @param tracer The tracer.
    */
-  constructor(logger: Logger, tracer: Tracer) {
+  constructor(logger: Logger, tracer: Tracer, stats: Stats|undefined) {
```
draffensperger: Why not just have it be `stats?: Stats`?

packages/opencensus-nodejs/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -140,8 +142,14 @@ export class PluginLoader {
         try {
           const plugin: Plugin = require(moduleName as string).plugin;
           this.plugins.push(plugin);
-          return plugin.enable(
-              exports, this.tracer, version, moduleConfig, basedir);
+          if (this.stats) {
```
draffensperger: Changing the type signature above to make `stats` optional would also allow collapsing down the two branches of this `if`

packages/opencensus-nodejs/test/test-plugin-loader.ts
```diff
@@ -94,7 +94,7 @@ describe('Plugin Loader', () => {
       });
 
       it('should load a plugin and patch the target modules', () => {
-        const pluginLoader = new PluginLoader(log, tracer);
+        const pluginLoader = new PluginLoader(log, tracer, undefined);
```
draffensperger: These changes to pass `undefined` would also be unnecessary with it being optional.

examples/http/package.json
mayurkale22: There are two ways to instrument supported plugins. 1) Manual approach: include respective dependency and use ```plugin.enable(...)``` interface. 2) Allow the library to instrument plugins automatically, for this start the trace instance as the first module in your application.

The changed the HTTP example to work with second (I think recommended) approach.

PR #389, opened 2019-03-05
Add a NoRecord RootSpan and Span

### LINKED ISSUE #388: Add a no-op RootSpan and Span
Currently, we return ```null``` when trace is not sampled or set to ```NeverSampler``` (_Sampler that samples no traces_). This will allow user to instrument a library without enabling sampling or using ```ProbabilitySampler``` (_Sampler that samples a given fraction of traces_).

All operations should be no-op.

mayurkale22:
This will allow user to instrument a library without enabling sampling (i.e. using ```NeverSampler```) or using ```ProbabilitySampler``` (Sampler that samples a given fraction of traces).

Fixes #388

### DIFF
--- a/packages/opencensus-core/src/trace/model/no-op/noop-span-base.ts
+++ b/packages/opencensus-core/src/trace/model/no-op/noop-span-base.ts
@@ -1,0 +1,151 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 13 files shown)


packages/opencensus-core/src/trace/model/no-op/noop-span-base.ts
```diff
@@ -0,0 +1,151 @@
...
+import * as configTypes from '../../config/types';
+import * as types from '../types';
+
+const STATUS_OK = {
+  code: types.CanonicalCode.OK
+};
+
+/** No-Op implementation of the SpanBase */
+export abstract class NoopSpanBase implements types.Span {
+  protected className: string;
```
draffensperger: What role does this field play? Can it be removed?

mayurkale22:
@draffensperger I had discussion with @songy23 and @rghetia regrading this. We have decided to rename classes from Noop to NoRecord (ex: ```NoopSpan -> NoRecordSpan```). Basically, no-op means not propagating anything on the wire, while NoRecord will provide a mechanism to propagate ```traceId``` and ```spanId``` on the wire and all other interfaces/functions are not doing anything.

draffensperger:
Sounds good. Glad there is cross-language collaboration going on with this. Propagating trace and span IDs even in cases when requests are not sampled is important for logs correlation.

mayurkale22:
Build failure is unrelated to this PR, will handle it later.

PR #394, opened 2019-03-06
Add support for overriding sampling per span

### LINKED ISSUE #391: Override sampling for a span
Specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#how-can-users-control-the-sampler-that-is-used-for-sampling

mayurkale22:
Specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/trace/Sampling.md#how-can-users-control-the-sampler-that-is-used-for-sampling

The logic for allocating ```traceId``` moved in ```tracer``` class. This helps to make sampling decision first and depending on decision we can then create either ```RecordRootSpan(RootSpan)``` or ```NoRecordRootSpan``` instead of always creating ```RootSpan```(just to get traceId for sampling decision) and discarding it later. 

This introduces signature changes to internal ```RootSpan``` class, But I think this won't be breaking change for users.

Fixes #391

### DIFF
--- a/packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
+++ b/packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
@@ -14,7 +14,6 @@
  * limitations under the License.
  */
 
-import * as uuid from 'uuid';
 import * as logger from '../../../common/console-logger';
 import * as types from '../types';
 import {NoRecordSpan} from './no-record-span';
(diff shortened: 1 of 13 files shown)


packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
```diff
@@ -35,22 +34,25 @@ export class NoRecordRootSpan extends NoRecordSpanBase implements
...
+   * @param name The displayed name for the new span.
+   * @param kind The kind of new span.
+   * @param traceId The trace Id.
+   * @param parentSpanId The id of the parent span, or empty if the new span is
+   *     a root span.
+   * @param traceState Optional traceState.
    */
-  constructor(tracer: types.Tracer, context?: types.TraceOptions) {
+  constructor(
+      tracer: types.Tracer, name: string, kind: types.SpanKind, traceId: string,
```
draffensperger: Should these new parameters be optional to make this less of a breaking change? If we keep them, could you modify the change list to include a mention of it?

packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -232,6 +232,8 @@ export interface TraceOptions {
   spanContext?: SpanContext;
   /** Span kind */
   kind?: SpanKind;
+  /** Determines the sampling rate. Ranges from 0.0 to 1.0 */
+  samplingRate?: number;
```
draffensperger: Should this be a `Sampler` instance per the specs?

mayurkale22: I am happy to change here, but the global sampler takes the same input https://github.com/census-instrumentation/opencensus-node/blob/f98ec08248d7e962738278473d762ab449eabe49/packages/opencensus-core/src/trace/config/types.ts#L35 To keep things consistent I use ```number```. WDYT?

draffensperger: Yeah, I'm OK with that, but could you create an issue to track changing it in both places to bring it closer to the specs?

packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
mayurkale22: Possible. On the other hand, this is internal API and won't affect users in any way.

packages/opencensus-core/src/trace/model/types.ts
mayurkale22: sure.

packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
draffensperger: OK, I see, this does not change the `RootSpan` interface type, just the implementations. In OpenCensus Web, I create spans directly from performance timings, but it seems like the instrumented Node libraries use the helper utilities in `Tracer` to create, start and end spans. So makes sense this is not really part of the exported API.

PR #395, opened 2019-03-06
Fix redis test when redis is not running locally

mayurkale22:


mayurkale22:
@vmarchaud Please review

packages/opencensus-instrumentation-redis/test/test-redis.ts
```diff
@@ -55,14 +55,30 @@ function assertSpan(
...
+ * @param url The redis URL to access.
+ */
+function accessCollection(url: string): Promise<redis.RedisClient> {
+  return new Promise((resolve, reject) => {
+    const client = redis.createClient({url});
+    client.on('error', (err) => {
+      reject(err);
+      return;
+    });
+    resolve(client);
```
vmarchaud: you should listen for the `ready` event to be able to resolve the promise i think, otherwise it might resolve once before catching the error.

mayurkale22: Done, PTAL

vmarchaud: Fine for me

packages/opencensus-instrumentation-redis/test/test-redis.ts
```diff
@@ -55,14 +55,32 @@ function assertSpan(
...
+/**
+ * Access the redis client.
+ * @param url The redis URL to access.
+ */
+function accessCollection(url: string): Promise<redis.RedisClient> {
+  return new Promise((resolve, reject) => {
+    const client = redis.createClient({url});
+    client.on('error', (err) => {
+      reject(err);
+      return;
```
draffensperger: Could you remove this extra `return;` at the end of the arrow function?

mayurkale22:
@vmarchaud Looks like https://github.com/census-instrumentation/opencensus-node/pull/410#discussion_r265519126 is valid in this case also. WDYT? I am ok to close this PR if we are going with that route.

vmarchaud:
@mayurkale22 Yep i believe we can close this one

PR #401, opened 2019-03-07
Upgrade grpc to 1.19

duxet:
I've upgraded grpc package to newest version, because 1.13 doesn't provide binaries for Node.js 11.

I had to adjust `protobuf` package version to fix the build: https://github.com/grpc/grpc-node/issues/176


packages/opencensus-instrumentation-grpc/tsconfig.json
```diff
@@ -5,6 +5,7 @@
     "rootDir": ".",
     "outDir": "build",
     "pretty": true,
+    "lib": ["es2016", "dom"],
```
draffensperger: Why is `dom` needed here?

duxet: It's needed to provide `XMLHttpRequest` type definition for protobufjs: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/dbabb10c626deb803e817f214bb495b65c4a7d4f/types/protobufjs/index.d.ts#L275

draffensperger: OK, makes sense.

PR #403, opened 2019-03-08
Code Enhancement/Refactor

mayurkale22:
This PR contains following things
1. Remove unsed ```parentSpanId``` from ```startChildSpan``` API.
2. Code refactor/enhancement.
3. Enforce ```noUnusedLocals ``` on opencensus-nodejs package.

### DIFF
--- a/packages/opencensus-core/src/trace/model/tracer.ts
+++ b/packages/opencensus-core/src/trace/model/tracer.ts
@@ -39,9 +39,6 @@ export class CoreTracer implements types.Tracer {
   private config: configTypes.TracerConfig;
   /** A list of end span event listeners */
   private eventListenersLocal: types.SpanEventListener[] = [];
-  /** A list of ended root spans */
-  // @ts-ignore
-  private endedTraces: types.RootSpan[] = [];
   /** Bit to represent whether trace is sampled or not. */
   private readonly IS_SAMPLED = 0x1;
   /** A sampler used to make sample decisions */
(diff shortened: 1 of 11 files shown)


packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -162,34 +159,30 @@ export class CoreTracer implements types.Tracer {
...
-        return this.logger.debug('cannot start trace - no active trace found');
-      }
-      if (this.currentRootSpan !== root) {
-        this.logger.debug(
-            'currentRootSpan != root on notifyStart. Need more investigation.');
-      }
-      this.notifyStartSpan(root);
+    if (!this.active) return;
+    if (!root) {
+      return this.logger.debug('cannot start trace - no active trace found');
```
draffensperger: What would you think about separating the `return` from the logger statement? Though it compiles like this, it looks unexpected to me as if it were returning a non-undefined value from the logger statement.

packages/opencensus-nodejs/src/trace/tracing.ts
```diff
@@ -33,12 +32,12 @@ export class Tracing implements core.Tracing {
   private defaultPlugins: core.PluginNames;
   /** A configuration object to start the tracing */
   private configLocal: core.Config = {};
-  /** An object to log information to */
-  private logger: core.Logger = null;
+  /** An object to log information to. Logs to the JS console by default. */
+  private logger: core.Logger = console;
```
draffensperger: Optional: I believe there is also a logger instance in https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/common/console-logger.ts

That does formatting a bit differently from the default `console` instance I believe. I used `console` in OC Web because I didn't want to introduce extra bytes of code for the logger, but that's not really much of an issue for Node.

packages/opencensus-core/src/trace/model/tracer.ts
mayurkale22: Good catch, done.

packages/opencensus-nodejs/src/trace/tracing.ts
mayurkale22: Makes sense, done.

PR #405, opened 2019-03-08
 Add support for recording exemplars

### LINKED ISSUE #385: Support recording Exemplars
Record and store exemplar.

[Proto definition](https://github.com/census-instrumentation/opencensus-proto/blob/master/opencensus/proto/stats/metrics/metrics.proto#L236-L250):

```
  // Exemplars are example points that may be used to annotate aggregated
  // Distribution values. They are metadata that gives information about a
  // particular value added to a Distribution bucket.
  message Exemplar {
    // Value of the exemplar point. It determines which bucket the exemplar
    // belongs to.
    double value = 1;

    // The observation (sampling) time of the above value.
    google.protobuf.Timestamp timestamp = 2;

    // Contextual information about the example value.
    map<string, string> attachments = 3;
  }
```
Counterpart: https://github.com/census-instrumentation/opencensus-python/pull/287

https://github.com/census-instrumentation/opencensus-node/blob/77eada8756142f5f9237d536f91abf5de0e1acc2/packages/opencensus-core/src/stats/view.ts#L233

mayurkale22:
Specs: https://github.com/census-instrumentation/opencensus-specs/blob/18ddbd73ffc6786622fb17a9f1798d6b668fded2/stats/Exemplars.md

Fixes #385

### DIFF
--- a/packages/opencensus-core/src/stats/types.ts
+++ b/packages/opencensus-core/src/stats/types.ts
@@ -65,10 +65,15 @@ export interface Stats {
    * Updates all views with the new measurements.
    * @param measurements A list of measurements to record
    * @param tags optional The tags to which the value is applied.
-   *  tags could either be explicitly passed to the method, or implicitly
-   *  read from current execution context.
+   *     tags could either be explicitly passed to the method, or implicitly
+   *     read from current execution context.
+   * @param attachments optional The contextual information associated with an
+   *     example value. THe contextual information is represented as key - value
+   *     string pairs.
    */
-  record(measurements: Measurement[], tags?: TagMap): void;
+  record(
+      measurements: Measurement[], tags?: TagMap,
+      attachments?: {[key: string]: string}): void;
 
   /**
    * Remove all registered Views and exporters from the stats.
@@ -181,8 +186,13 @@ export interface View {
    * Measurements with measurement type INT64 will have its value truncated.
    * @param measurement The measurement to record
    * @param tags The tags to which the value is applied
+   * @param attachments optional The contextual information associated with an
+   *     example value. THe contextual information is represented as key - value
+   *     string pairs.
    */
-  recordMeasurement(measurement: Measurement, tags: TagMap): void;
+  recordMeasurement(
+      measurement: Measurement, tags: TagMap,
+      attachments?: {[key: string]: string}): void;
   /**
    * Returns a snapshot of an AggregationData for that tags/labels values.
    * @param tagValues The desired data's tag values.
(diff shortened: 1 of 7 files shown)


packages/opencensus-core/src/stats/types.ts
```diff
@@ -181,8 +186,13 @@ export interface View {
    * Measurements with measurement type INT64 will have its value truncated.
    * @param measurement The measurement to record
    * @param tags The tags to which the value is applied
+   * @param attachments optional The contextual information associated with an
+   *     example value. THe contextual information is represented as key - value
```
draffensperger: Nit typo `The` not `THe`

packages/opencensus-core/src/stats/view.ts
```diff
@@ -114,8 +115,13 @@ export class BaseView implements View {
    * Measurements with measurement type INT64 will have its value truncated.
    * @param measurement The measurement to record
    * @param tags The tags to which the value is applied
+   * @param attachments optional The contextual information associated with an
+   *     example value. THe contextual information is represented as key - value
```
draffensperger: `The` not `THe`

packages/opencensus-core/src/stats/view.ts
```diff
@@ -221,18 +230,40 @@ export class BaseView implements View {
...
+        if (statsExemplar) {
+          // Bucket with an Exemplar.
+          metricBucket = {
+            count: bucketCount,
+            exemplar: {
+              value: statsExemplar.value,
+              timestamp: timestampFromMillis(statsExemplar.timestamp),
+              attachments: statsExemplar.attachments
+            }
+          } as metricBucket;
```
draffensperger: Optional: rather than use a type cast here with a `let` above, could you make a helper function say `getMetricBucket` that returns a `metricBucket` type and then you could even just call `buckets.push(getMetricBucket(statsExemplar, bucketCount))` with no intermediate variable?

packages/opencensus-core/src/stats/view.ts
```diff
@@ -221,18 +230,40 @@ export class BaseView implements View {
...
     let value;
-
     if (data.type === AggregationType.DISTRIBUTION) {
-      // TODO: Add examplar transition
-      const {count, sum, sumOfSquaredDeviation} = data;
+      const {count, sum, sumOfSquaredDeviation, exemplars} = data;
+      const buckets = [];
+      for (let bucket = 0; bucket < data.bucketCounts.length; bucket++) {
+        const bucketCount = data.bucketCounts[bucket];
+        let statsExemplar;
```
draffensperger: Rather than a `let`, what would you think about using `const` and a ternary expression, so:
`const statsExemplar = exmplars ? exemplars[bucket] : undefined;`?

packages/opencensus-core/test/test-recorder.ts
```diff
@@ -176,6 +176,41 @@ describe('Recorder', () => {
           }
         });
 
+    describe(`for distribution aggregation data with attachments`, () => {
```
draffensperger: Nit: I tend to only use template backticks when something needs to be interpolated

packages/opencensus-core/src/stats/view.ts
mayurkale22: Good suggestion, done.

packages/opencensus-core/src/stats/view.ts
mayurkale22: Accepted.

packages/opencensus-core/test/test-recorder.ts
mayurkale22: This is due to copy-paste from previous tests, fixed.

packages/opencensus-core/src/stats/types.ts
mayurkale22: Done, thanks.

mayurkale22:
> Very cool! LGTM with some nits. Will there be a follow up to automatically add metric exmplars for things like traced HTTP/gRPC calls in those metrics?

Yes, will create an issue to track this.

PR #406, opened 2019-03-08
Instrumentation/HTTP2/HTTPS: Enforce strictNullChecks and noUnusedLocals

mayurkale22:
```strictNullChecks``` will help us to make the code safer (referencing nulls or undefined values), more robust and reduce some boilerplate.

This is part of #348

### DIFF
--- a/packages/opencensus-instrumentation-http2/test/test-http2.ts
+++ b/packages/opencensus-instrumentation-http2/test/test-http2.ts
@@ -15,13 +15,10 @@
  */
 
 
-import {CoreTracer, RootSpan, Span, SpanEventListener, TracerConfig} from '@opencensus/core';
-import {logger} from '@opencensus/core';
+import {CoreTracer, logger, RootSpan, Span, SpanEventListener} from '@opencensus/core';
 import * as assert from 'assert';
 import * as http2 from 'http2';
-import * as mocha from 'mocha';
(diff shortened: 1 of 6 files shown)

### CODE REVIEW

File: packages/opencensus-instrumentation-http2/test/test-http2.ts [hunk 1]
```diff
@@ -100,12 +99,14 @@ describe('Http2Plugin', () => {
...
+    plugin.enable(http2, tracer, VERSION, {}, '');
     server = http2.createServer();
     server.on('stream', (stream, requestHeaders) => {
-      const statusCode = requestHeaders[':path'].length > 1 ?
-          +requestHeaders[':path'].slice(1) :
-          200;
+      const path = requestHeaders[':path'];
+      let statusCode = 200;
+      if (path) {
+        statusCode = path.length > 1 ? +path.slice(1) : 200;
```
draffensperger: Nit: what would you think about using `Number(path.slice(1))` to make the type conversion a bit more explicit?

And do we need to worry about malformed numbers here (that would give `NaN` to the `Number` constructor call)? If so, I'd say we can do in a follow up PR, but figured I'd mention as I'm thinking of it here.

mayurkale22: Good call, added ```isNaN``` check. PTAL

PR #408, opened 2019-03-09
Exporter/Stackdriver: Enforce strictNullChecks and noUnusedLocals

mayurkale22:
This is part of #348


packages/opencensus-exporter-stackdriver/src/stackdriver-monitoring.ts
```diff
@@ -39,8 +39,8 @@ export class StackdriverStatsExporter implements StatsEventListener {
   private projectId: string;
   private metricPrefix: string;
   private displayNamePrefix: string;
-  private onMetricUploadError: (err: Error) => void;
-  private timer: NodeJS.Timer;
+  private onMetricUploadError!: (err: Error) => void;
```
draffensperger: Why is the `!` needed here, could this be `?` instead and then just handle the cases where it's missing?

mayurkale22: Done.

PR #410, opened 2019-03-10
add ioredis instrumentation plugin

vmarchaud:
cc @mayurkale22 @draffensperger

### DIFF
--- a/packages/opencensus-instrumentation-ioredis/README.md
+++ b/packages/opencensus-instrumentation-ioredis/README.md
@@ -1,0 +1,17 @@
+# OpenCensus ioredis Instrumentation for Node.js
+[![Gitter chat][gitter-image]][gitter-url]
+
+OpenCensus IORedis Instrumentation allows the user to automatically collect trace data.
+This plugin only support the `ioredis` module starting from the version 2 to latest.
+
+This project is still at an early stage of development. It's subject to change.
+
+## Useful links
+
+- For more information on OpenCensus, visit: <https://opencensus.io/>
+- To checkout the OpenCensus for Node.js, visit: <https://github.com/census-instrumentation/opencensus-node>
+- For help or feedback on this project, join us on [gitter](https://gitter.im/census-instrumentation/Lobby)
+- IORedis repository: https://github.com/luin/ioredis
+
+[gitter-image]: https://badges.gitter.im/census-instrumentation/lobby.svg
+[gitter-url]: https://gitter.im/census-instrumentation/lobby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge
--- a/packages/opencensus-instrumentation-ioredis/package.json
+++ b/packages/opencensus-instrumentation-ioredis/package.json
@@ -1,0 +1,64 @@
+{
+  "name": "@opencensus/instrumentation-ioredis",
+  "version": "0.0.1",
+  "description": "Opencensus IORedis automatic instrumentation package.",
+  "main": "build/src/index.js",
+  "types": "build/src/index.d.ts",
+  "repository": "census-instrumentation/opencensus-node",
+  "scripts": {
+    "test": "nyc mocha build/test/**/*.js",
+    "codecov": "nyc report --reporter=json && codecov -f coverage/*.json",
+    "clean": "rimraf build/*",
+    "check": "gts check",
+    "compile": "tsc -p .",
+    "fix": "gts fix",
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "posttest": "npm run check"
+  },
+  "keywords": [
+    "opencensus",
+    "nodejs",
+    "tracing",
+    "profiling",
+    "stats",
+    "metrics"
+  ],
+  "author": "Google Inc.",
+  "license": "Apache-2.0",
+  "engines": {
+    "node": ">=6.0"
+  },
+  "files": [
+    "build/src/**/*.js",
+    "build/src/**/*.d.ts",
+    "doc",
+    "CHANGELOG.md",
+    "LICENSE",
+    "README.md"
+  ],
+  "publishConfig": {
+    "access": "public"
+  },
+  "devDependencies": {
+    "@types/mocha": "^5.2.6",
+    "@types/node": "^10.12.12",
+    "@types/semver": "^5.5.0",
+    "@types/shimmer": "^1.0.1",
+    "@types/ioredis": "^4.0.9",
+    "codecov": "^3.1.0",
+    "gts": "^0.9.0",
+    "mocha": "^6.0.0",
+    "ncp": "^2.0.0",
+    "nyc": "^13.0.0",
+    "rimraf": "^2.6.2",
+    "ts-node": "^8.0.0",
+    "typescript": "~3.2.0",
+    "ioredis": "^4.6.2"
+  },
+  "dependencies": {
+    "@opencensus/core": "^0.0.9",
+    "semver": "^5.5.0",
+    "shimmer": "^1.2.0"
+  }
+}
--- a/packages/opencensus-instrumentation-ioredis/src/ioredis.ts
+++ b/packages/opencensus-instrumentation-ioredis/src/ioredis.ts
@@ -1,0 +1,131 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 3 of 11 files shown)


packages/opencensus-instrumentation-ioredis/README.md
```diff
@@ -0,0 +1,15 @@
+# OpenCensus ioredis Instrumentation for Node.js
+[![Gitter chat][gitter-image]][gitter-url]
+
+OpenCensus Redis Instrumentation allows the user to automatically collect trace data.
+This plugin only support the `ioredis` module starting from the version 2 to latest.
+
+This project is still at an early stage of development. It's subject to change.
+
+## Useful links
```
draffensperger: Nit: do you need a blank line after this heading so that the bulletet list is formatted correctly?

packages/opencensus-instrumentation-ioredis/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "fix": "gts fix",
+    "prepare": "npm run compile",
+    "pretest": "npm run compile",
+    "posttest": "npm run check"
+  },
+  "keywords": [
+    "opencensus",
+    "nodejs",
+    "tracing",
+    "profiling"
```
draffensperger: Could you add `"stats"` and `"metrics"`?

packages/opencensus-instrumentation-ioredis/src/ioredis.ts
```diff
@@ -0,0 +1,132 @@
...
+ * limitations under the License.
+ */
+
+import {BasePlugin, CanonicalCode, Span, SpanKind} from '@opencensus/core';
+import * as ioredis from 'ioredis';
+import * as semver from 'semver';
+import * as shimmer from 'shimmer';
+
+export type IORedisCommand = {
+  reject: (err: Error) => void|undefined;
```
draffensperger: Nit, I think returning `undefined` is essentially the same as `void`, so what would you think about justing using `=> void` here and for `resolve` below?

packages/opencensus-instrumentation-ioredis/test/test-ioredis.ts
```diff
@@ -0,0 +1,202 @@
...
+  }
+  onEndSpan(root: RootSpan) {
+    this.endedRootSpans.push(root);
+  }
+}
+
+/**
+ * Asserts root spans attributes.
+ * @param rootSpanVerifier An instance of rootSpanVerifier to analyse RootSpan
+ * instances from.
```
draffensperger: Nit: could you indent `instances from.` by 4 spaces to make it visually clear that it's a continuation of the JS doc `@param` tag?

packages/opencensus-instrumentation-ioredis/test/test-ioredis.ts
```diff
@@ -0,0 +1,202 @@
...
+    await client.del('hash');
+  });
+
+  after(() => {
+    if (client) {
+      client.quit();
+    }
+  });
+
+  /** Should intercept query */
```
draffensperger: Optional: can this comment be removed since it's implied by the `describe` and `it` messages below? If not, could they be rewritten in a such a way that it could be removed?

packages/opencensus-instrumentation-ioredis/test/test-ioredis.ts
```diff
@@ -0,0 +1,202 @@
...
+            assertSpan(
+                rootSpanVerifier, `redis-del`, SpanKind.CLIENT, (span) => {
+                  return span.attributes.arguments === undefined;
+                });
+            done();
+          });
+    });
+  });
+
+  /** Should intercept command */
```
draffensperger: Similar comment: can this be removed?

packages/opencensus-instrumentation-ioredis/README.md
mayurkale22: Can you add ioredis link here ?

packages/opencensus-instrumentation-ioredis/README.md
```diff
@@ -0,0 +1,15 @@
+# OpenCensus ioredis Instrumentation for Node.js
+[![Gitter chat][gitter-image]][gitter-url]
+
+OpenCensus Redis Instrumentation allows the user to automatically collect trace data.
```
mayurkale22: s/Redis/ioredis ?

packages/opencensus-instrumentation-ioredis/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "codecov": "^3.1.0",
+    "gts": "^0.9.0",
+    "mocha": "^6.0.0",
+    "ncp": "^2.0.0",
+    "nock": "^10.0.0",
+    "nyc": "^13.0.0",
+    "rimraf": "^2.6.2",
+    "ts-node": "^8.0.0",
+    "typescript": "~3.2.0",
+    "@types/ioredis": "^4.0.9",
```
mayurkale22: Can you move @types on top, we try to list dependencies in sorted order. Also looks like @types/uuid is not required here.

packages/opencensus-instrumentation-ioredis/src/ioredis.ts
```diff
@@ -0,0 +1,132 @@
...
+  resolve: (result: {}) => void | undefined;
+  promise: Promise<{}>;
+  args: Array<string|Buffer|number>;
+  callback: Function | undefined;
+  name: string;
+};
+
+/** Redis instrumentation plugin for Opencensus */
```
mayurkale22: Rename Redis to IORedis everywhere.

packages/opencensus-instrumentation-ioredis/package.json
```diff
@@ -0,0 +1,65 @@
...
+    "@types/nock": "^9.1.3",
+    "@types/node": "^10.12.12",
+    "@types/semver": "^5.5.0",
+    "@types/shimmer": "^1.0.1",
+    "@types/uuid": "^3.4.3",
+    "codecov": "^3.1.0",
+    "gts": "^0.9.0",
+    "mocha": "^6.0.0",
+    "ncp": "^2.0.0",
+    "nock": "^10.0.0",
```
mayurkale22: Looks like ```nock``` and ```@types/nock``` are not needed here. If so, please remove it.

vmarchaud:
I believe i've addressed every issue

packages/opencensus-instrumentation-ioredis/src/ioredis.ts
```diff
@@ -0,0 +1,131 @@
...
+    super(moduleName);
+  }
+
+  /**
+   * Patches IORedis operations.
+   */
+  protected applyPatch() {
+    this.logger.debug('Patched ioredis');
+
+    if (!semver.satisfies(this.version, '>=2.0.0 <5.0.0')) {
```
mayurkale22: Can you add supported version in ReadMe?

packages/opencensus-instrumentation-ioredis/README.md
```diff
@@ -0,0 +1,17 @@
+# OpenCensus ioredis Instrumentation for Node.js
```
mayurkale22: Optional : ```s/ioredis/IORedis?```

packages/opencensus-instrumentation-ioredis/test/test-ioredis.ts
```diff
@@ -0,0 +1,200 @@
+/**
+ * Copyright 2018, OpenCensus Authors
```
mayurkale22: ```s/Copyright 2018/Copyright 2019```

packages/opencensus-instrumentation-ioredis/test/test-ioredis.ts
```diff
@@ -0,0 +1,200 @@
...
+
+  const tracer = new CoreTracer();
+  const rootSpanVerifier = new RootSpanVerifier();
+  let client: ioredis.Redis;
+
+  before((done) => {
+    tracer.start({samplingRate: 1, logger: logger.logger(1)});
+    tracer.registerSpanEventListener(rootSpanVerifier);
+    plugin.enable(ioredis, tracer, VERSION, {}, '');
+    client = new ioredis(URL);
```
mayurkale22: If redis is not installed locally, do you think this will work?

packages/opencensus-instrumentation-ioredis/src/ioredis.ts
vmarchaud: Already the case, i've added a max range to avoid breakage on the next major but i will remove it.

packages/opencensus-instrumentation-ioredis/test/test-ioredis.ts
vmarchaud: No, it will not work but it's a test file so i would think that it's normal to install it as a dependency to be able to run the test.

vmarchaud:
@mayurkale22 Are you sure to add them in the global readme even though they are not currently enabled with the latest version ? We would need a separate PR to add them into `instrumentation-all` package so i believe this change should be done in the separate PR

mayurkale22:
> @mayurkale22 Are you sure to add them in the global readme even though they are not currently enabled with the latest version ? We would need a separate PR to add them into `instrumentation-all` package so i believe this change should be done in the separate PR

Makes sense 👍

mayurkale22:
Thanks for the contribution again!

PR #418, opened 2019-03-12
Stats: Record against implicit (current) tag context

mayurkale22:
Specs are available here: https://github.com/census-instrumentation/opencensus-specs/blob/master/stats/Record.md#recording-stats

This will allow users to set and get the current tag context, which can be used when recoding the stats if explicit tags are not provided.

### DIFF
--- a/packages/opencensus-core/src/stats/stats.ts
+++ b/packages/opencensus-core/src/stats/stats.ts
@@ -20,7 +20,7 @@ import {StatsEventListener} from '../exporters/types';
 import {Metric} from '../metrics/export/types';
 import {Metrics} from '../metrics/metrics';
 import {TagMap} from '../tags/tag-map';
-import {getCurrentTagMap} from '../tags/tagger';
+import {getCurrentTagContext} from '../tags/tagger';
 import {TagKey} from '../tags/types';
 import {MetricProducerForStats} from './metric-producer';
 import {AggregationType, Measure, Measurement, MeasureType, MeasureUnit, Stats, View} from './types';
@@ -189,7 +189,7 @@ export class BaseStats implements Stats {
 
     if (!tags) {
       // Record against implicit (current) context
-      tags = getCurrentTagMap();
+      tags = getCurrentTagContext();
     }
 
     for (const measurement of measurements) {
--- a/packages/opencensus-core/src/stats/stats.ts
+++ b/packages/opencensus-core/src/stats/stats.ts
@@ -20,8 +20,8 @@ import {StatsEventListener} from '../exporters/types';
 import {Metric} from '../metrics/export/types';
 import {Metrics} from '../metrics/metrics';
 import {TagMap} from '../tags/tag-map';
+import {getCurrentTagMap} from '../tags/tagger';
 import {TagKey} from '../tags/types';
-
 import {MetricProducerForStats} from './metric-producer';
 import {AggregationType, Measure, Measurement, MeasureType, MeasureUnit, Stats, View} from './types';
 import {BaseView} from './view';
@@ -80,8 +80,8 @@ export class BaseStats implements Stats {
    * @param aggregation The view aggregation type
    * @param tagKeys The view columns (tag keys)
    * @param description The view description
-   * @param bucketBoundaries The view bucket boundaries for a distribution
-   * aggregation type
+   * @param bucketBoundaries optional The view bucket boundaries for a
+   *     distribution aggregation type
    */
   createView(
       name: string, measure: Measure, aggregation: AggregationType,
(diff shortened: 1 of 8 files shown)


packages/opencensus-core/src/stats/stats.ts
```diff
@@ -188,8 +188,8 @@ export class BaseStats implements Stats {
     }
 
     if (!tags) {
-      // TODO(mayurkale): read tags current execution context
-      tags = new TagMap();
+      // Record against implicit (current) context
+      tags = getCurrentTagMap();
```
draffensperger: Will users be able to have fine-grained control over which tags for a given metric come from the context and which should be explicitly specified? Or over which tags in the context are associated with which metric?

songy23: No we intentionally don't provide that level of control. Tags in the context are treated as "opaque" to prevent users accidentally leak confidential information. "Opaque" here means users can insert or upsert tags into context, but they cannot see what tags are in current context or iterate over them.

draffensperger: OK. Is it possible to have a case where a metric does not have labels associated with all the tags in the context?

mayurkale22: Yes, it is possible. This is the logic to handle that ->  https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/stats/recorder.ts#L47

draffensperger: Makes sense, thanks

packages/opencensus-core/test/test-stats.ts
```diff
@@ -225,5 +226,40 @@ describe('Stats', () => {
...
+      const measurement = {measure, value: 1};
+      tagger.withTagContext(tags, () => {
+        globalStats.record([measurement]);
+      });
+
+      assert.strictEqual(testExporter.recordedMeasurements.length, 1);
+      assert.deepEqual(testExporter.recordedMeasurements[0], measurement);
+      aggregationData =
+          testExporter.registeredViews[0].getSnapshot(
+              [{value: 'value1'}, {value: 'value2'}]) as LastValueData;
```
draffensperger: What's the role of the `as LastValueData`, since it looks like `aggregationData` already has that type? Can that safely be removed here and below?

mayurkale22: The return type of ```getSnapshot()``` method is ```AggregationData```, which is combination of four types ```SumData|CountData|LastValueData|DistributionData```. The conversion (```as LastValueData```) is added to get exact type of ```AggregationData```.

draffensperger: Make sense, thanks!

packages/opencensus-core/src/tags/tag-map.ts
```diff
@@ -31,6 +31,14 @@ export class TagMap {
...
       throw new Error(`Invalid TagValue: ${tagValue.value}`);
     }
+    let existingKey;
+    for (const key of this.registeredTags.keys()) {
+      if (key.name === tagKey.name) {
+        existingKey = key;
+        break;
+      }
+    }
+    if (existingKey) this.registeredTags.delete(existingKey);
```
songy23: Just curious, in Node if I want to update the value for a key in a map, I need to delete the original <key, value> first?

packages/opencensus-core/test/test-tagger.ts
```diff
@@ -0,0 +1,100 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import * as assert from 'assert';
+import {TagMap} from '../src';
+import * as tagger from '../src/tags/tagger';
+
+describe('tagger()', () => {
```
songy23: Consider adding a test on `clear`.

mayurkale22: done, thanks

packages/opencensus-core/src/tags/tag-map.ts
mayurkale22: As per my understanding, this is true if we use ```object``` as a key for ```map```. The maps detect existence ```( Map.prototype.has ( key ))``` of keys based on equality (===). Primitives like strings and numbers are compared by their value, while objects (arrays, plain objects) are compared by their reference. That comparison by reference basically checks to see if the objects given refer to the same location in memory. 

Basically to check the objects' "**value equality**" we essentially have to iterate over every property in the objects to see whether they are equal.

A potential option is to use ```tagkey.name``` as a key which is a string type. This required lot more changes and would like to handle in separate PR if we really want to do it.

songy23: Make sense, thanks for the explanation. I'd prefer to keep using the `TagKey` object as the map key.

Issue #340, opened 2019-02-12
B3 Propagation UX issues


kjin:
Working on support for propagation via OpenCensus propagation modules in Stackdriver Trace (https://github.com/googleapis/cloud-trace-nodejs/pull/950), I ran into a couple of issues:

* Because `strictNullChecks` is turned off, `B3Format` does an unsafe non-null cast ([source](https://github.com/census-instrumentation/opencensus-node/blob/38ed0a2dde9c927054c7e392ae7012b89b921577/packages/opencensus-propagation-b3/src/b3-format.ts)). The result is that consumers of this propagation library will get invalid values whenever the B3 header is not set, and crash if they depended on these values being valid (which is a reasonable assumption for TS users).
* Not a behavioral bug but there are a few superfluous falsiness checks in `B3Format` such as for `getter` and `setter`.
* The B3 propagation module exports a different interface than the Stackdriver propagation module. Stackdriver propagation exports an implementation of `Propagation`, while B3 propagation exports a class constructor. I believe they should all export the former rather than the latter.

These issues may also exist for other propagation mechanisms, I only looked at B3 in particular.

mayurkale22:
@kjin Thanks for filling an issue. B3 and TraceContext format are using the common core propagation interface (https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/propagation/types.ts#L36-L40), it makes sense to implement same for the Stackdriver propagation module.

mayurkale22:
@kjin Did you get a chance to look at this? I remember you volunteer to help with this. Let me know if you still have bandwidth to take care of this.

kjin:
I will not have bandwidth to look at this within the next two weeks. But if it can wait I can circle back to it.

PR #421, opened 2019-03-13
Propagation/B3: Enforce strictNullChecks and noUnusedLocals

mayurkale22:
This is part of #348 and #340

### DIFF
--- a/packages/opencensus-propagation-b3/src/b3-format.ts
+++ b/packages/opencensus-propagation-b3/src/b3-format.ts
@@ -20,7 +20,7 @@ import * as uuid from 'uuid';
 
 export const X_B3_TRACE_ID = 'x-b3-traceid';
 export const X_B3_SPAN_ID = 'x-b3-spanid';
-export const X_B3_PARENT_SPAN_ID = 'x-x3-parentspanid';
+export const X_B3_PARENT_SPAN_ID = 'x-b3-parentspanid';
 export const X_B3_SAMPLED = 'x-b3-sampled';
 export const SAMPLED_VALUE = 0x1;
 export const NOT_SAMPLED_VALUE = 0x0;
--- a/packages/opencensus-propagation-b3/src/b3-format.ts
+++ b/packages/opencensus-propagation-b3/src/b3-format.ts
@@ -32,17 +32,13 @@ export class B3Format implements Propagation {
    * in the headers, null is returned.
    * @param getter
    */
-  extract(getter: HeaderGetter): SpanContext|null {
-    if (!getter) return null;
-    let opt = getter.getHeader(X_B3_SAMPLED);
-    if (opt instanceof Array) {
-      opt = opt[0];
-    }
+  extract(getter: HeaderGetter): SpanContext {
+    const opt = this.parseHeader(getter.getHeader(X_B3_SAMPLED));
     return {
-      traceId: getter.getHeader(X_B3_TRACE_ID),
-      spanId: getter.getHeader(X_B3_SPAN_ID),
+      traceId: this.parseHeader(getter.getHeader(X_B3_TRACE_ID)),
+      spanId: this.parseHeader(getter.getHeader(X_B3_SPAN_ID)),
       options: isNaN(Number(opt)) ? NOT_SAMPLED_VALUE : Number(opt)
-    } as SpanContext;
+    };
   }
 
   /**
@@ -54,7 +50,7 @@ export class B3Format implements Propagation {
     setter.setHeader(X_B3_TRACE_ID, spanContext.traceId || '');
     setter.setHeader(X_B3_SPAN_ID, spanContext.spanId || '');
     if (spanContext &&
-        (spanContext.options || NOT_SAMPLED_VALUE & SAMPLED_VALUE) !== 0) {
+        ((spanContext.options || NOT_SAMPLED_VALUE) & SAMPLED_VALUE) !== 0) {
       setter.setHeader(X_B3_SAMPLED, `${SAMPLED_VALUE}`);
     } else {
       setter.setHeader(X_B3_SAMPLED, `${NOT_SAMPLED_VALUE}`);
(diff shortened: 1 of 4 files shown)


packages/opencensus-propagation-b3/src/b3-format.ts
```diff
@@ -36,21 +32,17 @@ export class B3Format implements Propagation {
...
+    let opt = getter.getHeader(X_B3_SAMPLED);
+    if (opt instanceof Array) {
+      opt = opt[0];
     }
-    return null;
+    return {
+      traceId: getter.getHeader(X_B3_TRACE_ID),
+      spanId: getter.getHeader(X_B3_SPAN_ID),
+      options: isNaN(Number(opt)) ? NOT_SAMPLED_VALUE : Number(opt)
+    } as SpanContext;
```
draffensperger: Can the `as SpanContext` be removed since the return type is implied by the function signature?

packages/opencensus-propagation-b3/src/b3-format.ts
```diff
@@ -36,21 +32,17 @@ export class B3Format implements Propagation {
...
-      }
-      const spanContext = {
-        traceId: getter.getHeader(X_B3_TRACE_ID),
-        spanId: getter.getHeader(X_B3_SPAN_ID),
-        options: isNaN(Number(opt)) ? undefined : Number(opt)
-      } as SpanContext;
-
-      return spanContext;
+  extract(getter: HeaderGetter): SpanContext|null {
+    if (!getter) return null;
```
kjin: You removed this check in `inject`, would you still need this one?

packages/opencensus-propagation-b3/src/b3-format.ts
mayurkale22: Done.

packages/opencensus-propagation-b3/src/b3-format.ts
```diff
@@ -59,14 +51,13 @@ export class B3Format implements Propagation {
...
-      setter.setHeader(X_B3_SPAN_ID, spanContext.spanId || 'undefined');
-      if (spanContext && (spanContext.options & SAMPLED_VALUE) !== 0) {
-        setter.setHeader(X_B3_SAMPLED, `${SAMPLED_VALUE}`);
-      } else {
-        setter.setHeader(X_B3_SAMPLED, `${NOT_SAMPLED_VALUE}`);
-      }
+    setter.setHeader(X_B3_TRACE_ID, spanContext.traceId || '');
+    setter.setHeader(X_B3_SPAN_ID, spanContext.spanId || '');
+    if (spanContext &&
+        (spanContext.options || NOT_SAMPLED_VALUE & SAMPLED_VALUE) !== 0) {
```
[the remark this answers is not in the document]
mayurkale22: Sure, done.

packages/opencensus-propagation-b3/src/b3-format.ts
mayurkale22: Done, thanks

packages/opencensus-propagation-b3/src/b3-format.ts
```diff
@@ -37,20 +33,12 @@ export class B3Format implements Propagation {
...
-        spanId: getter.getHeader(X_B3_SPAN_ID),
-        options: isNaN(Number(opt)) ? undefined : Number(opt)
-      } as SpanContext;
-
-      return spanContext;
-    }
-    return null;
+    const opt = this.parseHeader(getter.getHeader(X_B3_SAMPLED));
+    return {
+      traceId: this.parseHeader(getter.getHeader(X_B3_TRACE_ID)),
```
kjin: IIUC the `extract` method should [return `null`](https://github.com/census-instrumentation/opencensus-node/blob/3111ab8c29b5d60e68a762c439ef909e3af9589b/packages/opencensus-core/src/trace/propagation/types.ts#L37) if there is no incoming traceID/spanID, rather than fill an object with empty strings.

packages/opencensus-propagation-b3/src/b3-format.ts
```diff
@@ -59,14 +47,13 @@ export class B3Format implements Propagation {
...
   inject(setter: HeaderSetter, spanContext: SpanContext): void {
-    if (setter) {
-      setter.setHeader(X_B3_TRACE_ID, spanContext.traceId || 'undefined');
-      setter.setHeader(X_B3_SPAN_ID, spanContext.spanId || 'undefined');
-      if (spanContext && (spanContext.options & SAMPLED_VALUE) !== 0) {
-        setter.setHeader(X_B3_SAMPLED, `${SAMPLED_VALUE}`);
-      } else {
-        setter.setHeader(X_B3_SAMPLED, `${NOT_SAMPLED_VALUE}`);
-      }
+    setter.setHeader(X_B3_TRACE_ID, spanContext.traceId || '');
```
kjin: On a similar note as above, I'm not sure if we should be writing an empty value to a header. That being said the [`SpanContext` interface](https://github.com/census-instrumentation/opencensus-node/blob/3111ab8c29b5d60e68a762c439ef909e3af9589b/packages/opencensus-core/src/trace/model/types.ts#L252-L261) should guarantee that `traceId` and `spanId` are not blank.

packages/opencensus-propagation-b3/src/validators.ts
```diff
@@ -0,0 +1,72 @@
+/**
```
mayurkale22: This file is taken from ```opencensus-propagation-tracecontext``` -> https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-propagation-tracecontext/src/validators.ts

packages/opencensus-propagation-b3/src/b3-format.ts
mayurkale22: This is valid point, I have changed the implementation to return null in case of no incoming ```traceID/spanID```.

packages/opencensus-propagation-b3/src/b3-format.ts
mayurkale22: Agreed, does not make sense to write empty values to a header. Added check for the same. PTAL

mayurkale22:
@kjin Please take a look again once you have time.

PR #422, opened 2019-03-13
Add package-lock.json to .gitignore

jinmel:


draffensperger:
Could you explain the motivation for this?

My understanding was that package locks should be committed to ensure reproducible builds. See also this [StackOverflow question](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5) about it.

jinmel:
@draffensperger then we should commit the result after npm install on root package shouldn't we?

mayurkale22:
> Could you explain the motivation for this?
> 
> My understanding was that package locks should be committed to ensure reproducible builds. See also this [StackOverflow question](https://stackoverflow.com/questions/44206782/do-i-commit-the-package-lock-json-file-created-by-npm-5) about it.

+1. I think package-lock is required to ensure a consistent install and compatible dependencies, especially on shared projects like this one.

jinmel:
@mayurkale22 is your opinion on saving package-lock.json to repo?

mayurkale22:
> @mayurkale22 is your opinion on saving package-lock.json to repo?

I have already added my opinion here https://github.com/census-instrumentation/opencensus-node/pull/422#issuecomment-472560143. I think we should continue to commit package-lock to repo.

jinmel:
understood. closing this PR

PR #436, opened 2019-03-20
Make Stackdriver propagation module consistent with other propagations interface

### LINKED ISSUE #340: B3 Propagation UX issues

mayurkale22:
This PR fixes two things:

1. Stackdriver propagation exports an implementation of Propagation, while `B3`, `Binary Format`, `TraceContext` and `Jaeger` propagation exports a class constructor. Fixes #340
2. Enforce `strictNullChecks` and `noUnusedLocals` on Stackdriver propagation package. This is part of #348

### DIFF
--- a/packages/opencensus-propagation-stackdriver/src/stackdriver-format.ts
+++ b/packages/opencensus-propagation-stackdriver/src/stackdriver-format.ts
@@ -1,0 +1,103 @@
+/**
+ * Copyright 2019 OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This file implements propagation for the Stackdriver Trace v1 Trace Context
+ * format.
+ *
+ * The header specification is:
+ * "X-Cloud-Trace-Context: TRACE_ID/SPAN_ID;o=TRACE_TRUE"
+ * Where:
+ *  {TRACE_ID} is a 32-character hexadecimal value representing a 128-bit
+ *    number. It should be unique between your requests, unless you
+ *    intentionally want to bundle the requests together.
+ *  {SPAN_ID} is the decimal representation of the (unsigned) span ID. It
+ *    should be 0 for the first span in your trace. For subsequent requests,
+ *    set SPAN_ID to the span ID of the parent request.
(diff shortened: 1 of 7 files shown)


packages/opencensus-propagation-stackdriver/test/test-stackdriver-format.ts
```diff
@@ -0,0 +1,145 @@
...
+import {StackdriverFormat, TRACE_CONTEXT_HEADER_NAME} from '../src';
+
+const stackdriverFormat = new StackdriverFormat();
+
+describe('StackdriverPropagation', () => {
+  describe('extract()', () => {
+    it('should extract context of a sampled span from headers', () => {
+      const headers: {[key: string]: string|string[]|undefined} = {};
+      headers[TRACE_CONTEXT_HEADER_NAME] = '123456/667;o=1';
+      const getter: HeaderGetter = {
```
draffensperger: Would it make sense to abstract this into a helper function to share it between these different test cases?

packages/opencensus-propagation-stackdriver/src/stackdriver-format.ts
```diff
@@ -0,0 +1,103 @@
...
+      header += `;o=${spanContext.options}`;
+    }
+
+    setter.setHeader(TRACE_CONTEXT_HEADER_NAME, header);
+  }
+
+  /** Generate SpanContexts */
+  generate(): SpanContext {
+    return {
+      traceId: uuid.v4().split('-').join(''),
```
draffensperger: Optional: would it make sense to do `uuid.v4().replace('-', '')`

mayurkale22: SG, done. As per https://jsfiddle.net/basickarl/re5ywuum/ test, fastest is replace.

packages/opencensus-propagation-stackdriver/test/test-stackdriver-format.ts
mayurkale22: Sure, done

mayurkale22:
@kjin As this fixes your issue #340, It will be nice to have reviews from you.

mayurkale22:
> LGTM, in the future it would be great if the CI shows green before requesting reviews.

Sure, btw current build failure wasn't related to this change.

PR #437, opened 2019-03-20
Add support for nested spans

### LINKED ISSUE #105: Cannot create nested Child Spans

kenashcraft:
`rootSpan.startChildSpan()` takes an options object as input where one of the fields is `parentSpanId`.  However, `parentSpanId` isn't wired up.  This PR does that wiring and adds tests for both the core library as well as the Zipkin exporter.

Fixes #105

### DIFF
--- a/packages/opencensus-core/test/test-root-span.ts
+++ b/packages/opencensus-core/test/test-root-span.ts
@@ -77,6 +77,28 @@ describe('RootSpan', () => {
     });
   });
 
+  describe('nested spans', () => {
+    it('should get nested spans from rootspan instance', () => {
+      const root = new RootSpan(tracer, name, kind, traceId, parentSpanId);
+      root.start();
+      assert.equal(root.numberOfChildren, 0);
+      const child =
+          root.startChildSpan('childName', types.SpanKind.UNSPECIFIED);
+      assert.equal(root.numberOfChildren, 1);
+      const grandchild = root.startChildSpan({
+        name: 'grandchildName',
+        kind: types.SpanKind.UNSPECIFIED,
+        parentSpanId: child.id
+      });
+      assert.equal(root.numberOfChildren, 2);
+      assert.equal(grandchild.parentSpanId, child.id);
+
+      assert.equal(root.spans.length, 2);
+      assert.equal(child, root.spans[0]);
+      assert.equal(grandchild, root.spans[1]);
+    });
+  });
+
   /**
    * Should get trace id from rootspan instance
    */
--- a/packages/opencensus-core/src/trace/model/root-span.ts
+++ b/packages/opencensus-core/src/trace/model/root-span.ts
@@ -138,11 +138,17 @@ export class RootSpan extends SpanBase implements types.RootSpan {
         typeof nameOrOptions === 'object' ? nameOrOptions.name : nameOrOptions;
     const spanKind =
         typeof nameOrOptions === 'object' ? nameOrOptions.kind : kind;
+    const spanParentId = typeof nameOrOptions === 'object' ?
+        nameOrOptions.parentSpanId :
+        undefined;
     if (spanName) {
       child.name = spanName;
     }
     if (spanKind) {
       child.kind = spanKind;
+    }
+    if (spanParentId) {
+      child.parentSpanId = spanParentId;
     }
 
     child.start();
--- a/packages/opencensus-exporter-zipkin/src/zipkin.ts
+++ b/packages/opencensus-exporter-zipkin/src/zipkin.ts
@@ -166,7 +166,7 @@ export class ZipkinTraceExporter implements Exporter {
       timestamp: span.startTime.getTime() * MICROS_PER_MILLI,
       duration: Math.round(span.duration * MICROS_PER_MILLI),
       debug: true,
-      shared: true,
+      shared: span.isRootSpan, // Only the root span could be shared.
       localEndpoint: {serviceName: this.serviceName},
       tags: this.createTags(span.attributes, span.status),
       annotations: this.createAnnotations(span.annotations, span.messageEvents)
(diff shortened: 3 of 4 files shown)


kenashcraft:
I'm changing `opencensus-core` and `opencensus-exporter-zipkin` simultaneously.  `opencensus-exporter-zipkin` specifies a published version of `opencensus-core`.  Is this expected to work or should it be split into two PRs?

packages/opencensus-core/test/test-root-span.ts
```diff
@@ -77,6 +77,28 @@ describe('RootSpan', () => {
...
   });
 
+  describe('nested spans', () => {
+    it('should get nested spans from rootspan instance', () => {
+      const root = new RootSpan(tracer, name, kind, traceId, parentSpanId);
+      root.start();
+      assert.equal(root.numberOfChildren, 0);
+      const child =
+          root.startChildSpan('childName', types.SpanKind.UNSPECIFIED);
+      assert.equal(root.numberOfChildren, 1);
```
draffensperger: My understanding is that `numberOfChildren` is supposed to measure the number of *direct* children, which enables checks for trace completeness (if you all spans have child spans for all their children then you know the whole trace has been ingested). But making a span a parent of something other than the root breaks that contract.

My [suggestion for how to fix that](https://github.com/census-instrumentation/opencensus-node/issues/105#issuecomment-468304948) would require some additional changes to make the `Span` and `RootSpan` classes be more similar.

But we might be able to do something simpler in the short run - what if rather than passing in a parent span ID, it passed in a parent span, which would then have its child count incremented?

draffensperger:
Thanks for the contribution BTW! I have been thinking about this type of functionality and see https://github.com/census-instrumentation/opencensus-node/issues/105 for more

packages/opencensus-core/test/test-root-span.ts
kenashcraft: Ah, maybe that is why the Zipkin UI still shows the grandchildren as cascading rather than nesting.

draffensperger: Possibly (though that would somewhat surprise me), is there a way you can look at the raw trace to make sure all the parent links seem correct?

kenashcraft: I found that the issue was that `shared` needs to be false for all spans other than the root span:
https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-exporter-zipkin/src/zipkin.ts#L169

I'll include this in my PR.

kenashcraft: @draffensperger, you suggested changing the child count of the appropriate parent. Currently, only `RootSpan` has `numberOfChildren`. Are you suggesting that `numberOfChildren` move to `SpanBase`?
Also, what should happen when a non-root span ends? Should all of its children be truncated like is done in `RootSpan.end()`? That would mean that `spansLocal` would also move to `SpanBase`.

draffensperger: Yes, I believe moving `numberOfChildren` to `SpanBase` makes sense. I do think that moving the `spansLocal` to `SpanBase` also makes sense in addition to some other refactors I describe in https://github.com/census-instrumentation/opencensus-node/issues/105#issuecomment-468304948

mayurkale22: > I found that the issue was that shared needs to be false for all spans other than the root span.

I think `shared` needs to be `True` if span started by another tracer (on a different host). What's the plan for handling this scenerio, if we change to `False`. Also, it would be nice to have a seperate PR to handle this change.

mayurkale22: > Possibly (though that would somewhat surprise me), is there a way you can look at the raw trace to make sure all the parent links seem correct?

I think it is possible with [console](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/exporters/console-exporter.ts#L63) exporter, need to add parentId in the result.

kenashcraft: @mayurkale22 The Zipkin UI has a button you can click that provides the JSON.  Super easy.

kenashcraft: @draffensperger I sketched out the consolidation of Span and RootSpan in #441.  Please let me know if that is on the right track.

PR #438, opened 2019-03-20
Remove deprecated @types/pkg-dir and pkg-dir 

mayurkale22:
The `@types/pkg-dir` package has been deprecated. https://www.npmjs.com/package/@types/pkg-dir


mayurkale22:
@draffensperger I have removed dependency on `pkg-dir` package also, as it is not supported for node 6 -> https://github.com/sindresorhus/pkg-dir/blob/master/package.json#L13

Once you have time, please take a look again. Thanks.

PR #443, opened 2019-03-25
Add support for gRPC tags propagation

### LINKED ISSUE #325: gRPC tag propagation
Like the span context, we need to be able to propagate the TagMap along with the request. We need to do this for both gRPC and HTTP.

#326 added binary serialization for the tag map. The last step is to update the gRPC client interceptor to read the tag map from the gRPC metadata and insert it into the OC execution context and update the gRPC server interceptor to attach the tag map to gRPC responses as metadata. This is similar to the way the gRPC client and server handle the trace header now.

See also binary serialization and gRPC propagation in the [java client](https://github.com/census-instrumentation/opencensus-java/pull/511) and [go client](https://github.com/census-instrumentation/opencensus-go/blob/7a32882185281dc6ea972e9b80d6a58448359977/tag/map_codec.go).

mayurkale22:
Fixes #325

This PR contains the following changes
1. Share same `contextManager` across tracer and stats singleton instances.
2. Add `withTagContext` and `getCurrentTagContext ` API on stats, so that plugins can use these API to get current `TagContext/TagMap`.
3. Make a deep copy of `TagMap` during get/set into `contextManager`.
4. Insert/Get gRPC metadata with `grpc-tags-bin` header with serialized `TagMap`.
5. Update `tag-context-example.js` example.

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -560,8 +560,8 @@ export class GrpcPlugin extends BasePlugin {
       if (!tags) return null;
       return tags;
     } catch (ignore) {
+      return null;
     }
-    return null;
   }
 
   /**
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -14,18 +14,22 @@
  * limitations under the License.
  */
 
-import {BasePlugin, CanonicalCode, PluginInternalFiles, RootSpan, Span, SpanContext, SpanKind, TagMap, TagTtl, TraceOptions} from '@opencensus/core';
+import {BasePlugin, CanonicalCode, deserializeBinary, PluginInternalFiles, RootSpan, serializeBinary, Span, SpanContext, SpanKind, TagMap, TagTtl, TraceOptions} from '@opencensus/core';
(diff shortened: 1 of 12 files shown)


packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -533,6 +545,37 @@ export class GrpcPlugin extends BasePlugin {
...
+    // Entry doesn't exist.
+    if (!metadataValue) return null;
+    try {
+      const tags = deserializeBinary(metadataValue);
+      // Value is malformed.
+      if (!tags) return null;
+      return tags;
+    } catch (ignore) {
+    }
+    return null;
```
draffensperger: Nit: would it make sense for the `return null;` to be above within the `catch` block?

mayurkale22: Sure, done

mayurkale22:
> Cool! It looks like the `grpc-tags-bin` header is only being serialized/deserialized in the `getTagContext`/`setTagContext` helper functions. Are those called automatically when gRPC calls happen? (Or will that be in a follow-up CL?)

I have added hook to use these functions during gRPC [incoming request](https://github.com/census-instrumentation/opencensus-node/pull/443/files#diff-0bce0feb76aa55b365563773f9016fc1R390) and [outgoing request ](https://github.com/census-instrumentation/opencensus-node/pull/443/files#diff-45fbd1f191ba8f9e49f23b7de8feb559R750).

PR #444, opened 2019-03-25
Use ts-mocha to run tests

kenashcraft:
This gives good stack traces during tests and means we can skip the compile step.  Converts all packages except:

- opencensus-nodejs: has tests that expect .js files/syntax
- opencensus-instrumentation-redis: I couldn't figure out what was going wrong
- opencensus-exporter-ocagent: Depends on the compilation step to copy files around


mayurkale22:
@kenashcraft thanks for the contribution. Please rebase the PR, this should fix the build. We have already removed the dependencies on ` @types/pkg-dir` and `pkg-dir` https://github.com/census-instrumentation/opencensus-node/pull/438.

packages/opencensus-core/package.json
```diff
@@ -6,15 +6,14 @@
...
-    "test": "nyc mocha build/test/**/*.js",
+    "test": "nyc ts-mocha -p ./tsconfig.json test/**/*.ts",
     "codecov": "nyc report --reporter=json && codecov -f coverage/*.json",
     "clean": "rimraf build/*",
     "check": "gts check",
     "compile": "tsc -p .",
     "compile:release": "tsc -p tsconfig-release.json",
     "fix": "gts fix",
     "prepare": "npm run compile:release",
-    "pretest": "npm run compile",
```
mayurkale22: I think we should still keep this and add into unit test workflow. Something like below in `.circleci/config.yml` WDYT?
```
   - run:
        name: Compile code
        command: npm run compile
```

kenashcraft: It looks like that config is for the entire repo.  That would cause all of the other packages to run the compile step twice.  Is that ok and/or should I change all of the other packages to use ts-mocha?

mayurkale22: I would prefer to change all packages to use `ts-mocha`. Also, does it make sense to remove dependencies on `mocha` and `@types/mocha` after this? @draffensperger any thoughts?

kenashcraft: No, we can't remove the dependencies.  The ts-mocha page says we need them: https://www.npmjs.com/package/ts-mocha

mayurkale22: Ok SG.

kenashcraft:
@mayurkale22 Looks like I finally got all of the tests passing.  I did my best to ensure I was changing only the files that needed to be changed, but please double check before we merge.

mayurkale22:
Can you please create an issue to track remaining packages.

kenashcraft:
Added issue #447.

packages/opencensus-core/package.json
```diff
@@ -38,6 +37,16 @@
     "LICENSE",
     "README.md"
   ],
+  "nyc": {
```
mayurkale22: Optional: Can we move these configuration options in `.nycrc`, later on we can add coverage threshold and other options. Also this will make `package.json` more cleaner. WDYT?

kenashcraft: The official docs didn't offer that as an option: https://istanbul.js.org/docs/tutorials/typescript/ . I don't know if it is supported.

mayurkale22: Ok. I referred this link https://github.com/istanbuljs/nyc#configuring-nyc

Issue #297, opened 2019-01-17
Tags: Support HTTP propagation.


mayurkale22:
1. Agree on the text format for tag HTTP propagation.
2. Add HTTP plugin for tags.

PR #445, opened 2019-03-25
Add HTTP/W3C text format serializer to Tag propagation component

mayurkale22:
Updates #297  This is the HTTP version of of #431.

Based on **W3C correlation context format**: https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md

There will be another PR to use these serializer functions in the HTTP plugin to propagate tags on wire.

### DIFF
--- a/packages/opencensus-core/src/tags/propagation/text-format.ts
+++ b/packages/opencensus-core/src/tags/propagation/text-format.ts
@@ -1,0 +1,97 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


packages/opencensus-core/src/tags/propagation/text-format.ts
```diff
@@ -0,0 +1,97 @@
...
+ */
+
+/**
+ * This module contains the functions for serializing and deserializing
+ * TagMap (TagContext) with W3C Correlation Context as the HTTP text format.
+ * It allows tags to propagate across requests.
+ *
+ * OpenCensus uses W3C Correlation Context as the HTTP text format.
+ * https://github.com/w3c/correlation-context/blob/master/correlation_context/
+ * HTTP_HEADER_FORMAT.md
```
draffensperger: Nit: could you move the `HTTP_HEADER_FORMAT.md` to the line above so the link is not broken across lines? I think links are a good exception to the 80 char line limit

mayurkale22:
> LGTM, I assume there will be another PR adding the `Correlation-Context` header?

Yes, the actual use of the `Correlation-Context` header will be done in a follow up PR.

packages/opencensus-core/src/tags/propagation/text-format.ts
mayurkale22: SG, done

PR #446, opened 2019-03-26
Message events: use counters instead of hexdecimal value

### LINKED ISSUE #442: 400 bad request when exporting traces to Stackdriver
I'm seeing the following error when sending the trace information to `https://cloudtrace.googleapis.com` using the `'@opencensus/exporter-stackdriver' package:

```json
{
  "error": {
    "code": 400,
    "message": "Invalid value at 'spans[0].time_events.time_event[0].message_event.id' (TYPE_INT64), \"78f514e909094a98804da597eb454ba7\"",
    "errors": [
      {
        "message": "Invalid value at 'spans[0].time_events.time_event[0].message_event.id' (TYPE_INT64), \"78f514e909094a98804da597eb454ba7\"",
        "domain": "global",
        "reason": "badRequest"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```

From the [Google Documentation for cloud tracing](https://cloud.google.com/trace/docs/reference/v2/rest/v2/TimeEvents#messageevent), the type for the a `MessageEvent` should be a `string (int64 format)` whereas the id that is set is a hexadecimal value.

I'm not sure if the API recently made this change or something is wrong in my setup. Anyone able to confirm if this is/isn't an issue? Happy to lend a hand to fix it as well.

[Full request/response log](https://gist.github.com/hollandben/d7d75e0b99e12e110e993ea8db69500e)

mayurkale22:
Fixes #442 

As per the [Specs](https://github.com/census-instrumentation/opencensus-proto/blob/master/src/opencensus/proto/trace/v1/trace.proto#L202-L206), The `MessageId` must be calculated as two different counters starting from 1 one for sent messages and one for received message. This way we guarantee that the values will be consistent between different implementations.

~~To avoid breaking changes, I preferred to keep `string` type for id field. I am happy to change that or add support for both `string|number`.~~

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
-import {BasePlugin, CanonicalCode, deserializeBinary, PluginInternalFiles, RootSpan, serializeBinary, Span, SpanContext, SpanKind, TagMap, TagTtl, TraceOptions} from '@opencensus/core';
+import {BasePlugin, CanonicalCode, deserializeBinary, MessageEventType, PluginInternalFiles, RootSpan, serializeBinary, Span, SpanContext, SpanKind, TagMap, TagTtl, TraceOptions} from '@opencensus/core';
 import {deserializeSpanContext, serializeSpanContext} from '@opencensus/propagation-binaryformat';
 import {EventEmitter} from 'events';
 import * as grpcTypes from 'grpc';
 import * as lodash from 'lodash';
 import * as shimmer from 'shimmer';
+
 import * as clientStats from './grpc-stats/client-stats';
 import * as serverStats from './grpc-stats/server-stats';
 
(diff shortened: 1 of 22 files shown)


packages/opencensus-instrumentation-grpc/src/grpc.ts
```diff
@@ -84,6 +84,8 @@ export class GrpcPlugin extends BasePlugin {
   static readonly ATTRIBUTE_GRPC_STATUS_CODE = 'grpc.status_code';
   static readonly ATTRIBUTE_GRPC_ERROR_NAME = 'grpc.error_name';
   static readonly ATTRIBUTE_GRPC_ERROR_MESSAGE = 'grpc.error_message';
+  private sentSeqId = 1;
```
draffensperger: My reading of the specs is that these counters are supposed to be per-request so that the messages for a given request can be correlated between services. Would it be possible to handle these on a per request basis?

mayurkale22: Ah! You're right! I checked in other libraries. In [Go](https://github.com/census-instrumentation/opencensus-go/blob/264a2a48d94c062252389fffbc308ba555e35166/plugin/ocgrpc/trace_common.go#L93), id is 0 for all requests. In [Python](https://github.com/census-instrumentation/opencensus-python/blob/master/contrib/opencensus-ext-grpc/opencensus/ext/grpc/utils.py#L31), wraps  a request or response iterator to add message events to the span. Looks like Java has most accurate implementation, But I am not sure how it translate previous hop count (seq) on wire. I will talk to @songy23 and @rghetia to understand it better.

PR #453, opened 2019-03-29
[Core] Enforce strictNullChecks

### LINKED ISSUE #348: feat: enable strictNullChecks on this repo.

mayurkale22:
Fixes #348. This was the last package remaining for the enforcing `strictNullChecks` option.

### DIFF
--- a/packages/opencensus-core/src/metrics/metric-registry.ts
+++ b/packages/opencensus-core/src/metrics/metric-registry.ts
@@ -198,7 +198,13 @@ class MetricProducerForRegistry extends BaseMetricProducer {
    * @returns {Metric[]} The list of metrics.
    */
   getMetrics(): Metric[] {
-    return Array.from(
-        this.registeredMetrics, ([_, meter]) => meter.getMetric());
+    const metricArray = [];
+    for (const meter of this.registeredMetrics.values()) {
+      const metric = meter.getMetric();
+      if (metric) {
+        metricArray.push(metric);
+      }
+    }
+    return metricArray;
   }
 }
--- a/packages/opencensus-core/src/metrics/metric-registry.ts
+++ b/packages/opencensus-core/src/metrics/metric-registry.ts
@@ -198,13 +198,8 @@ class MetricProducerForRegistry extends BaseMetricProducer {
    * @returns {Metric[]} The list of metrics.
    */
   getMetrics(): Metric[] {
-    const metricArray = [];
-    for (const meter of this.registeredMetrics.values()) {
-      const metric = meter.getMetric();
-      if (metric) {
-        metricArray.push(metric);
-      }
-    }
-    return metricArray;
+    return Array.from(this.registeredMetrics.values())
+               .map(meter => meter.getMetric())
+               .filter(meter => !!meter) as Metric[];
   }
 }
--- a/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
+++ b/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
@@ -16,6 +16,7 @@
 
 import * as path from 'path';
 import * as semver from 'semver';
+import {noopLogger} from '../../common/noop-logger';
 import {Logger} from '../../common/types';
 import {Stats} from '../../stats/types';
 import * as modelTypes from '../model/types';
@@ -42,7 +43,7 @@ export abstract class BasePlugin implements types.Plugin {
   /** The module version. */
   protected version!: string;
   /** a logger */
-  protected logger!: Logger;
+  protected logger: Logger = noopLogger;
   /** list of internal files that need patch and are not exported by default */
   protected readonly internalFileList!: types.PluginInternalFiles;
   /**  internal files loaded */
--- a/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
+++ b/packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
@@ -13,13 +13,12 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 import * as path from 'path';
 import * as semver from 'semver';
-
 import {Logger} from '../../common/types';
 import {Stats} from '../../stats/types';
 import * as modelTypes from '../model/types';
-
 import * as types from './types';
 
 /**
@@ -39,21 +38,21 @@ export abstract class BasePlugin implements types.Plugin {
   /** The module name */
   protected moduleName: string;
   /** A tracer object. */
-  protected tracer: modelTypes.Tracer;
+  protected tracer!: modelTypes.Tracer;
   /** The module version. */
-  protected version: string;
+  protected version!: string;
   /** a logger */
-  protected logger: Logger;
+  protected logger!: Logger;
   /** list of internal files that need patch and are not exported by default */
-  protected readonly internalFileList: types.PluginInternalFiles;
+  protected readonly internalFileList!: types.PluginInternalFiles;
   /**  internal files loaded */
-  protected internalFilesExports: ModuleExportsMapping;
+  protected internalFilesExports!: ModuleExportsMapping;
   /** module directory - used to load internal files */
-  protected basedir: string;
+  protected basedir!: string;
   /** plugin options */
-  protected options: types.PluginConfig;
+  protected options!: types.PluginConfig;
   /** A stats object. */
-  protected stats: Stats;
+  protected stats?: Stats;
 
   /**
    * Constructs a new BasePlugin instance.
@@ -79,7 +78,7 @@ export abstract class BasePlugin implements types.Plugin {
     this.moduleExports = moduleExports;
     this.tracer = tracer;
     this.version = version;
-    this.basedir = basedir;
+    if (basedir) this.basedir = basedir;
     this.logger = tracer.logger;
     this.options = options;
     this.stats = stats;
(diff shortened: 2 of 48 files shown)


packages/opencensus-core/src/metrics/metric-registry.ts
```diff
@@ -198,7 +198,13 @@ class MetricProducerForRegistry extends BaseMetricProducer {
    * @returns {Metric[]} The list of metrics.
    */
   getMetrics(): Metric[] {
-    return Array.from(
-        this.registeredMetrics, ([_, meter]) => meter.getMetric());
+    const metricArray = [];
```
draffensperger: What would you think about rewriting this loop as `return this.registeredMetrics.values().map(m => m.getMetric()).filter(m => !!m);`?

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
```diff
@@ -39,21 +38,21 @@ export abstract class BasePlugin implements types.Plugin {
   /** The module name */
   protected moduleName: string;
   /** A tracer object. */
-  protected tracer: modelTypes.Tracer;
+  protected tracer!: modelTypes.Tracer;
```
draffensperger: Normally I would discourage `!` for properties like this, but I think it is worth it to get the strict null checks enforced.

I think ideally these would be initialized in the constructor for the plugin, but I'm guessing that would involve a bunch more refactoring.

packages/opencensus-core/src/trace/model/no-record/no-record-span-base.ts
```diff
@@ -44,19 +44,19 @@ export abstract class NoRecordSpanBase implements types.Span {
   /** Pointers from the current span to another span */
   links: types.Link[] = [];
   /** If the parent span is in another process. */
-  remoteParent: boolean;
+  remoteParent = false;
   /** The span ID of this span's parent. If it's a root span, must be empty */
-  parentSpanId: string = null;
+  parentSpanId!: string;
```
draffensperger: What about `parentSpanId = ''`?

packages/opencensus-core/src/trace/model/no-record/no-record-span-base.ts
```diff
@@ -44,19 +44,19 @@ export abstract class NoRecordSpanBase implements types.Span {
...
   /** The resource name of the span */
-  name: string = null;
+  name!: string;
```
draffensperger: Is there a default `name` we give spans? Could we specify it here or even call it `'no-record'` or similar?

packages/opencensus-core/src/trace/model/no-record/no-record-span-base.ts
```diff
@@ -44,19 +44,19 @@ export abstract class NoRecordSpanBase implements types.Span {
...
   /** Kind of span. */
   kind: types.SpanKind = types.SpanKind.UNSPECIFIED;
   /** A final status for this span */
   status: types.Status = STATUS_OK;
   /** set isRootSpan  */
   abstract get isRootSpan(): boolean;
   /** Trace Parameters */
-  activeTraceParams: configTypes.TraceParams;
+  activeTraceParams!: configTypes.TraceParams;
```
draffensperger: Can we just have a default of `{}`?

packages/opencensus-core/src/trace/model/no-record/no-record-span-base.ts
```diff
@@ -34,7 +34,7 @@ export abstract class NoRecordSpanBase implements types.Span {
   /** The Span ID of this span */
   readonly id: string;
   /** An object to log information to */
-  logger: Logger;
+  logger!: Logger;
```
draffensperger: Maybe default this to a console logger or even a `NopLogger` that does nothing? That would prevent exceptions if people try to invoke log functions on a NoRecordSpan I would think.

packages/opencensus-core/src/trace/model/no-record/no-record-span-base.ts
```diff
@@ -100,12 +100,7 @@ export abstract class NoRecordSpanBase implements types.Span {
 
   /** Gives the TraceContext of the span. */
   get spanContext(): types.SpanContext {
-    return {
-      traceId: this.traceId,
-      spanId: this.id,
-      options: 0,
-      traceState: this.traceState
-    };
+    return {traceId: this.traceId, spanId: this.id, options: 0};
```
draffensperger: Why is this removing `traceState`?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -27,7 +27,7 @@ const STATUS_OK = {
 export abstract class SpanBase implements types.Span {
   protected className: string;
   /** The clock used to mesure the beginning and ending of a span */
-  private clock: Clock = null;
+  private clock!: Clock;
```
draffensperger: Is there a default clock we can give this?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -38,7 +38,7 @@ export abstract class SpanBase implements types.Span {
   /** The Span ID of this span */
   readonly id: string;
   /** An object to log information to */
-  logger: Logger;
+  logger!: Logger;
```
draffensperger: Similar here on a default logger? (And could giving these defaults allow removing any initialization from the subclasses?)

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -48,19 +48,19 @@ export abstract class SpanBase implements types.Span {
   /** Pointers from the current span to another span */
   links: types.Link[] = [];
   /** If the parent span is in another process. */
-  remoteParent: boolean;
+  remoteParent = false;
   /** The span ID of this span's parent. If it's a root span, must be empty */
-  parentSpanId: string = null;
+  parentSpanId!: string;
```
draffensperger: Default of `''`?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -48,19 +48,19 @@ export abstract class SpanBase implements types.Span {
...
   /** The resource name of the span */
-  name: string = null;
+  name!: string;
```
draffensperger: Default?

packages/opencensus-core/src/trace/model/span-base.ts
```diff
@@ -48,19 +48,19 @@ export abstract class SpanBase implements types.Span {
...
   /** Kind of span. */
   kind: types.SpanKind = types.SpanKind.UNSPECIFIED;
   /** A final status for this span */
   status: types.Status = STATUS_OK;
   /** set isRootSpan  */
   abstract get isRootSpan(): boolean;
   /** Trace Parameters */
-  activeTraceParams: configTypes.TraceParams;
+  activeTraceParams!: configTypes.TraceParams;
```
draffensperger: Default to `{}`?

packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -246,7 +246,7 @@ export interface SpanOptions {
   parentSpanId?: string;
 }
 
-export type TraceState = string;
+export type TraceState = string|undefined;
```
draffensperger: What would you think about instead making the places that use `TraceState` explicitly include `|undefined` (maybe even with a find/replace)?

I think making a type include `undefined` or `null` is kind of risks confusing people reading the code (and maybe even bugs) because when they see a type that naturally assume it wouldn't be null/undefined.

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
mayurkale22: Currently users need to pass `tracer` instance, when enabling the plugins.  It would need more work and breaking change in order to initialize same in the constructor.

packages/opencensus-core/src/metrics/metric-registry.ts
mayurkale22: Done

mayurkale22:
@draffensperger Finally, a build is passed and all comments are resolved. PTAL

packages/opencensus-core/src/trace/instrumentation/base-plugin.ts
draffensperger: OK, makes sense!

PR #460, opened 2019-04-02
Add gauge example

### LINKED ISSUE #459: Add gauge example
Add gauge example, just like one in [java](https://github.com/census-instrumentation/opencensus-java/tree/master/examples/src/main/java/io/opencensus/examples/gauges).

mayurkale22:
Fixes #459

### DIFF
--- a/examples/gauge/derived-gauge-quickstart.js
+++ b/examples/gauge/derived-gauge-quickstart.js
@@ -1,0 +1,51 @@
+/**
+ * Copyright 2019, OpenCensus Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * This is an example shows how to create a Derived Gauge metric. This gauge is
+ * self sufficient once created
+ */
+
+const { Metrics, MeasureUnit } = require('@opencensus/core');
+
+// [UNCOMMENT THIS BLOCK to visualize the data =======================]
+// // Enable OpenCensus exporters to export gauges to Stackdriver Monitoring.
+// const { StackdriverStatsExporter } = require('@opencensus/exporter-stackdriver');
+// const exporter = new StackdriverStatsExporter({ projectId: 'projectId' });
+// const { globalStats } = require('@opencensus/core');
+// globalStats.registerExporter(exporter);
+// [END setup_exporter ==============================================]
+
+setTimeout(() => {
+  console.log('Timeout executed.');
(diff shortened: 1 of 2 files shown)


examples/gauge/derived-gauge-quickstart.js
```diff
@@ -0,0 +1,51 @@
...
+// [UNCOMMENT THIS BLOCK to visualize the data =======================]
+// // Enable OpenCensus exporters to export gauges to Stackdriver Monitoring.
+// const { StackdriverStatsExporter } = require('@opencensus/exporter-stackdriver');
+// const exporter = new StackdriverStatsExporter({ projectId: 'projectId' });
+// const { globalStats } = require('@opencensus/core');
+// globalStats.registerExporter(exporter);
+// [END setup_exporter ==============================================]
+
+setTimeout(() => {
+  console.log('Timeout executed.');
```
draffensperger: Could you document what this timeout is for? I remember in another example the timeout kept the Node process alive while the metric was being collected.

examples/gauge/derived-gauge-quickstart.js
```diff
@@ -0,0 +1,51 @@
...
+
+// a new gauge instance - builds a new Int64 gauge to be added to the registry.
+const metricOptions = {
+  description: 'Number of active handles',
+  unit: MeasureUnit.UNIT,
+  labelKeys: labelKeys
+};
+const gauge = metricRegistry.addDerivedInt64Gauge('active_handles_total', metricOptions);
+
+gauge.createTimeSeries(labelValues, process._getActiveHandles());
```
draffensperger: Why do you pass `process._getActiveHandles()` vs. passing the function itself `process._getActiveHandles`? I would have expected it to take a function that it calls to calculate the gauge value rather than the output of a function for a single value.

examples/gauge/gauge-quickstart.js
```diff
@@ -0,0 +1,53 @@
...
+
+// [UNCOMMENT THIS BLOCK to visualize the data =======================]
+// // Enable OpenCensus exporters to export gauges to Stackdriver Monitoring.
+// const { StackdriverStatsExporter } = require('@opencensus/exporter-stackdriver');
+// const exporter = new StackdriverStatsExporter({ projectId: 'projectId' });
+// const { globalStats } = require('@opencensus/core');
+// globalStats.registerExporter(exporter);
+// [END setup_exporter ==============================================]
+
+const nowInSeconds = Math.round(Date.now() / 1000 - process.uptime());
```
draffensperger: Why subtract the process uptime?

examples/gauge/derived-gauge-quickstart.js
mayurkale22: The `createTimeSeries` method accepts the object with `AccessorInterface` type. 

```
AccessorInterface = LengthAttributeInterface | LengthMethodInterface | 
                    SizeAttributeInterface | SizeMethodInterface | ToValueInterface;
```

I have changed the example to show derived gauge example with queue depth. Do you think we should add support for function based extractor?

draffensperger: Ah, makes sense. Yes, I think supporting a function as well would be nice so that people can just pass in an anonymous arrow function easily.

mayurkale22: Ok sure, I will open an issue to include function based extractor in derived gauge API. Thanks

PR #468, opened 2019-04-03
 Gauge: Add support for constant labels

mayurkale22:
This is as per the specs: https://github.com/census-instrumentation/opencensus-specs/pull/248

### DIFF
--- a/packages/opencensus-core/src/common/validations.ts
+++ b/packages/opencensus-core/src/common/validations.ts
@@ -50,8 +50,7 @@ export function validateArrayElementsNotNull<T>(
 export function validateMapElementNotNull<T>(
     map: Map<T, T>, errorMessage: string) {
   for (const [key, value] of map.entries()) {
-    if (key === null || key === undefined || value === null ||
-        value === undefined) {
+    if (key == null || value == null) {
       throw new Error(`${errorMessage} elements should not be a NULL`);
     }
   }
--- a/packages/opencensus-core/src/common/validations.ts
+++ b/packages/opencensus-core/src/common/validations.ts
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import {LabelKey, LabelValue} from '../metrics/export/types';
+
 /**
  * Validates that an object reference passed as a parameter to the calling
  * method is not null.
@@ -41,5 +43,29 @@ export function validateArrayElementsNotNull<T>(
       element => element !== null && typeof element !== 'undefined');
   if (!areAllDefined) {
     throw new Error(`${errorMessage} elements should not be a NULL`);
+  }
+}
+
+/** Throws an error if any of the map elements is null. */
+export function validateMapElementNotNull<T>(
+    map: Map<T, T>, errorMessage: string) {
+  for (const [key, value] of map.entries()) {
+    if (key === null || key === undefined || value === null ||
+        value === undefined) {
+      throw new Error(`${errorMessage} elements should not be a NULL`);
+    }
+  }
+}
+
+/** Throws an error if any of the array element present in the map. */
+export function validateDuplicateKeys(
+    keys: LabelKey[], map: Map<LabelKey, LabelValue>) {
+  const keyNames: Set<string> = new Set();
+  keys.forEach(key => keyNames.add(key.key));
+  Array.from(map.keys()).forEach(key => keyNames.add(key.key));
+
+  if (keyNames.size !== (keys.length + map.size)) {
+    throw new Error(
+        `The keys from LabelKeys should not be present in constantLabels`);
   }
 }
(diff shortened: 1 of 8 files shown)


packages/opencensus-core/src/common/validations.ts
```diff
@@ -43,3 +45,27 @@ export function validateArrayElementsNotNull<T>(
     throw new Error(`${errorMessage} elements should not be a NULL`);
   }
 }
+
+/** Throws an error if any of the map elements is null. */
+export function validateMapElementNotNull<T>(
+    map: Map<T, T>, errorMessage: string) {
+  for (const [key, value] of map.entries()) {
+    if (key === null || key === undefined || value === null ||
```
draffensperger: Optional: you can use `== null` to test for both null and undefined, which is a permissible exception IMO to general (good) advice to always use `===`. See https://stackoverflow.com/questions/2559318/how-to-check-for-an-undefined-or-null-variable-in-javascript for details.

packages/opencensus-core/src/common/validations.ts
```diff
@@ -43,3 +45,27 @@ export function validateArrayElementsNotNull<T>(
...
+        value === undefined) {
+      throw new Error(`${errorMessage} elements should not be a NULL`);
+    }
+  }
+}
+
+/** Throws an error if any of the array element present in the map. */
+export function validateDuplicateKeys(
+    keys: LabelKey[], map: Map<LabelKey, LabelValue>) {
+  const keyNames: Set<string> = new Set();
```
draffensperger: Optional: could this whole block become just be something like `const keysAndConstantKeys = new Set([...keyNames, ...map.keys].map(k => k.key))`?

packages/opencensus-core/src/common/validations.ts
```diff
@@ -43,3 +45,27 @@ export function validateArrayElementsNotNull<T>(
...
+    if (key === null || key === undefined || value === null ||
+        value === undefined) {
+      throw new Error(`${errorMessage} elements should not be a NULL`);
+    }
+  }
+}
+
+/** Throws an error if any of the array element present in the map. */
+export function validateDuplicateKeys(
+    keys: LabelKey[], map: Map<LabelKey, LabelValue>) {
```
draffensperger: Should `map` be called `constantLabels` here?

packages/opencensus-core/src/metrics/gauges/derived-gauge.ts
```diff
@@ -94,12 +95,19 @@ export class DerivedGauge implements types.Meter {
...
+   *     labels for the Metric.
    */
   constructor(
       name: string, description: string, unit: string,
-      type: MetricDescriptorType, labelKeys: LabelKey[]) {
-    this.metricDescriptor = {name, description, unit, type, labelKeys};
+      type: MetricDescriptorType, labelKeys: LabelKey[],
+      readonly constantLabels: Map<LabelKey, LabelValue>) {
     this.labelKeysLength = labelKeys.length;
+    const combinedLabelKeys = labelKeys.concat(...constantLabels.keys());
```
draffensperger: I notice below you use `[...gaugeEntry.labelValues, ...this.constantLabelValues]`, so should this be `[...labelKeys, ...consantLabels.keys()]` to match?

packages/opencensus-core/src/common/validations.ts
```diff
@@ -43,3 +45,27 @@ export function validateArrayElementsNotNull<T>(
...
+  const keyNames: Set<string> = new Set();
+  keys.forEach(key => keyNames.add(key.key));
```
songy23: nit: consider adding a separate check on duplicated keys in `keys`:
```
if (keyNames.size !== keys.length) {
  throw new Error(...);
}
```

packages/opencensus-core/src/metrics/gauges/derived-gauge.ts
mayurkale22: Good Idea, Done.

packages/opencensus-core/src/common/validations.ts
mayurkale22: Sure

packages/opencensus-core/src/common/validations.ts
mayurkale22: Done

packages/opencensus-core/src/common/validations.ts
mayurkale22: For simplicity, I have combined with existing check.

packages/opencensus-core/src/common/validations.ts
mayurkale22: SGTM, done. Thanks for sharing the link.

PR #469, opened 2019-04-04
Derived Gauge: Add support for extracting value from a function

### LINKED ISSUE #463: Derived Gauge: Add support for function based extractor
The `createTimeSeries ` method should allow users to pass (anonymous) function, that it calls to calculate or extract the gauge value whenever metrics are collected.

mayurkale22:
Fixes #463

### DIFF
--- a/examples/gauge/derived-gauge-quickstart.js
+++ b/examples/gauge/derived-gauge-quickstart.js
@@ -32,7 +32,7 @@ const { Metrics, MeasureUnit } = require('@opencensus/core');
 // To instrument a queue's depth.
 class QueueManager {
   constructor () { this.depth = 0; }
-  getValue () { return this.depth; }
+  getPendingJobs () { return this.depth; }
   addJob () { this.depth++; }
 }
 
@@ -54,6 +54,8 @@ const gauge = metricRegistry.addDerivedInt64Gauge('active_handles_total', metric
 const queue = new QueueManager();
 queue.addJob();
 
-// The value of the gauge is observed from the obj whenever metrics are
+// The value of the gauge is observed from a function whenever metrics are
 // collected. In this case it will be 1.
-gauge.createTimeSeries(labelValues, queue);
+gauge.createTimeSeries(labelValues, () => {
+  queue.getPendingJobs();
+});
--- a/examples/gauge/derived-gauge-quickstart.js
+++ b/examples/gauge/derived-gauge-quickstart.js
@@ -56,6 +56,4 @@ queue.addJob();
 
 // The value of the gauge is observed from a function whenever metrics are
 // collected. In this case it will be 1.
-gauge.createTimeSeries(labelValues, () => {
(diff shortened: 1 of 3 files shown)


examples/gauge/derived-gauge-quickstart.js
```diff
@@ -54,6 +54,8 @@ const gauge = metricRegistry.addDerivedInt64Gauge('active_handles_total', metric
 const queue = new QueueManager();
 queue.addJob();
 
-// The value of the gauge is observed from the obj whenever metrics are
+// The value of the gauge is observed from a function whenever metrics are
 // collected. In this case it will be 1.
-gauge.createTimeSeries(labelValues, queue);
+gauge.createTimeSeries(labelValues, () => {
```
draffensperger: Can this just be `queue.getPendingJobs` with no need for a wrapper arrow function?

packages/opencensus-core/src/metrics/gauges/derived-gauge.ts
```diff
@@ -159,16 +163,18 @@ export class DerivedGauge implements types.Meter {
...
-    } else if (DerivedGauge.isLengthAttributeInterface(obj)) {
-      this.extractor = () => obj.length;
-    } else if (DerivedGauge.isLengthMethodInterface(obj)) {
-      this.extractor = () => obj.length();
-    } else if (DerivedGauge.isSizeAttributeInterface(obj)) {
-      this.extractor = () => obj.size;
-    } else if (DerivedGauge.isSizeMethodInterface(obj)) {
-      this.extractor = () => obj.size();
+    if (typeof objOrFn === 'function') {
+      this.extractor = () => objOrFn();
```
draffensperger: Can you just say `this.extractor = objOrFun`?

mayurkale22: To allow this I have to add new type (`ValueExtractorOrFunction`) which is type of `Function|ValueExtractor`. Let me know if this is what is desired.

draffensperger: Why is that? I would have assumed that the internal `extractor` variable would always of type `() => number` (a function that returns a number). Then these different if statements are mapping from the various shapes of input to a simple function that returns a number. So in the case when the input is a function, could we just assign it directly?

Or am I missing something here?

examples/gauge/derived-gauge-quickstart.js
mayurkale22: Done.

packages/opencensus-core/src/metrics/gauges/derived-gauge.ts
mayurkale22: Defining a strongly-typed function solved this problem.

draffensperger: Cool!

PR #471, opened 2019-04-04
add redis instrumentation plugins by default

vmarchaud:
As mentioned before, adding a new plugin to the default list need 2 different global release currently:
- the first one to push it to npm and add it as a dependency of the `instrumentation-all` module
- and the second to publish the new version of `instrumentation-all` so the new plugin are enabled by default.

We might want to change that to release new plugins faster. How do you think we could do that ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,8 @@ OpenCensus can collect tracing data automatically using plugins. Users can also 
 - [HTTPS](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-instrumentation-https/README.md)
 - [HTTP2](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-instrumentation-http2/README.md)
 - [GRPC](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-instrumentation-grpc/README.md)
+- [IORedis](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-instrumentation-ioredis/README.md)
+- [Redis](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-instrumentation-redis/README.md)
 
 ## Propagation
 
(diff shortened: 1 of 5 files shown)


mayurkale22:
> We might want to change that to release new plugins faster. How do you think we could do that ?

If it's really required, I can do the patch release. For now, I would suggest to update the CHANGELOG (this way users will know that it is still in unreleased version). WDYT?

vmarchaud:
> Thanks for the contribution, It will be nice to see the working example for Redis and IORedis.

Do you want me to add some examples with the zipkin UI ?



> > We might want to change that to release new plugins faster. How do you think we could do that ?
> 
> If it's really required, I can do the patch release. For now, I would suggest to update the CHANGELOG (this way users will know that it is still in unreleased version). WDYT?

I would not mind for now, but i think in the long run users might be frustrated to wait 2 releases to be able to use new plugins. Do you want me to modify the CHANGELOG in this PR or open up another one ?

mayurkale22:
> Do you want me to add some examples with the zipkin UI ?

That would be awesome, thanks.

> Do you want me to modify the CHANGELOG in this PR or open up another one ?

Updating CHANGELOG in this PR makes sense to me. 

> I think in the long run users might be frustrated to wait 2 releases to be able to use new plugins. 

I think we can do in one release only, Once you add new plugin we can add into `instrumentation-all` module with 0.0.1 version.  [This is](https://github.com/census-instrumentation/opencensus-node/pull/354) what I did for binaryformat package. If you do `npm install ` on top/main package lerna will help you to bind local (unreleased) package dependency properly. No issues for build or running locally.

One caveat is that if you do `npm install` on individual package where you have added this new dependency, it will fail to find the package. The workaround is you should run `npm install` on parent package. 

I hope you got me, otherwise let me know.

vmarchaud:
Sorry i didn't had the time to update the PR as i was traveling back to France. I will open a new one

mayurkale22:
> Sorry i didn't had the time to update the PR as i was traveling back to France. I will open a new one

np, I will do that for you in post release.

Issue #450, opened 2019-03-27
Metrics: Add Cumulative APIs


mayurkale22:
Specs tracking issue: census-instrumentation/opencensus-specs#241.

PR #480, opened 2019-04-09
Add support for Cumulative API

mayurkale22:
Updates #450 

Adding **Derived Cumulative API** and registering with **[Metric-Registry](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/metrics/metric-registry.ts)** will be in a follow-up PR.

### DIFF
--- a/packages/opencensus-core/src/metrics/cumulative/cumulative.ts
+++ b/packages/opencensus-core/src/metrics/cumulative/cumulative.ts
@@ -170,11 +170,11 @@ export class CumulativePointEntry implements CumulativePoint {
    * @param {number} val The new value.
    */
   inc(val?: number): void {
-    if (val && !Number.isFinite(val)) {
+    if ((val && !Number.isFinite(val)) || (val !== undefined && isNaN(val))) {
       throw new TypeError(`Value is not a valid number: ${val}`);
     }
     if (val && val < 0) {
-      throw new Error('It is not possible to decrease a counter');
+      throw new Error('It is not possible to decrease a cumulative metric');
     }
     const incValue = (val === null || val === undefined) ? 1 : val;
     this.value += incValue;
--- a/packages/opencensus-core/src/metrics/cumulative/cumulative.ts
+++ b/packages/opencensus-core/src/metrics/cumulative/cumulative.ts
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018, OpenCensus Authors
+ * Copyright 2019, OpenCensus Authors
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


packages/opencensus-core/src/metrics/cumulative/cumulative.ts
```diff
@@ -0,0 +1,196 @@
...
+    }
+
+    const point =
+        new CumulativePointEntry([...labelValues, ...this.constantLabelValues]);
+    this.registeredPoints.set(hash, point);
+    return point;
+  }
+
+  /**
+   * Provides a Metric with one or more TimeSeries.
```
draffensperger: Can you update this comment? And also explain when it would return `null`?

packages/opencensus-core/src/metrics/cumulative/cumulative.ts
```diff
@@ -0,0 +1,196 @@
...
+  reset(): void {
+    this.value = 0;
+    this.startTimestamp = getTimestampWithProcessHRTime();
+  }
+
+  /**
+   * Increment the cumulative metric.
+   * @param {number} val The new value.
+   */
+  inc(val?: number): void {
```
draffensperger: Nit: should this be `increment` to be clearer?

packages/opencensus-core/src/metrics/cumulative/cumulative.ts
```diff
@@ -0,0 +1,196 @@
...
+    if (val && !Number.isFinite(val)) {
```
draffensperger: Does this catch `NaN` (I'm guessing not), and assuming not, could you make it do that?

packages/opencensus-core/src/metrics/types.ts
```diff
@@ -0,0 +1,42 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {MeasureUnit} from './../stats/types';
+import {LabelKey, LabelValue, Metric} from './export/types';
+
+export interface Meter {
```
draffensperger: Could you add a top-level comment for this? In particularly, I'm curious to have it say the difference between a `Meter` and a `Metric`

packages/opencensus-core/test/test-cumulative.ts
```diff
@@ -0,0 +1,298 @@
...
+      }, /^Error: It is not possible to decrease a counter$/);
+    });
+
+    it('should throw a TypeError when the inc number is not finite', () => {
+      const point = instance.getOrCreateTimeSeries(LABEL_VALUES_200);
+      assert.throws(() => {
+        point.inc(100 / 0);
+      }, /^TypeError: Value is not a valid number: Infinity$/);
+    });
+
```
draffensperger: Can you add a test case for `NaN`?

packages/opencensus-core/src/metrics/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  /**
+   * Provides a Metric with one or more TimeSeries.
```
draffensperger: Could you explain when this should return `null`?

packages/opencensus-core/src/metrics/cumulative/cumulative.ts
mayurkale22: Done, basically `null` is returned when TimeSeries is not present in Metric.

packages/opencensus-core/src/metrics/cumulative/cumulative.ts
mayurkale22: hmm... In the [specs](https://github.com/census-instrumentation/opencensus-specs/pull/254/files) we have mentioned `inc` name and same has been used in Go also.

packages/opencensus-core/src/metrics/types.ts
mayurkale22: Done.

packages/opencensus-core/test/test-cumulative.ts
mayurkale22: Done.

packages/opencensus-core/src/metrics/types.ts
mayurkale22: Done. `Metric` contains MetricDescriptor and one or more timeseries while `Meter` interface is to generate `Metric` which is implemented by `Gauge` and `Cumulative` metric.

packages/opencensus-core/src/metrics/cumulative/cumulative.ts
mayurkale22: Actually `increment` looks good :). I am open for both, let me know your preference.

draffensperger: I don't think it's a major deal either way. In general, my sense is that the Go community prefers terse variable/function names, while the JS community likes more descriptive ones. But short names have benefits too.

I checked and it looks like a Prometheus JS client uses `inc()` (https://github.com/siimon/prom-client#counter), so let's leave it as `inc`

mayurkale22: Ok makes sense. Thanks

Issue #481, opened 2019-04-09
http-instrumentation: propagation error when using Expect header


vmarchaud:
`google-trace-nodejs` had the same issue before: https://github.com/googleapis/cloud-trace-nodejs/pull/766

I will make a PR later today to fix that

mayurkale22:
Addressed in #482

PR #482, opened 2019-04-09
http-instrumentation: fix propagation errors when using Expect header #481

vmarchaud:


packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -360,12 +360,28 @@ export class HttpPlugin extends BasePlugin {
 
       const setter: HeaderSetter = {
         setHeader(name: string, value: string) {
-          request.setHeader(name, value);
+          // https://github.com/googleapis/cloud-trace-nodejs/pull/766
```
draffensperger: What would you think about removing this PR reference and instead putting a short explanation of the logic?

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -360,12 +360,28 @@ export class HttpPlugin extends BasePlugin {
...
 
       const propagation = plugin.tracer.propagation;
       if (propagation) {
+        // If outgoing request headers contain the "Expect" header, the returned
+        // ClientRequest will throw an error if any new headers are added. For
+        // this reason, only in this scenario, we opt to clone the options
+        // object to inject the trace context header instead of using
+        // ClientRequest#setHeader. (We don't do this generally because cloning
+        // the options object is an expensive operation.)
+        // https://github.com/googleapis/cloud-trace-nodejs/pull/766
```
draffensperger: The explanation above is great - can we remove the PR reference? (People can use the blame layer to find it if needed)

vmarchaud:
@draffensperger I've rewritten the explanation and removed the link to the google cloud PR

draffensperger:
Looks good, thanks!

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -374,6 +375,25 @@ describe('HttpPlugin', () => {
...
+         const options = Object.assign(
+             {headers: {Expect: '100-continue'}},
+             url.parse('http://google.fr/'));
+         await httpRequest.get(options).then((result) => {
+           assert.strictEqual(rootSpanVerifier.endedRootSpans.length, 1);
+           assert.ok(
+               rootSpanVerifier.endedRootSpans[0].name.indexOf('GET /') >= 0);
+
+           const span = rootSpanVerifier.endedRootSpans[0];
+           assertSpanAttributes(span, 301, 'GET', 'google.fr', '/', undefined);
```
mayurkale22: Can you remove this `undefined`, the `userAgent` is an optional param.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -360,12 +360,28 @@ export class HttpPlugin extends BasePlugin {
...
         }
       };
 
       const propagation = plugin.tracer.propagation;
       if (propagation) {
+        // If outgoing request headers contain the "Expect" header, the returned
+        // ClientRequest will throw an error if any new headers are added.
+        // So we need to clone the options object to be able to inject new
+        // header.
+        if (options.headers && plugin.hasExpectHeader(options)) {
```
mayurkale22: Is it safe to remove `options.headers` check as we are doing same in `hasExpectHeader` function?

vmarchaud:
Done @mayurkale22

mayurkale22:
Thanks for the contribution @vmarchaud

Issue #462, opened 2019-04-02
Separate CLS from Tracer and make it optional


hekike:
**Is your feature request related to a problem? Please describe.**

Continues local storage has an overhead which maybe makes sense in applications where the developer doesn't control the instrumentation and it should happen automatically but is unnecessary in services with explicit instrumentation.

Read more about common issues with CLS in https://github.com/othiym23/node-continuation-local-storage/issues/59.

**Describe the solution you'd like**

My proposal would be to remove CLS usage from the [CoreTracer](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/model/tracer.ts#L33) class and create a new class `CoreTracerCLS` which extends the `CoreTracer` and enhances it with `CLS`. Then we could expose the `CoreTracer` and keep `CoreTracerCLS` the default Tracer to avoid breaking change.

**Describe alternatives you've considered**

A bigger change would be to keep the Core OpenCensus Node module explicit and completely CLS and monkey patching free and create a new module that does the automatic instrumentation via the pure core module with CLS and require hooks.

**Additional context**

This proposal is slightly related to these changes which give more control to the developer:

- https://github.com/census-instrumentation/opencensus-node/pull/437
- https://github.com/census-instrumentation/opencensus-node/pull/441

I'm happy to help with the implementation and open a PR.

This Issue was created based on our conversation with @bogdandrutu, @mayurkale22, @kjin, @ofrobots, and @kenashcraft

mayurkale22:
Thanks for filling the issue, feel free to shoot the PR.

mayurkale22:
@hekike  Can we close this one via #484?

hekike:
Yes, thank you!

PR #484, opened 2019-04-11
feat(tracer): separate cls from tracer

hekike:
implements: https://github.com/census-instrumentation/opencensus-node/issues/462

**Changes**

- removed the argument notation interface of the `tracer.startChildSpan` because it brings ambiguity into span creation
- Separation of `CoreTracer` and `CoreTracerCls`

**Breaking Changes**

- `startChildSpan` only accepts `options` object
- `Tracer` become `TracerCls`
- `Tracer` is the basic non CLS version now

**Action Items**

- ~~separate tests out to `test-tracer.ts` and `test-tracer-cls.ts`~~ (done)
- ~~Improve non-cls tracer test coverage~~ (done)

**Child of child with CLS Tracer**

It's worth calling out that the CLS implementation still supports only one level of relationships as the `rootSpan` in the namespace context remains the same for child span creation. I guess it's expected based on current instrumentations.

**A couple of further optimization ideas**

- turn `startRootSpan` sync in the non-CLS Tracer
- merge `startRootSpan` and `startChildSpan` in the non CLS Tracer (`startSpan`)
- if we merge the two and make name required, we could get closer to the OpenTracing API: https://github.com/opentracing/opentracing-javascript/blob/master/src/tracer.ts#L79 which I guess after the project convergence is prefered (but maybe this is too early and ambitious)


packages/opencensus-core/src/trace/model/tracer-cls.ts
```diff
@@ -0,0 +1,142 @@
...
+ */
+
+import * as cls from '../../internal/cls';
+
+import {NoRecordSpan} from './no-record/no-record-span';
+import {CoreTracer} from './tracer';
+import * as types from './types';
+
+/**
+ * This class represent a tracer with CLS.
```
mayurkale22: Nit: `s/represent/represents`

packages/opencensus-core/src/trace/model/tracer-cls.ts
```diff
@@ -0,0 +1,142 @@
...
+  /** Clears the current root span. */
+  clearCurrentTrace() {
+    // TODO: Remove null reference and ts-ignore check.
+    //@ts-ignore
+    this.currentRootSpan = null;
+  }
+
+  /**
+   * Starts a span.
+   * @param [options] span options
```
mayurkale22: Nit: `s/@param [options] span options/@param options A SpanOptions object to start a child span./`

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -126,43 +109,39 @@ export class CoreTracer implements types.Tracer {
...
-          options && options.kind ? options.kind : types.SpanKind.UNSPECIFIED;
-
-      let parentSpanId = '';
-      let traceState;
-      if (options && options.spanContext) {
-        // New child span.
-        parentSpanId = options.spanContext.spanId || '';
-        traceState = options.spanContext.traceState;
+    let traceId;
+    if (options && options.spanContext && options.spanContext.traceId) {
```
mayurkale22: Can we just do `options = options || {name: 'span'};` at the start and remove falsiness check on `options` (multiple places below)?

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -229,60 +200,17 @@ export class CoreTracer implements types.Tracer {
...
-    // TODO: Remove null reference and ts-ignore check.
-    //@ts-ignore
-    this.currentRootSpan = null;
-  }
-
   /**
    * Starts a span.
-   * @param nameOrOptions Span name string or SpanOptions object.
-   * @param kind Span kind if not using SpanOptions object.
+   * @param [options] span options
```
mayurkale22: Same as above

packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -516,18 +513,25 @@ export interface Tracer extends SpanEventListener {
...
-
   /**
    * Start a new Span instance to the currentRootSpan
-   * @param name Span name
-   * @param kind Span kind
-   * @param options Span Options
+   * @param childOf Span
+   * @param [name] Span name
+   * @param [kind] Span kind
+   * @param [options] Span Options
```
mayurkale22: Same as above and just keep options param.

packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -516,18 +513,25 @@ export interface Tracer extends SpanEventListener {
...
-  startChildSpan(name?: string, kind?: SpanKind): Span;
   startChildSpan(options?: SpanOptions): Span;
+}
+
+/** Interface for TracerCls */
+export interface TracerCls extends Tracer {
+  /** Get and set the currentRootSpan to tracer instance */
+  currentRootSpan: Span;
+
+
```
mayurkale22: Nit: Remove extra blank line.

hekike:
A couple of further optimization ideas:

- turn `startRootSpan` sync in the non CLS Tracer
- merge `startRootSpan` and `startChildSpan` in the non CLS Tracer (`startSpan`)
- if we merge the two and make name required, we could get closer to the opentracing api: https://github.com/opentracing/opentracing-javascript/blob/master/src/tracer.ts#L79 which I guess after the project convergence is prefered (but maybe this is too early and ambitious)

hekike:
It's worth calling out that the CLS implementation still supports only one level of relationships as the `rootSpan` in the namespace context remains the same for child span creation. I guess it's expected based on current instrumentations.

packages/opencensus-core/src/trace/model/tracer.ts
hekike: actually, we could do just:

`function (options = { name: 'span' })`

hekike:
@mayurkale22 earlier I posted some API ideas as a comment earlier, I just moved them to the PR description now. Could you please share what do you think about them?

mayurkale22:
> @mayurkale22 earlier I posted some API ideas as a comment here, I just moved them to the PR description now. Could you please share what do you think about them?

Thanks for the detailed description, the **Changes** and **Action items** makes sense to me. In terms  of a **breaking changes**, could you please elaborate more on ambiguity on span creation? My initial impression was to keep default tracer as `TracerCls`, so this won't create any breaking change for the 
current users. Also, if feasible, (one approach) we can add `basicTracer` to `Tracing` instance, which users will call to get the basic non CLS version of tracer object. 

> merge startRootSpan and startChildSpan in the non CLS Tracer (startSpan)

SGTM, But would love to see same for CLS Tracer (easier to maintain and switch around). WDYT?

hekike:
@mayurkale22 thanks for the feedback! About the API ambiguity: The `startChildSpan`'s original syntax is `startChildSpan(name?: string, kind?: types.SpanKind)`, now we need to add a third `childOf` argument. This is problematic as all the arguments are optional today. If we add it as a first argument, the user needs to pass `undefined` for `tracerCls.startChildSpan(undefined, 'span-name')`. If we add it as the last argument you need to pass `undefined` when you want optional `name` and `kind`. We could build variadic arguments but in my understanding, it is against TypeScript philosophy, but maybe I'm missing something. That's why in think the OpenTracing span creation API is a good tradeoff with a required first string argument and optional secondary options: `tracer.startSpan('foo', { 
childOf: rootSpan })`.

mayurkale22:
> Only keeping the object options notations looks clear and robust solution to me, due to the TypeScript types it's easy to discover the available option properties.

Makes sense to me, please update CHANGELOG.md with a breaking change line. Thanks

packages/opencensus-core/test/test-tracer.ts
```diff
@@ -354,12 +355,12 @@ describe('Tracer', () => {
     it('should support nested children', () => {
       tracer.startRootSpan(options, (rootSpan) => {
         assert.strictEqual(rootSpan.numberOfChildren, 0);
-        const child1 =
-            tracer.startChildSpan('child1', types.SpanKind.UNSPECIFIED);
+        const child1 = tracer.startChildSpan(
+            {name: 'child1', kind: types.SpanKind.UNSPECIFIED});
```
kenashcraft: To simplify these tests and the API, can `kind` just have a default value?  Then we wouldn't need to specify it again and again.

packages/opencensus-core/src/trace/model/tracer-cls.ts
kenashcraft: Explain what CLS is and when you would or would not want to use it.

hekike:
@mayurkale22 @kenashcraft thanks for the review, I addressed the PR comments:

- non-breaking change `CoreTracer` and `CoreTracerBase` separation
- fixes in docs and typos
- documenting CLS usage
- updating `CHANGELOG.md` with `startChildSpan` changes
- cleaning up `startRootSpan` defaults

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -14,48 +14,28 @@
...
-import {NoRecordRootSpan} from './no-record/no-record-root-span';
+
 import {NoRecordSpan} from './no-record/no-record-span';
-import {RootSpan} from './root-span';
+import {CoreTracerBase} from './tracer-base';
 import * as types from './types';
 
 /**
- * This class represent a tracer.
+ * This class represents a tracer with Cointunues Local Storage (CLS).
```
mayurkale22: Nit: `s/Cointunues/Continuation/`

examples/http/server.js
```diff
@@ -42,7 +42,7 @@ function startServer (port) {
 
 /** A function which handles requests and send response. */
 function handleRequest (request, response) {
-  const span = tracer.startChildSpan('octutorials.handleRequest');
+  const span = tracer.startChildSpan({ name: 'octutorials.handleRequest' });
```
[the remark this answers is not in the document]
mayurkale22: Any update here?

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -0,0 +1,228 @@
...
+  private activeLocal: boolean;
+  /** A configuration for starting the tracer */
+  private config!: configTypes.TracerConfig;
+  /** A list of end span event listeners */
+  private eventListenersLocal: types.SpanEventListener[] = [];
+  /** Bit to represent whether trace is sampled or not. */
+  private readonly IS_SAMPLED = 0x1;
+  /** A sampler used to make sample decisions */
+  sampler!: samplerTypes.Sampler;
+  /** A configuration for starting the tracer */
```
mayurkale22: Please update this comment to something like -> ` An object to log information to`

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -0,0 +1,228 @@
...
+
+    if (this.active) {
+      const sampleDecision = this.makeSamplingDecision(options, traceId);
+      if (sampleDecision) {
+        const rootSpan =
+            new RootSpan(this, name, kind, traceId, parentSpanId, traceState);
+        rootSpan.start();
+        return fn(rootSpan);
+      }
+    } else {
```
mayurkale22: Can we remove this else, and update the debug msg: `Tracer is inactive, starting new no record root span`?

packages/opencensus-core/src/trace/model/tracer-cls.ts
mayurkale22: +1

packages/opencensus-core/src/trace/model/tracer.ts
```diff
@@ -14,48 +14,28 @@
...
+ *
+ * CLS helps keep tracking the root span over function calls automatically.
```
mayurkale22: Can we add ->  `It is capable of storing, propagating and retrieving arbitrary continuation-local data (also called "context").`

packages/opencensus-core/test/test-tracer.ts
```diff
@@ -52,7 +52,7 @@ describe('Tracer', () => {
     });
   });
 
-  /** Should get/set the current RootSpan from tracer instance */
+  // /** Should get/set the current RootSpan from tracer instance */
```
mayurkale22: Is this typo or intended stuff?

packages/opencensus-core/src/trace/model/tracer.ts
hekike: Added

hekike:
@mayurkale22 I addressed your new round of comments.

mayurkale22:
> Action Items (Will do after initial feedback)
separate tests out to test-tracer.ts and test-tracer-cls.ts
Improve non-cls tracer test coverage

@hekike Thanks for the work! What's your plan on above action items. Also, How do you plan to expose basic tracer (with no CLS) using tracing instance? OR as a user, how do I use basic tracer in my app?

hekike:
@mayurkale22 I finished the open action items: separates CoreTracer and CoreTracerBase tests and added some context specific tests.

mtwo:
@hekike Mayur has been on vacation for the past two weeks, but he's back next Monday. Do you need someone to approve prior to that?

hekike:
It would be helpful if we could move this work forward. I'm planning to open further PRs which would require this work.

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -0,0 +1,237 @@
...
+import * as types from './types';
+
+/**
+ * This class represents a tracer.
+ */
+export class CoreTracerBase implements types.TracerBase {
+  /** Indicates if the tracer is active */
+  private activeLocal: boolean;
+  /** A configuration for starting the tracer */
+  private config!: configTypes.TracerConfig;
```
draffensperger: Rather than using a `!` would it be possible to assign some reasonable default config?

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -0,0 +1,237 @@
...
+  /** A list of end span event listeners */
+  private eventListenersLocal: types.SpanEventListener[] = [];
+  /** Bit to represent whether trace is sampled or not. */
+  private readonly IS_SAMPLED = 0x1;
+  /** A sampler used to make sample decisions */
+  sampler!: samplerTypes.Sampler;
```
draffensperger: Similar comment here on the sampler - can we assign some reasonable default sampler, e.g. one that samples at the default rate?

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -0,0 +1,237 @@
...
+    if (options && options.samplingRate !== undefined &&
+        options.samplingRate !== null) {
+      return SamplerBuilder.getSampler(options.samplingRate)
+          .shouldSample(traceId);
+    }
+    let propagatedSample = null;
+    // if there is a context propagation, keep the decision
+    if (options && options.spanContext && options.spanContext.options) {
+      propagatedSample =
+          ((options.spanContext.options & this.IS_SAMPLED) !== 0);
```
draffensperger: Nit: can the outer parentheses around this expression be removed?

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    let sampleDecision = !!propagatedSample;
```
draffensperger: Could this logic be simplified to `return !!propagatedSample || this.sampler.shouldSample(traceId)`?

packages/opencensus-core/src/trace/model/tracer-base.ts
hekike: The default config leaves in the `opencensus-node` today. I could move it to `opencensus-core` and add as a static property to this class. Then use it here as defaults and consume from here in the `opencensus-node`. My only concern is that this is not a change in this PR and I would prefer to keep the scope focused. But I'm happy to do this as a separate PR.

PR #494, opened 2019-05-03
fix(exporter-zpages): zpages template scope

hekike:
The template expects the span properties to be in the "global" scope.
(postmortem idea: we should start to consider testing on template outputs)


mayurkale22:
Please fix the build (lint error).

hekike:
@mayurkale22 thanks for the review, PR refactored, comments addressed, CI passing.

mayurkale22:
@hekike Thanks for the PR.

Issue #454, opened 2019-03-29
Refactor resource util to match specs


mayurkale22:
Specs: census-instrumentation/opencensus-specs#235

PR #495, opened 2019-05-03
Exporting TracerBase as a separate @opencensus/nodejs-base package

hekike:
As `@opencensus/nodejs` exports a Tracing instance as default, which also initializes CLS I would like to propose a new `@opencensus/nodejs-base` package to avoid breaking change for existing customers. The existing `@opencensus/nodejs` become a thin layer on the new `@opencensus/nodejs-base` after this change.

The alternative solution would be to only export `Tracing` and `TracingBase` classes and let the user initialize them. It would also require to start CLS only when `Tracing` is initialized instead of the current global instance. This would be a breaking change.

This PR is related to https://github.com/census-instrumentation/opencensus-node/pull/484


packages/opencensus-nodejs-base/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "intercept-stdout": "^0.1.2",
+    "mocha": "^6.0.0",
+    "nyc": "14.0.0",
+    "shimmer": "^1.2.0",
+    "ts-node": "^8.0.0",
+    "typescript": "~3.2.0"
+  },
+  "dependencies": {
+    "@opencensus/core": "^0.0.11",
+    "@opencensus/instrumentation-all": "^0.0.11",
```
draffensperger: Do you still need this dependency if no plugins are pulled in by default?

packages/opencensus-nodejs-base/src/trace/instrumentation/plugin-loader.ts
```diff
@@ -33,7 +33,7 @@ enum HookState {
  */
 export class PluginLoader {
   /** The tracer */
-  private tracer: Tracer;
+  private tracer: Tracer|TracerBase;
```
draffensperger: Can this just be `TracerBase` since `Tracer` inherits from it? Same below.

packages/opencensus-nodejs/README.md
```diff
@@ -44,22 +44,7 @@ tracing.start();
...
-    childSpan = tracing.tracer.startChildSpan({ name: 'name of your child span' });
-    // Do some operation...
-    // Finish the child span at the end of it's operation
-    childSpan.end();
-
-    // Finish the root span at the end of the operation
-    rootSpan.end();
-});
-```
+For manual insturmentation see the `@opencensus/nodejs-base` package.
```
draffensperger: What if someone wants both automatic and manual instrumentation? I would think they would be able to do it by using the `@opencensus/nodejs` package as well? Maybe we should still have the doc above, but mention that if you *only* want manual instrumentation to see the `@opencensus/nodejs-base` package?

hekike: That's a very good point, I'll revert it and add the extra context.

packages/opencensus-nodejs-base/src/trace/tracing-base.ts
```diff
@@ -0,0 +1,132 @@
+/**
+ * Copyright 2018, OpenCensus Authors
```
mayurkale22: Nit: `s/2018/2019`

packages/opencensus-nodejs/README.md
```diff
@@ -61,6 +63,8 @@ tracing.tracer.startRootSpan(rootSpanOptions, (rootSpan) => {
 });
 ```
 
+For manual only instrumentation see the `@opencensus/nodejs-base` package.
```
mayurkale22: You may want to add link for the `@opencensus/nodejs-base` package.

packages/opencensus-nodejs/package.json
```diff
@@ -55,8 +55,7 @@
   },
   "dependencies": {
     "@opencensus/core": "^0.0.11",
-    "@opencensus/instrumentation-all": "^0.0.11",
-    "extend": "^3.0.2",
-    "require-in-the-middle": "^4.0.0"
+    "@opencensus/nodejs-base": "^0.0.11",
+    "@opencensus/instrumentation-all": "^0.0.11"
```
mayurkale22: Nit: Move `@opencensus/instrumentation-all` above `@opencensus/nodejs-base` to keep the ordering intact.

hekike:
@mayurkale22 thanks for the review. I'm trying to fix the build for days now and I couldn't figure out what's happening. It was passing on my machine, now with a clean copy, I can least reproduce. I feel it's more a mono repo thing and I'm missing something. Any chance you have a hint?

hekike:
@mayurkale22 I could figure out, both CI and your comments are fixed now.

packages/opencensus-nodejs-base/test/test-tracing.ts
```diff
@@ -0,0 +1,278 @@
...
+      const tracing = TracingBase.instance;
+      assert.ok(tracing instanceof TracingBase);
+    });
+  });
+
+  /** Should return a started tracing instance */
+  describe('start()', () => {
+    let aTracingBase: core.Tracing;
+    const tracing = new TracingBase();
+    // tslint:disable:no-any
```
mayurkale22: Could you please remove this TSLint rule flag?

packages/opencensus-nodejs/src/trace/tracing.ts
```diff
@@ -1,5 +1,5 @@
 /**
- * Copyright 2018, OpenCensus Authors
+ * Copyright 2019, OpenCensus Authors
```
mayurkale22: Usually, we don't update the copyright header when we touch the file.

hekike:
@mayurkale22 thanks, fixed and separate PR opened for `lerna fix` https://github.com/census-instrumentation/opencensus-node/pull/508

mayurkale22:
Just wondering: What do you think about renaming this package to either `@opencensus/basictracer-nodejs` or `@opencensus/basictracer` or `@opencensus/basetracer-nodejs` or something similar?  For me, existing `@opencensus/nodejs` name does not make good sense, but it is hard to change that at this moment.

hekike:
I also don't like the current name, but wouldn't it be worst to have two different naming concepts in the same project? I would keep the current naming convention in OpenCensus and fix it in OpenTelemetry.

mayurkale22:
Ok makes sense.

hekike:
For OpenTelemetry, I don't even think we need `nodejs` in the naming. npm is for JS packages. 
it could be just:
- `@opencensus/basictracer
- `@opencensus/tracer`

Or even more building block philosophy:
- `@opencensus/tracer` (basic tracer)
- `@opencensus/context-propagation` (cls)
- `@opencensus/automatic` (starting tracer with cls: @opencensus/tracer wrapped with @opencensus/context-propagation)

mayurkale22:
I will merge this later today, if no other comments from either @draffensperger or @justindsmith

hekike:
@mayurkale22 can we also cut a release when it's merged?

mayurkale22:
> @mayurkale22 can we also cut a release when it's merged?

Sure, I would like to include #503 and #510 (if possible #454). FYI (Release Milestone) : https://github.com/census-instrumentation/opencensus-node/milestone/3. I will cut a release later tomorrow. Is that ok?

hekike:
Perfect, thanks!

mayurkale22:
> @mayurkale22 can we also cut a release when it's merged?

The release (0.0.12) is completed for @opencensus/* packages. PTAL.

PR #496, opened 2019-05-03
feat(core): support object type attribute on spans

hekike:
It is expensive to add object type values to the span now because the user needs to serialize them even if the sampling decision is none. After this change, the `NoRecordSpan` would handle it in a zero cost way.

### DIFF
--- a/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
+++ b/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
@@ -149,7 +149,7 @@ export class NoRecordSpan implements types.Span {
   }
 
   /** No-op implementation of this method. */
-  addAttribute(key: string, value: string|number|boolean|string) {}
+  addAttribute(key: string, value: string|number|boolean|object) {}
 
   /** No-op implementation of this method. */
   addAnnotation(
--- a/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
+++ b/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
@@ -149,7 +149,7 @@ export class NoRecordSpan implements types.Span {
   }
 
   /** No-op implementation of this method. */
-  addAttribute(key: string, value: string|number|boolean) {}
+  addAttribute(key: string, value: string|number|boolean|string) {}
 
   /** No-op implementation of this method. */
   addAnnotation(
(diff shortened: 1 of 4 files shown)


packages/opencensus-core/src/trace/model/types.ts
```diff
@@ -406,7 +406,7 @@ export interface Span {
    * @param key Describes the value added.
    * @param value The result of an operation.
    */
-  addAttribute(key: string, value: string|number|boolean): void;
+  addAttribute(key: string, value: string|number|boolean|object): void;
```
kenashcraft: Document that `object` must be JSON.stringify-able or else it will raise an exception.

hekike: Documentation added

packages/opencensus-core/src/trace/model/span.ts
```diff
@@ -193,9 +193,10 @@ export class Span implements types.Span {
   /**
    * Adds an atribute to the span.
    * @param key Describes the value added.
-   * @param value The result of an operation.
+   * @param value The result of an operation. If the value is a typeof object
+   * it has to be JSON.stringify-able, cannot contain circular dependencies.
```
mayurkale22: Nit: could you indent "it has..." by 4 spaces to show it is a continuation of the @param JSDoc tag?

packages/opencensus-core/src/trace/model/types.ts
mayurkale22: Could you please update signature in `NoRecordSpan` class? https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts#L152

hekike:
@mayurkale22 thanks for the review, PR refactored, comments addressed, CI passed.

packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
```diff
@@ -149,7 +149,7 @@ export class NoRecordSpan implements types.Span {
   }
 
   /** No-op implementation of this method. */
-  addAttribute(key: string, value: string|number|boolean) {}
+  addAttribute(key: string, value: string|number|boolean|string) {}
```
mayurkale22: Typo: It should be the `object` instead of `string`.

hekike:
@mayurkale22 good catch, fixed.

Issue #279, opened 2019-01-11
typescript@3.2.0 update failed plugins and core packages.


mayurkale22:
Affected packages:

- opencensus-core
- opencensus-instrumentation-http
- opencensus-instrumentation-https
- opencensus-instrumentation-http2
- opencensus-instrumentation-grpc
- opencensus-instrumentation-mongodb

**Build error: on opencensus-core package**

> @opencensus/core@0.0.8 prepare opencensus-node/packages/opencensus-core
> npm run compile:release

> @opencensus/core@0.0.8 compile:release opencensus-node/packages/opencensus-core
> tsc -p tsconfig-release.json

src/internal/cls-ah.ts:80:34 - error TS2345: Argument of type 'IArguments' is not assignable to parameter of type 'any[]'.
  Type 'IArguments' is missing the following properties from type 'any[]': pop, push, concat, join, and 24 more.

80       const res = cb.apply(this, arguments) as T;
                                    ~~~~~~~~~

src/internal/cls-ah.ts:106:20 - error TS2554: Expected 1 arguments, but got 3.

106             return oldMethod.call(this, event, ns.bind(cb));
                       ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

Found 4 errors.

mayurkale22:
#252

PR #503, opened 2019-05-07
Add support for Derived Cumulative API

mayurkale22:
Updates #450 

Specs: https://github.com/census-instrumentation/opencensus-specs/pull/254/files?short_path=0aaf645#diff-0aaf6455b2b0b7a5f2ed45f6b36792ee

This is mostly same as [derived-gauge](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/metrics/gauges/derived-gauge.ts). 

Last piece, registering with **[Metric-Registry](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/metrics/metric-registry.ts)** will be in a follow-up PR.

### DIFF
--- a/packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
+++ b/packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
@@ -149,20 +149,22 @@ export class DerivedCumulative implements Meter {
       descriptor: this.metricDescriptor,
       timeseries: Array.from(
           this.registeredPoints,
-          ([_, cumulativeEntry]) => {
-            const newValue = cumulativeEntry.extractor();
-            const value = newValue > cumulativeEntry.prevValue ?
-                newValue :
-                cumulativeEntry.prevValue;
-            cumulativeEntry.prevValue = value;
+          ([_, cumulativeEntry]):
+              TimeSeries => {
+                const newValue = cumulativeEntry.extractor();
+                const value = newValue > cumulativeEntry.prevValue ?
+                    newValue :
+                    cumulativeEntry.prevValue;
+                cumulativeEntry.prevValue = value;
 
-            return {
-              labelValues:
-                  [...cumulativeEntry.labelValues, ...this.constantLabelValues],
-              points: [{value, timestamp}],
-              startTimestamp: this.startTime
-            } as TimeSeries;
-          })
+                return {
+                  labelValues: [
+                    ...cumulativeEntry.labelValues, ...this.constantLabelValues
+                  ],
+                  points: [{value, timestamp}],
+                  startTimestamp: this.startTime
+                };
+              })
     };
   }
 }
--- a/packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
+++ b/packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
@@ -44,15 +44,14 @@ export class DerivedCumulative implements Meter {
   /**
    * Constructs a new DerivedCumulative instance.
    *
-   * @param {string} name The name of the metric.
-   * @param {string} description The description of the metric.
-   * @param {string} unit The unit of the metric.
-   * @param {MetricDescriptorType} type The type of metric.
-   * @param {LabelKey[]} labelKeys The list of the label keys.
-   * @param {Map<LabelKey, LabelValue>} constantLabels The map of constant
-   *     labels for the Metric.
-   * @param {Timestamp} startTime The time when the cumulative metric start
-   *     measuring the value.
+   * @param name The name of the metric.
+   * @param description The description of the metric.
+   * @param unit The unit of the metric.
+   * @param type The type of metric.
+   * @param labelKeys The list of the label keys.
+   * @param constantLabels The map of constant labels for the Metric.
+   * @param startTime The time when the cumulative metric start measuring the
+   *     value.
    */
   constructor(
       name: string, description: string, unit: string,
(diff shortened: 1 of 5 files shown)


packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
```diff
@@ -0,0 +1,170 @@
...
+import {getTimestampWithProcessHRTime} from '../../common/time-util';
+import {validateArrayElementsNotNull, validateNotNull} from '../../common/validations';
+import {LabelKey, LabelValue, Metric, MetricDescriptor, MetricDescriptorType, TimeSeries, Timestamp} from '../export/types';
+import {Meter} from '../types';
+import {AccessorInterface} from '../types';
+import * as util from '../utils';
+
+type ValueExtractor = () => number;
+
+interface CumulativeEntry {
```
draffensperger: Could you add JSDoc comments to this interface and its members?

packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
```diff
@@ -0,0 +1,170 @@
...
+  private labelKeysLength: number;
+  private registeredPoints: Map<string, CumulativeEntry> = new Map();
+  private extractor?: ValueExtractor;
+  private readonly constantLabelValues: LabelValue[];
+  private startTime: Timestamp;
+
+  /**
+   * Constructs a new DerivedCumulative instance.
+   *
+   * @param {string} name The name of the metric.
```
draffensperger: What would you think about removing all the JSDoc types since those are included in the TypeScript types anyway (and so feel redundant here)?

packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
```diff
@@ -0,0 +1,170 @@
...
+        {name, description, unit, type, labelKeys: keysAndConstantKeys};
+    this.startTime = startTime;
+  }
+
+  /**
+   * Creates a TimeSeries. The value of a single point in the TimeSeries is
+   * observed from a obj. The ValueExtractor is invoked whenever
+   * metrics are collected, meaning the reported value is up-to-date.
+   *
+   * @param {LabelValue[]} labelValues The list of the label values.
```
draffensperger: Similar comment here on not including the type in the JsDoc

packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
```diff
@@ -0,0 +1,170 @@
...
+    this.constantLabelValues = [...constantLabels.values()];
+
+    this.metricDescriptor =
+        {name, description, unit, type, labelKeys: keysAndConstantKeys};
+    this.startTime = startTime;
+  }
+
+  /**
+   * Creates a TimeSeries. The value of a single point in the TimeSeries is
+   * observed from a obj. The ValueExtractor is invoked whenever
```
draffensperger: Change to `from an object or function`?

packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
```diff
@@ -0,0 +1,170 @@
...
+                newValue :
+                cumulativeEntry.prevValue;
+            cumulativeEntry.prevValue = value;
+
+            return {
+              labelValues:
+                  [...cumulativeEntry.labelValues, ...this.constantLabelValues],
+              points: [{value, timestamp}],
+              startTimestamp: this.startTime
+            } as TimeSeries;
```
draffensperger: What would you think about moving the type from here to the arrow function signature?

So it would become ` ([_, cumulativeEntry]): TimeSeries => {`. I believe the `as TimeSeries` means that the compiler wouldn't report an error if the fields in the structure don't match the type correctly, whereas specifying it as a return type would.

packages/opencensus-core/src/metrics/utils.ts
```diff
@@ -40,3 +41,36 @@ export function hashLabelValues(labelValues: LabelValue[]): string {
 export function initializeDefaultLabels(count: number): LabelValue[] {
   return new Array(count).fill(UNSET_LABEL_VALUE);
 }
+
+// Checks if the specified collection is a LengthAttributeInterface.
+// tslint:disable-next-line:no-any
+export function isLengthAttributeInterface(obj: any):
```
draffensperger: Can you use `unknown` here instead of `any`? (Same for other functions below)

packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
mayurkale22: Done, have to fix same in other places. I will open new PR to make things consistent everywhere.

packages/opencensus-core/src/metrics/cumulative/derived-cumulative.ts
mayurkale22: > I believe the as TimeSeries means that the compiler wouldn't report an error if the fields in the structure don't match the type correctly, whereas specifying it as a return type would

Hmm, you are correct. Good catch! Thanks.

packages/opencensus-core/src/metrics/utils.ts
mayurkale22: Good Point, `unknown` type is available since TypeScript 3.0, we are still using [2.9.0 version](https://github.com/census-instrumentation/opencensus-node/issues/279) for core package. For now, I have added TODO comment: 229a9b7afe078f27c6896607aa64f09cfd0acc09. PTAL once you have time.

draffensperger: The TODO comment looks great, but I'd suggest just putting it right above this function and then removing the commented out code below it. LGTM to me overall though so approving again!

mayurkale22:
Thanks @draffensperger for the reviews.

PR #505, opened 2019-05-07
Plug-in Cumulative APIs into the Metric-Registry

mayurkale22:
Updates #450

Registering Derived Cumulative APIs with **[Metric-Registry](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/metrics/metric-registry.ts)** will be in a follow-up PR. Dependency on #503

### DIFF
--- a/packages/opencensus-core/src/metrics/metric-registry.ts
+++ b/packages/opencensus-core/src/metrics/metric-registry.ts
@@ -47,9 +47,9 @@ export class MetricRegistry {
    * convenient form when you want to manually increase and decrease values as
    * per your service requirements.
    *
-   * @param {string} name The name of the metric.
-   * @param {MetricOptions} options The options for the metric.
-   * @returns {Gauge} A Int64 Gauge metric.
+   * @param name The name of the metric.
+   * @param options The options for the metric.
+   * @returns A Int64 Gauge metric.
    */
   addInt64Gauge(name: string, options?: MetricOptions): Gauge {
     const description =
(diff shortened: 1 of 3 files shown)


packages/opencensus-core/src/metrics/metric-registry.ts
```diff
@@ -165,6 +166,68 @@ export class MetricRegistry {
     return derivedDoubleGauge;
   }
 
+  /**
+   * Builds a new Int64 cumulative to be added to the registry. This API is
+   * useful when you want to manually increase and reset values as per service
+   * requirements.
+   *
+   * @param {string} name The name of the metric.
```
draffensperger: What would you think about removing these types in the JSDoc that are the same as the TS types?

packages/opencensus-core/src/metrics/metric-registry.ts
```diff
@@ -165,6 +166,68 @@ export class MetricRegistry {
...
+    this.registerMetric(name, int64Cumulative);
+    return int64Cumulative;
+  }
+
+  /**
+   * Builds a new double cumulative to be added to the registry. This API is
+   * useful when you want to manually increase and reset values as per service
+   * requirements.
+   *
+   * @param {string} name The name of the metric.
```
draffensperger: Same here on types in JSDoc

packages/opencensus-core/src/metrics/metric-registry.ts
mayurkale22: sure, done.

PR #509, opened 2019-05-08
Fix README badge URLs

irnnr:
URLs for README badges related to required Node.js version and npm package site were pointing to the Prometheus exporter. This changes the URLs to point to the correct package.


irnnr:
update: Signed the CLA.

PR #513, opened 2019-05-09
Plug-in Derived Cumulative into the Metric-Registry

### LINKED ISSUE #450: Metrics: Add Cumulative APIs

mayurkale22:
Fixes #450

Completed earlier PRs.
- [x] Add support for Cumulative API
- [x] Plug-in Cumulative APIs into the Metric-Registry
- [x] Add support for Derived Cumulative API
- [ ] Plug-in Derived Cumulative APIs into the Metric-Registry (current)

### DIFF
--- a/packages/opencensus-core/src/metrics/metric-registry.ts
+++ b/packages/opencensus-core/src/metrics/metric-registry.ts
@@ -14,11 +14,13 @@
  * limitations under the License.
  */
 
+import {getTimestampWithProcessHRTime} from '../common/time-util';
 import {validateArrayElementsNotNull, validateDuplicateKeys, validateMapElementNotNull, validateNotNull} from '../common/validations';
 import {MeasureUnit} from '../stats/types';
 import {Cumulative} from './cumulative/cumulative';
+import {DerivedCumulative} from './cumulative/derived-cumulative';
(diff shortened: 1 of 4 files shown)


packages/opencensus-core/src/metrics/metric-registry.ts
```diff
@@ -228,6 +218,66 @@ export class MetricRegistry {
...
+  /**
+   * Builds a new derived Int64 Cumulative to be added to the registry.
+   *
+   * @param name The name of the metric.
+   * @param options The options for the metric.
+   * @returns A Int64 DerivedCumulative metric.
+   */
+  addDerivedInt64Cumulative(name: string, options?: MetricOptions):
+      DerivedCumulative {
+    const description =
```
draffensperger: Seems like some of this code is duplicated with other metric adding functions. That isn't necessarily bad but makes me wonder if there would be a more templatized approach to the different metric types. (On the other hand, [Duplication is better than the wrong abstraction](https://www.sandimetz.com/blog/2016/1/20/the-wrong-abstraction) so I think this is fine)

mayurkale22: Ok, thanks.

PR #514, opened 2019-05-09
feat(core): start tracer with span attributes

hekike:
Default attributes for Tracer that will be added to every Span created by Tracer.
Useful for system level attributes like cluster information.

### DIFF
--- a/packages/opencensus-core/src/trace/config/types.ts
+++ b/packages/opencensus-core/src/trace/config/types.ts
@@ -33,8 +33,8 @@ export interface BufferConfig {
 
 /** Defines tracer configuration parameters */
 export interface TracerConfig {
-  /** A set of attributes each in the format [KEY]:[VALUE] */
-  attributes?: Attributes;
+  /** A set of default attributes each in the format [KEY]:[VALUE] */
+  defaultAttributes?: Attributes;
   /** Determines the sampling rate. Ranges from 0.0 to 1.0 */
   samplingRate?: number;
   /** A logger object  */
--- a/packages/opencensus-core/src/trace/config/types.ts
+++ b/packages/opencensus-core/src/trace/config/types.ts
@@ -18,6 +18,7 @@ import {Logger} from '../../common/types';
 import {Exporter} from '../../exporters/types';
 import {Stats} from '../../stats/types';
 import {PluginNames} from '../instrumentation/types';
+import {Attributes} from '../model/types';
 import {Propagation} from '../propagation/types';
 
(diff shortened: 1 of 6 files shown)


packages/opencensus-core/src/trace/config/types.ts
```diff
@@ -32,6 +33,8 @@ export interface BufferConfig {
 
 /** Defines tracer configuration parameters */
 export interface TracerConfig {
+  /** A set of attributes each in the format [KEY]:[VALUE] */
+  attributes?: Attributes;
```
hekike: maybe calling it `spanAttributes`? `defaultSpanAttributes`?

packages/opencensus-core/src/trace/model/tracer-base.ts
```diff
@@ -124,6 +124,12 @@ export class CoreTracerBase implements types.TracerBase {
       if (sampleDecision) {
         const rootSpan =
             new RootSpan(this, name, kind, traceId, parentSpanId, traceState);
+        // Add tracer attributes
+        if (this.config.attributes) {
+          for (const key of Object.keys(this.config.attributes)) {
```
mayurkale22: I would prefer to use `forEach` statement for better readability. WDYT?

packages/opencensus-core/src/trace/config/types.ts
mayurkale22: Another option is either `constantAttributes` or `defaultAttributes`.

hekike:
@mayurkale22 thanks for the review, I renamed `attributes` to `defaultAttributes` and refactored the code based on your comments.

packages/opencensus-core/src/trace/config/types.ts
mayurkale22: Thanks for the change, would you mind to update CHANGELOG.md with this new feature?

hekike:
Added to changelog.
*(For OpenTelemetry it would be useful to have automatic changelog generation via https://www.conventionalcommits.org)*

mayurkale22:
> Added to changelog.
> _(For OpenTelemetry it would be useful to have automatic changelog generation via https://www.conventionalcommits.org)_

Not familiar with this tool, but happy to explore more. Thanks for sharing the link.

hekike:
The links is just the spec for the commit message format.
The tools around it are:
https://github.com/conventional-changelog/conventional-changelog

For non-monorepo:
https://github.com/Netflix/unleash

etc.

PR #515, opened 2019-05-10
Instrumentation/HTTP: Handle incoming requests with long request url …

### LINKED ISSUE #486: http instrumentation cannot handle incoming requests with long request url path (>255)

0.0.11

10

### What did you do?
If possible, provide a recipe for reproducing the error.

applied http instrumentation

### What did you see instead?

Error: Invalid TagValue: /v1/auth/oauth/google/callback?state=http://localhost:3001/signin&code=4/##############################################################&scope=email%20profile%20https://www.googleapis.com/auth/userinfo.profile%20https://www.googleapis.com/auth/userinfo.email
    at TagMap.set (node_modules/@opencensus/core/build/src/tags/tag-map.js:40:19)
    at ServerResponse.response.end (node_modules/@opencensus/instrumentation-http/build/src/http.js:177:34)
    at ServerResponse.end (node_modules/compression/index.js:107:21)

mayurkale22:
…path

Fixes #486

### DIFF
--- a/packages/opencensus-instrumentation-http/src/http.ts
+++ b/packages/opencensus-instrumentation-http/src/http.ts
@@ -228,7 +228,12 @@ export class HttpPlugin extends BasePlugin {
               rootSpan.addAttribute(
                   HttpPlugin.ATTRIBUTE_HTTP_ROUTE, requestUrl.path || '');
               tags.set(
-                  stats.HTTP_SERVER_ROUTE, {value: requestUrl.path || ''},
+                  stats.HTTP_SERVER_ROUTE, {
+                    value: ((requestUrl.path && requestUrl.path.length > 255) ?
+                                requestUrl.path.substring(0, 255) :
+                                requestUrl.path) ||
+                        ''
+                  },
                   UNLIMITED_PROPAGATION_MD);
             }
             if (userAgent) {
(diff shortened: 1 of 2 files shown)


vmarchaud:
I'm not sure if they are other places where we set tag value that may be above the 256 length mark.

mayurkale22:
> I'm not sure if they are other places where we set tag value that may be above the 256 length mark.

Other places are for `client/server > method` and `client/server > statusCode`, IMO they don't have this problem.

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -228,7 +228,12 @@ export class HttpPlugin extends BasePlugin {
               rootSpan.addAttribute(
                   HttpPlugin.ATTRIBUTE_HTTP_ROUTE, requestUrl.path || '');
               tags.set(
-                  stats.HTTP_SERVER_ROUTE, {value: requestUrl.path || ''},
+                  stats.HTTP_SERVER_ROUTE, {
+                    value: ((requestUrl.path && requestUrl.path.length > 255) ?
```
draffensperger: Can this be simplified to just `(requestUrl.path || '').substring(0, 255)`?

Calling substring on a shorter string just returns the string so for strings shorter than 255 that is fine.

Should 255 be a constant as well?

mayurkale22: Sure, done. I was thinking not to use `substring` unless it is really required by checking the length.

draffensperger: Makes sense - I'd say either is fine. I'm not sure whether the `substring` has a performance/memory penalty, but if it did that seems like a reason to only do it if needed.

mayurkale22: Not sure, already made the change as per your suggestion.

PR #521, opened 2019-05-14
`startChildSpan` with `SpanOptions` interface only.

mayurkale22:
Continuation of #484, this PR is to cleanup and match internal interfaces with top level APIs. i.e. Internal `startChildSpan` interface accepts same argument(`SpanOptions`) as public facing Tracer's  `startChildSpan`.

**This PR does not contain any breaking change.**

### DIFF
--- a/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
+++ b/packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
@@ -184,25 +184,12 @@ export class NoRecordSpan implements types.Span {
 
   /**
    * Starts a new no record child span in the no record root span.
-   * @param nameOrOptions Span name string or SpanOptions object.
-   * @param kind Span kind if not using SpanOptions object.
+   * @param [options] A SpanOptions object to start a child span.
    */
-  startChildSpan(
-      nameOrOptions?: string|types.SpanOptions,
-      kind?: types.SpanKind): types.Span {
+  startChildSpan(options?: types.SpanOptions): types.Span {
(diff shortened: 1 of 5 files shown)


packages/opencensus-core/src/trace/model/no-record/no-record-span.ts
```diff
@@ -184,25 +184,12 @@ export class NoRecordSpan implements types.Span {
 
   /**
    * Starts a new no record child span in the no record root span.
-   * @param nameOrOptions Span name string or SpanOptions object.
-   * @param kind Span kind if not using SpanOptions object.
+   * @param [options] A SpanOptions object to start a child span.
```
draffensperger: What's the role of the brackets [ ] around `options` in the JSDoc? Does that signify that it's optional?

mayurkale22: Yes, brackets are used to indicate optional param.

hekike:
Should we add these to the changelog update?

mayurkale22:
> Should we add these to the changelog update?

I thought about it, but as this is an internal interface change I didn't include in CHANGELOG. WDYT?

hekike:
Hmm, technically you can require `@opencensus/core` which exports types: https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/index.ts#L19 But I don't feel strong about it.

mayurkale22:
Yeah, this is a valid point. I will update the CHAGELOG. Thanks

PR #525, opened 2019-05-14
Cleanup type assertion

mayurkale22:


mayurkale22:
@draffensperger thanks for the reviews.

PR #528, opened 2019-05-15
Change CLS import to make stats types more general. (Fix #527)

### LINKED ISSUE #527: Avoid import CLS in @opencensus/core /stats/types.ts file to make it more general.
**Problem Description:**
In @opencensus/web-types we are trying to sync to your current @opencensus/core types files. However, we found you were importing `cls` in the stats types file. This is done in [c5a7d07c](https://github.com/census-instrumentation/opencensus-node/commit/c5a7d07c03ba32040784c3fd8b638e739be0038f#diff-3f8af03ea7f8295217dd1ec432ceda36)
 
You add this line ```import * as cls from '../internal/cls';``` and use `cls` only here ```withTagContext<T>(tags: TagMap, fn: cls.Func<T>): T;```.

As this is something internal to this project, we would like to have this more general, so we can keep on the track and copy the types.

**Solution:**
As the `cls` functionality is being used only once and it only declares a Default function type, what we propose is to declare this _default function type_ in the _stats/types.ts_ file to avoid this internal import. That is, add the next line and delete the import.

```
type Func<T> = (...args: any[]) => T;
```
Also, as this is a simple default function declaration, we think is fine to repeat this code there.

crdgonzalezca:
This PR fixes Issue [#527](https://github.com/census-instrumentation/opencensus-node/issues/527).

* Adds a Default function declaration in @opencensus/core stats/types.ts to avoid import of internal functionality.
* As small change, adds missing semicolon.

### DIFF
--- a/packages/opencensus-core/src/stats/types.ts
+++ b/packages/opencensus-core/src/stats/types.ts
@@ -15,10 +15,13 @@
  */
 
 import {StatsEventListener} from '../exporters/types';
-import * as cls from '../internal/cls';
 import {Metric} from '../metrics/export/types';
 import {TagMap} from '../tags/tag-map';
 import {TagKey, TagValue} from '../tags/types';
+
+/** Default type for functions */
+// tslint:disable:no-any
+type Func<T> = (...args: any[]) => T;
 
 /** Main interface for stats. */
 export interface Stats {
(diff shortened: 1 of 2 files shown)


mayurkale22:
@crdgonzalezca  Thanks for the issue and PR.

Issue #526, opened 2019-05-15
Upgrading to 0.0.12 breaks exporter-zipkin causes application to crash


drubin:
### What version of OpenCensus are you using?
0.0.12

### What version of Node are you using?
v10.15.3

### What did you do?
Running the following code breaks with 0.0.12 but works with 0.0.11. 

In order to trigger the crash, you need to  generate some spans and wait < 1 minute it should cause a crash

```js
const port = process.env.PORT || 3000;
const upstream_uri = process.env.UPSTREAM_URI || 'http://time.jsontest.com/';
const service_name = process.env.SERVICE_NAME || 'opencensus-test-1-v1';

const tracing = require('@opencensus/nodejs');
const propagation = require('@opencensus/propagation-b3');
const b3 = new propagation.B3Format();
const zipkin = require('@opencensus/exporter-zipkin');

const exporter = new zipkin.ZipkinTraceExporter({
  url: 'http://localhost:9411/api/v2/spans',
  serviceName: 'test',
  debug: true,
  logLevel: 10
})

tracing.start({
  propagation: b3,
  samplingRate: 1.0,
  bufferSize: 1,
  bufferTimeout: 1,
  exporter
})

const express = require('express')
const app = express()
const request = require('request-promise-native')

app.get('/', async(req, res) => {
  await request({url: upstream_uri})
  res.end(`done`)
})

app.listen(port, () => {
  console.log(`${service_name} listening on port ${port}!`)
})
```
After some time I get the following 

```
events.js:174
      throw er; // Unhandled 'error' event
      ^

Error: socket hang up
    at createHangUpError (_http_client.js:323:15)
    at Socket.socketOnEnd (_http_client.js:426:23)
    at Socket.emit (events.js:194:15)
    at endReadableNT (_stream_readable.js:1125:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
Emitted 'error' event at:
    at Socket.socketOnEnd (_http_client.js:426:9)
    at Socket.emit (events.js:194:15)
    at endReadableNT (_stream_readable.js:1125:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

My package.json

```
    "@opencensus/exporter-jaeger": "0.0.12",
    "@opencensus/exporter-zipkin": "0.0.12",
    "@opencensus/nodejs": "0.0.12",
    "@opencensus/propagation-b3": "0.0.12",
```

### What did you expect to see?
Sending spans to zipkin

[...]

Issue #256, opened 2018-12-22
Store child spans in CLS context


adinunzio84:
**NB:** Before opening a feature request against this repo, consider whether the feature should/could be implemented in other the OpenCensus libraries in other languages. If so, please [open an issue on opencensus-specs](https://github.com/census-instrumentation/opencensus-specs/issues/new) first.

**Is your feature request related to a problem? Please describe.**
It would be convenient to have more control over child spans instead of just having root spans for incoming requests and child spans for outgoing requests. For example, if I have a client library that makes several requests to other services for authentication/authorization, I'd like to group them together in a single span.

#105 mentions the inability to create child spans from child spans, but this is a request to specifically store this information in a CLS namespace in a way that is accessible from anywhere (via `tracer`), as well as a request to adapt the design to have arbitrary spans not specific to receiving/sending requests. 

**Describe the solution you'd like**
I'd like to have something like a `withSpan(() => { ... })` function where the span is accessible via CLS (instead of coming as an argument in the callback), perhaps with `tracer.currentSpan()`. Maybe this could just wrap the function in a try/finally block to ensure that the span is closed. It would have to work with async and synchronous functions.

If there were a way to make use of decorators, that'd be even better, so marking a function with `@Span` with some configuration would create child spans (or follows-from spans once OpenTracing support is added) that start when the function starts and end when the function ends (or throws, etc.). I don't know enough about decorators to say how this would work / if it's possible.

**Describe alternatives you've considered**

I've considered writing my own solution using CLS, but this is a feature that I think would benefit others and I believe would belong in OpenCensus-node.

I came up with a (half-working) solution that maintained which span was active by maintaining a stack of spans on the CLS namespace. When a span was created, it was added to the stack, and it was removed when the span ended. I was not sure how well this would work for spans that start in one continuation and end in another, or how event-based spans would work.

I also considered just [creating a new namespace context](https://github.com/othiym23/node-continuation-local-storage#namespacecreatecontext) when a span is created (which would essentially be like maintaining a stack, since I think that maintains a stack internally). This would be cleaner than maintaining the stack myself, but has similar potential problems to above.

**Additional context**

In my attempted solution, I ran into a problem where (what OpenCensus calls) root spans would not be modifiable, since all that was available was the SpanContext (using OpenTracing) which is immutable. I resorted to creating child spans in middleware, so the root span was never accessible, and there was always a Span available to modify with tags and logs.

PR #532, opened 2019-05-16
set CurrentRootSpan correctly

mayurkale22:
While debugging #526, I found we are not assigning `currentRootSpan` correctly, especially in case of tracer with cls. Ideally, we should set `currentRootSpan` before the `onStartSpan` event. This change is done in https://github.com/census-instrumentation/opencensus-node/commit/92edaa9d406ac74e1805a4e80b756acedc413001#diff-b46c2c01fbf239aae015e19185803e0dL154

The `onStartSpan` and `onEndSpan` events contains following check -> 
```
if (this.currentRootSpan.traceId !== root.traceId) {
      this.logger.debug(
          'currentRootSpan != root on notifyStart. Need more investigation.');
    }
```

### DIFF
--- a/packages/opencensus-core/src/trace/model/span.ts
+++ b/packages/opencensus-core/src/trace/model/span.ts
@@ -305,6 +305,9 @@ export class Span implements types.Span {
       parentSpanId: this.parentSpanId,
       traceState: this.traceState
     });
+    if (!this.parentSpanId) {
+      this.tracer.setCurrentRootSpan(this);
+    }
 
     this.tracer.onStartSpan(this);
   }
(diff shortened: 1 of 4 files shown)


mayurkale22:
@hekike This is related to #484, Could you please review?

hekike:
LGTM

Related topic: should a child span become a parent for the next child span with CLS? 
It's not really possible with the current sync `startChildSpan` interface as we need to update span in CLS context.

mayurkale22:
> Related topic: should a child span become a parent for the next child span with CLS? It's not really possible with the current sync `startChildSpan` interface as we need to update span in CLS context.

Yes, currently it's not possible with `startChildSpan` interface. We already have issue for same https://github.com/census-instrumentation/opencensus-node/issues/256. If you are feeling motivated, maybe you can attempt to write solution and open a PR. :)

hekike:
@mayurkale22 what do you think about this proposal https://github.com/census-instrumentation/opencensus-node/pull/498 it would make it possible.

PR #538, opened 2019-05-20
Exporter/Prometheus: Fix missing tags for HTTP metrics

mayurkale22:
Currently, the `tags` is a type of map with `TagKey` as Key and `TagValue` as value. We used `Map.prototype.has` function to detect if a key is already defined, if no skip it otherwise get the associated value for it. AFAIK, The maps detects existence ( Map.prototype.has ( key )) of keys based on equality (===). Primitives like strings and numbers are compared by their value, while objects (arrays, plain objects) are compared by their reference. That comparison by reference basically checks to see if the objects given refer to the same location in memory. 

To solve the problem (short-term), I would prefer to create copy of original map with string as the key type. The alternative solution would be to change `tags: Map<TagKey, TagValue>` to `tags: Map<string, TagValue>`. It would require lots of refactoring, and and an API level breaking change.

These changes have all been tested locally and been confirmed to be working.

### DIFF
--- a/packages/opencensus-exporter-prometheus/test/test-prometheus-stats.ts
+++ b/packages/opencensus-exporter-prometheus/test/test-prometheus-stats.ts
@@ -24,7 +24,7 @@ describe('Prometheus Stats Exporter', () => {
   const prometheusServerUrl = `http://localhost:${options.port}/metrics`;
   const tagKey1 = {name: 'tagKey1'};
   const tagValue1 = {value: 'tagValue1'};
-  const tagKeys = [tagKey1];
+  const tagKeys = [{name: 'tagKey1'}];
   const tagMap = new TagMap();
   tagMap.set(tagKey1, tagValue1);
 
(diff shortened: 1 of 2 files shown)


packages/opencensus-exporter-prometheus/test/test-prometheus-stats.ts
```diff
@@ -24,7 +24,7 @@ describe('Prometheus Stats Exporter', () => {
   const prometheusServerUrl = `http://localhost:${options.port}/metrics`;
   const tagKey1 = {name: 'tagKey1'};
   const tagValue1 = {value: 'tagValue1'};
-  const tagKeys = [tagKey1];
+  const tagKeys = [{name: 'tagKey1'}];
```
mayurkale22: Earlier we used the same reference of `tagKey1`, hence tests passed.

mayurkale22:
> LGTM - maybe the changed type is worth considering for OpenTelemetry though

Sure, I will make sure we consider/implement this in OpenTelemetry. Thanks

Issue #537, opened 2019-05-20
Exporter/Stackdriver: Provide other way to set credential instead of ADC


yamadayuki:
**Is your feature request related to a problem? Please describe.**

The `@opencensus/exporter-stackdriver` requires the `GOOGLE_APPLICATION_CREDENTIALS` environment variable to send the metrics to Stackdriver using ADC. I don’t want to use `GOOGLE_APPLICATION_CREDENTIALS` environment variable to set the credentials. I’d like to set the google auth client from `google-auth-library` or pass the credential object as an option. If it is achieved, I can use the credentials encoded with Base64.

**Describe the solution you'd like**

This package requires the `auth` object to use JWT. So we can use the client using `auth.fromJSON(credentials)` if the `credentials` option is provided or the create new `GoogleAuth` instance with the `credentials` option. And then I can use `@opencensus/exporter-stackdriver` as below:

```ts
import { StackdriverTraceExporter } from "@opencensus/exporter-stackdriver";

// load the environment variable with keys encoded with Base64
const jsonContent = Buffer.from(process.env.CREDENTIALS_BASE64, "base64").toString();
const credentials = JSON.parse(jsonContent);

const exporter = StackdriverTraceExporter({
  projectId: "project-id",
  credentails: credentials,
});
```

**Describe alternatives you've considered**

I consider to inject `auth` object from `google-auth-library`. But `auth.getClient()` returns the JWT client with cache so if the developer calls `auth.getClient()` before calling `StackdriverTraceExporter.authorize` internally, this client may not include the scope `https://www.googleapis.com/auth/cloud-platform` and then the request fails.

[google-auth-library-nodejs/googleauth.ts at 37bb8c7cd0a6501103274284d9cddd6816cc881e · googleapis/google-auth-library-nodejs · GitHub](https://github.com/googleapis/google-auth-library-nodejs/blob/37bb8c7cd0a6501103274284d9cddd6816cc881e/src/auth/googleauth.ts#L692-L718)

**Additional context**

I investigated how to set the credentials in some implementation. Golang and Ruby implementations are able to inject the credentials from the JSON object.

**Golang usage**

```go
import (
	"context"
	"encoding/base64"
	"os"

	"cloud.google.com/go/trace/apiv2"
	"contrib.go.opencensus.io/exporter/stackdriver"
	"golang.org/x/oauth2/google"
	"google.golang.org/api/option"
)

func main() {
[...]

PR #541, opened 2019-05-21
 Enable to receive credentials object in creating Stackdriver exporter

yamadayuki:
To fix https://github.com/census-instrumentation/opencensus-node/issues/537

This PR enables us to set the Stackdriver's credentials from the environment variables instead of ADC. c.f. https://cloud.google.com/docs/authentication/production
It's ok to use `GOOGLE_APPLICATION_CREDENTIALS` as before or use the `credentials` option passing to the `StackdriverTraceExporter / StackdriverStatsExporter` constructor.

```ts
import { StackdriverTraceExporter } from "@opencensus/exporter-stackdriver";

// load the environment variable with keys encoded with Base64
const jsonContent = Buffer.from(process.env.CREDENTIALS_BASE64, "base64").toString();
const credentials = JSON.parse(jsonContent);

const exporter = StackdriverTraceExporter({
  projectId: "project-id",
  credentails: credentials,
});
```

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/types.ts
+++ b/packages/opencensus-exporter-stackdriver/src/types.ts
@@ -15,7 +15,7 @@
  */
 
 import {Bucket, ExporterConfig} from '@opencensus/core';
-import {JWT} from 'google-auth-library';
+import {JWT, JWTInput} from 'google-auth-library';
 
 export type Span = {
   name?: string,
@@ -151,6 +151,11 @@ export interface StackdriverExporterOptions extends ExporterConfig {
    * of a stackdriver metric. Optional
    */
   prefix?: string;
+  /**
+   * Create a JWT instance using the given credentials input containing
+   * client_email and private_key properties. Optional
+   */
+  credentials?: JWTInput;
 
   /**
    * Is called whenever the exporter fails to upload metrics to stackdriver.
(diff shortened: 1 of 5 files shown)


packages/opencensus-exporter-stackdriver/src/types.ts
```diff
@@ -151,6 +151,11 @@ export interface StackdriverExporterOptions extends ExporterConfig {
    * of a stackdriver metric. Optional
    */
   prefix?: string;
+  /**
+   * Create a JWT instance using the given credentials input containing
+   * client_email and private_key properties. Optional
```
mayurkale22: Can you update this field comment with: "If this field is set, its contents will be used for authentication instead of your application default credentials."?

mayurkale22:
@yamadayuki  As we don't have system tests, did you verified these changes against the actual stackdriver instance?

packages/opencensus-exporter-stackdriver/src/types.ts
yamadayuki: OK! Thanks for your suggestion.

yamadayuki:
> @yamadayuki As we don't have system tests, did you verified these changes against the actual stackdriver instance?

Yes. I released the `opencensus-exporter-stackdriver` including this feature to npm with own scope `@yamadayuki`. And I verified these changes using `@yamadayuki/exporter-stackdriver` in own applications.

mayurkale22:
> Yes. I released the opencensus-exporter-stackdriver including this feature to npm with own scope.

Awesome! Thanks for the work again!

PR #547, opened 2019-05-23
Trace Context Propagation: return null instead of empty spancontext

### LINKED ISSUE #546: Trace Context Propagation with HTTP instrumentation produce the invalid RootSpan

I use the OpenCensus libraries which versions are 0.0.13.

- `@opencensus/core@0.0.13`
- `@opencensus/exporter-stackdriver@0.0.13`
- `@opencensus/instrumentation-http@0.0.13`
- `@opencensus/nodejs@0.0.13`
- `@opencensus/propagation-tracecontext@0.0.13`

Node v10.15.3 (on macOS Mojave 10.14.4)

```
$ node -v 
v10.15.3
```

### What did you do?

I have written a simple Express app with OpenCensus.

```ts
import { Tracer } from "@opencensus/core";
import { StackdriverTraceExporter } from "@opencensus/exporter-stackdriver";
import * as instrumentationHttp from "@opencensus/instrumentation-http";
import tracing from "@opencensus/nodejs";
import { TraceContextFormat } from "@opencensus/propagation-tracecontext";
import express from 'express';
import * as http from 'http';

const propagation = new TraceContextFormat();
const exporter = new StackdriverTraceExporter({ projectId: "project-id" });

const tracer = tracing.start({ propagation, exporter }).tracer;
instrumentationHttp.plugin.enable(http, tracer as Tracer, process.versions.node, {});

const app = express();

app.get('/ping', (_req, res) =>  {
  res.send('pong');
});

app.listen(3000, () => {
  console.log("App listening on 3000");
});
```

When I request the `curl http://localhost:3000/ping` and then the Stackdriver exporter doesn’t export any trace. The error told me the RootSpan's TraceID is invalid.

I want to use `instrumentation-http` with `propagation-tracecontext`. The TraceID should be a valid string.

### What did you see instead?

HTTP instrumentation produces the invalid Span when I use with Trace Context propagation. This “invalid” means the TraceID is to be an empty string. But the TraceID should be hexadecimal 32-bits string. So the exporter raises the error and the exporter does not export.

### Additional context

I tried to use HTTP instrumentation with B3 Propagation and Stackdriver Propagation. Both B3 and Stackdriver propagation and HTTP instrumentation combination create proper Span.
So this problem may be caused by using HTTP instrumentation and Trace Context Propagation.

mayurkale22:
Fixes #546 

This PR contains the following changes:
1. Get rid of superfluous falseness checks in Trace Context Propagation.
2. If there is no trace context in the headers, or if the parsed `TraceId` or `SpanId` is invalid, return null instead of empty `SpanContext` object.

These changes have all been tested locally and been confirmed to be working.

### DIFF
--- a/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
+++ b/packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
@@ -16,18 +16,51 @@
 
 import {HeaderGetter, HeaderSetter, Propagation, SpanContext} from '@opencensus/core';
 import * as crypto from 'crypto';
-import {isValidOption, isValidSpanId, isValidTraceId, isValidVersion} from './validators';
+import {isValidSpanId, isValidTraceId, isValidVersion} from './validators';
 
-// Header names
+/** The traceparent header key. */
 export const TRACE_PARENT = 'traceparent';
+/** The tracestate header key. */
 export const TRACE_STATE = 'tracestate';
+/** The default trace options. This defaults to unsampled. */
 export const DEFAULT_OPTIONS = 0x0;
+/** Regular expression that represents a valid traceparent header. */
+const TRACE_PARENT_REGEX = /^[\da-f]{2}-[\da-f]{32}-[\da-f]{16}-[\da-f]{2}$/;
+
+/**
+ * Parses a traceparent header value into a SpanContext object, or null if the
+ * traceparent value is invalid.
+ */
+function traceParentToSpanContext(traceParent: string): SpanContext|null {
+  const match = traceParent.match(TRACE_PARENT_REGEX);
+  if (!match) return null;
+  const parts = traceParent.split('-');
+  const version = parts[0];
+  const traceId = parts[1];
+  const spanId = parts[2];
+  // tslint:disable-next-line:ban Needed to parse hexadecimal.
+  const options = parseInt(parts[3], 16);
+
+  if (!isValidVersion(version) || !isValidTraceId(traceId) ||
+      !isValidSpanId(spanId)) {
+    return null;
+  }
+  return {traceId, spanId, options};
+}
+
+/** Converts a headers type to a string. */
+function parseHeader(str: string|string[]|undefined): string|undefined {
+  if (Array.isArray(str)) {
+    return str[0];
+  }
+  return str;
+}
 
 /**
  * Propagates span context through Trace Context format propagation.
  *
  * Based on the Trace Context specification:
- * https://w3c.github.io/distributed-tracing/report-trace-context.html
+ * https://www.w3.org/TR/trace-context/
  *
  * Known Limitations:
  * - Multiple `tracestate` headers are merged into a single `tracestate`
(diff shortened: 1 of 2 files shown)


packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -16,18 +16,51 @@
...
+
+/**
+ * Parses a traceparent header value into a SpanContext object, or null if the
+ * traceparent value is invalid.
+ */
+function traceParentToSpanContext(traceParent: string): SpanContext|null {
+  const match = traceParent.match(TRACE_PARENT_REGEX);
+  if (!match) return null;
+  const parts = traceParent.split('-');
+  const version = parts[0];
```
draffensperger: Could you use destructuring here so `const [version, traceId, spanId] = ...`?

packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -16,18 +16,51 @@
...
+  if (!isValidVersion(version) || !isValidTraceId(traceId) ||
+      !isValidSpanId(spanId)) {
+    return null;
+  }
+  return {traceId, spanId, options};
+}
+
+/** Converts a headers type to a string. */
+function parseHeader(str: string|string[]|undefined): string|undefined {
+  if (Array.isArray(str)) {
```
draffensperger: Optional: this could be a ternary: `return Array.isArray(str) ? str[0] : str;`

packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
mayurkale22: Sure, done.

packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
mayurkale22: done.

packages/opencensus-propagation-tracecontext/src/tracecontext-format.ts
```diff
@@ -16,18 +16,46 @@
...
 export const DEFAULT_OPTIONS = 0x0;
+/** Regular expression that represents a valid traceparent header. */
+const TRACE_PARENT_REGEX = /^[\da-f]{2}-[\da-f]{32}-[\da-f]{16}-[\da-f]{2}$/;
+
+/**
+ * Parses a traceparent header value into a SpanContext object, or null if the
+ * traceparent value is invalid.
+ */
+function traceParentToSpanContext(traceParent: string): SpanContext|null {
+  const match = traceParent.match(TRACE_PARENT_REGEX);
```
yamadayuki: I think `TRACE_PARENT_REGEX.test(traceParent)` is little safer than `traceParent.match(TRACE_PARENT_REGEX)` (and little faster). How about this?

mayurkale22: I agree `.test` is faster if we want to do boolean check, But `.match` does the matching a string against a regular expression and returns an array with the matches. With the `.test` I need to use split operation to get the data.

yamadayuki: I agree with you. I saw the `match` variable is used only to check the `traceParent` format is appropriate, so I suggested to use `.test`.

PR #564, opened 2019-05-29
Object Trace Exporter for programatic span access

hekike:
I'm considering to use an in-memory exporter approach for unit testing.
This PR adds an `object-exporter` that stores started, ended and published spans in memory and makes them accessible.

### DIFF
--- a/packages/opencensus-exporter-object/README.md
+++ b/packages/opencensus-exporter-object/README.md
@@ -1,0 +1,71 @@
+# OpenCensus Object Exporter
+[![Gitter chat][gitter-image]][gitter-url]
+
+OpenCensus Object Trace Exporter allows the user to collect and
+programatically access traces with [OpenCensus Node.js](https://github.com/census-instrumentation/opencensus-node). This module is useful for when you need
+to access collected spans programatically, for example for testing purposes.
+
+## Installation
+
+Install OpenCensus Object Trace Exporter with:
+
+```bash
+npm install @opencensus/nodejs
+npm install @opencensus/exporter-object
+```
+
+## Usage
+
+For javascript:
+
+```javascript
+const tracing = require('@opencensus/nodejs');
+const { ObjectTraceExporter } = require('@opencensus/exporter-object');
+
+const exporter = new ObjectTraceExporter();
+```
+
+Similarly for Typescript:
+
+```typescript
+import * as tracing from '@opencensus/nodejs';
+import { ObjectTraceExporter } from '@opencensus/exporter-object';
+const exporter = new ObjectTraceExporter();
+```
+
+Now, register the exporter and start tracing.
+
+```javascript
+tracing.start({'exporter': exporter});
+```
+
+or
+
+```javascript
+tracing.registerExporter(exporter).start();
+```
+
+## Viewing your traces
+
+```javascript
+exporter.startedSpans.forEach((span: Span)) => {}
+exporter.endedSpans.forEach((span: Span)) => {}
+exporter.publishedSpans.forEach((span: Span)) => {}
+```
+
+## Reset exporter
+
+Empties `startedSpans`, `endedSpans` and `publishedSpans` span stores.
+
+```javascript
+exporter.reset();
+```
+
+## Useful links
+
+- For more information on OpenCensus, visit: <https://opencensus.io/>
+- To checkout the OpenCensus for Node.js, visit: <https://github.com/census-instrumentation/opencensus-node>
+- For help or feedback on this project, join us on [gitter](https://gitter.im/census-instrumentation/Lobby)
+
+[gitter-image]: https://badges.gitter.im/census-instrumentation/lobby.svg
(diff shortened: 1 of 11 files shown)


packages/opencensus-exporter-object/README.md
```diff
@@ -0,0 +1,71 @@
+# OpenCensus Object Exporter
+[![Gitter chat][gitter-image]][gitter-url]
+
+OpenCensus Object Trace Exporter allows the user to collect and
+programatically access traces with [OpenCensus Node.js](https://github.com/census-instrumentation/opencensus-node). This module is useful for when you need
```
mayurkale22: typo: `s/programatically/programmatically/g`

packages/opencensus-exporter-object/package.json
```diff
@@ -0,0 +1,66 @@
+{
+	"name": "@opencensus/exporter-object",
+	"version": "0.0.13",
+	"description": "OpenCensus Object Exporter allows the user to collect and access traces with OpenCensus Node.js.",
```
mayurkale22: Do you want to publish this package to npm? If no, can we add `private: true`?

packages/opencensus-exporter-object/test/test-object.ts
```diff
@@ -0,0 +1,106 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {CoreTracer, Span, SpanKind, TracerConfig} from '@opencensus/core';
+import * as assert from 'assert';
+
```
mayurkale22: Optional: remove the extra space

packages/opencensus-exporter-object/src/object.ts
```diff
@@ -0,0 +1,68 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {Exporter, ExporterBuffer, ExporterConfig, Span} from '@opencensus/core';
+import {logger, Logger} from '@opencensus/core';
+
+/** Object Exporter manager class */
+export class ObjectTraceExporter implements Exporter {
+  buffer: ExporterBuffer;
```
mayurkale22: Can this be a private?

packages/opencensus-exporter-object/src/object.ts
```diff
@@ -0,0 +1,68 @@
...
+    this.startedSpans.push(span);
+  }
+
+  /**
+   * Is called whenever a span is ended.
+   * @param span the ended span
+   */
+  onEndSpan(span: Span) {
+    this.endedSpans.push(span);
+    this.buffer.addToBuffer(span);
```
mayurkale22: Just curious to know the usage of `buffer` here.

packages/opencensus-exporter-object/package.json
mayurkale22: Because looks like real use case is for unit testing only, right?

hekike: How could a developer consume it without npm? I'd expect it will be installed as a dev dependency.

packages/opencensus-exporter-object/src/object.ts
hekike: Maybe I'm missing something from how OpenCensus works. I thought core module calls publish with buffer content externally. Am I correct?

packages/opencensus-exporter-object/package.json
mayurkale22: Makes sense. I thought, intention was to use in this project only.

hekike: No, sorry for not explaining the context well. I meant testing instrumentation in external services, My use case is that I instrumented an application with OpenCensus and now I want to be sure it collects the right spans during the application lifecycle. I could stub on Tracer internals but I feel it would be a cleaner approach.

packages/opencensus-exporter-object/src/object.ts
mayurkale22: Yes, correct. Earlier, I didn't understand the context well, thanks for explaining https://github.com/census-instrumentation/opencensus-node/pull/564#discussion_r288740351.  Ignore this comment.

packages/opencensus-exporter-object/src/object.ts
mayurkale22: To fix the build either make this variable public or remove check on buffer queue length.

hekike: Build fixed.

mayurkale22:
@hekike https://www.npmjs.com/package/@opencensus/exporter-object is avaiable in npm after latest release (0.0.14).

hekike:
Thank you!

PR #567, opened 2019-05-30
HTTP Propagation: set `Correlation-Context` header

### LINKED ISSUE #297: Tags: Support HTTP propagation.

mayurkale22:
Continuation of https://github.com/census-instrumentation/opencensus-node/pull/445 and Fixes #297

Specs: https://github.com/census-instrumentation/opencensus-specs/blob/master/tags/TagMap.md#over-http


packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -518,6 +532,20 @@ export class HttpPlugin extends BasePlugin {
     }
   }
 
+  /**
+   * Returns a TagMap on incoming HTTP header if it exists and is well-formed,
+   * or null otherwise.
+   * @param headers The incoming HTTP header object from which TagMap should be
+   *     retrieved.
+   */
+  static getTagContext(headers: IncomingHttpHeaders): TagMap|null {
```
songy23: s/TagContext/TagMap

packages/opencensus-instrumentation-http/src/http.ts
```diff
@@ -518,6 +532,20 @@ export class HttpPlugin extends BasePlugin {
...
+    const contextValue = (headers[CORRELATION_CONTEXT.toLocaleLowerCase()] ||
```
songy23: Why `toLocaleLowerCase`?

packages/opencensus-instrumentation-http/test/test-http.ts
```diff
@@ -290,6 +303,60 @@ describe('HttpPlugin', () => {
       });
     });
 
+    it('should create a child span for GET requests with tag context', () => {
```
songy23: I'm not sure if I understand these tests - why test spans? Same for other tests.

packages/opencensus-instrumentation-http/src/http.ts
mayurkale22: AFAIK Node core automatically converts all HTTP headers to lowercase. Without much frills, I easily found [the location](https://github.com/nodejs/node/blob/master/lib/_http_outgoing.js#L464) where outgoing HTTP headers are converted to lowercase.

@draffensperger correct me if this statement is wrong.

packages/opencensus-instrumentation-http/src/http.ts
mayurkale22: My Bad, corrected!

packages/opencensus-instrumentation-http/src/http.ts
draffensperger: What would you think about just making the constant itself be in lower case so you don't need to convert it here along with a comment that Node converts headers to lower case?

I'm not super familiar with how headers in Node are handled. Could this be related? https://github.com/nodejs/node-v0.x-archive/issues/1954
How are headers for trace context handled? Do we assume those are lowercase?

mayurkale22: TraceContext headers are defined in lower case only. As per [official documentation](https://w3c.github.io/trace-context/#header-name) -> In order to increase interoperability across multiple protocols and encourage successful integration by default it is recommended to keep the header name lower case. **Platforms and libraries MUST expect header name in any casing and SHOULD send header name in lower case.**

draffensperger: OK, I guess I'm a little confused by the code then, to me it looks like you are looking for the header key in lower case rather than converting all the incoming headers to lower case to compare them?

Or am I missing something here?

songy23: > TraceContext headers are defined in lower case only.

That's for `TraceContext`. I don't think it's the case for `Correlation-Context`, per https://github.com/w3c/correlation-context/blob/master/correlation_context/HTTP_HEADER_FORMAT.md#examples-of-http-headers.

draffensperger: Phrased a different way, how do you know that `headers: IncomingHttpHeaders` here will always have its values in lower case? Could they ever be in upper-case, and if so, should you just convert them to lower case to make sure before comparing them to a lower-cased header?

mayurkale22: > how do you know that headers: IncomingHttpHeaders here will always have its values in lower case? Could they ever be in upper-case?

Not 100% sure on this, perhaps that's why we use [both the variant](https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-instrumentation-http/src/http.ts#L218) of `user-agent` key. I think it is safe bet to compare both the variant until we don;t know the complete picture.

draffensperger: Makes sense. Maybe you can just making both the incoming header and the constant be lower case and compare those?

draffensperger: I'm going to approve this, and if this comes up as a bug in the future, it should be fixed.

PR #570, opened 2019-06-01
Fixes #566: set CurrentRootSpan whenever created span is RootSpan

### LINKED ISSUE #566: Regression: Trace id not correctly propagated between calls.

0.0.13

8.112.0 (Mac)

### What did you do?
I have a node micro-service that is instrumented using Open Census. 

I have legacy code that invokes the service using:

```
curl --request POST \
  --url http://localhost:3971/datalayer/graphql \
  --header 'accept: application/json' \
  --header 'authorization: removed' \
  --header 'content-type: application/json' \
  --header 'x-b3-parentspanid: 3f55cb323a971bee' \
  --header 'x-b3-traceid: 3f55cb323a971b0c792caabcd31897cc' \
  --data '{"query":"{\n  resource_list(cursor: \"null\", types: [\"cf-application\"]) {\n    items {\n      ... on ResourceSummary {\n        crn\n        name\n        type\n        category\n        self_link {\n          uri\n        }\n        icon\n        description\n        created_at\n        updated_at\n        status\n        group\n        actions {\n          title\n          type\n        }\n        location\n        primaryIpAddress\n        primaryBackendIpAddress\n        tags\n      }\n    }\n    cursor\n  }\n}\n"}'
```

The trace id header should be present in the generated Zipkin record.

### What did you see instead?
A new trace id header is generated here: 
https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/src/trace/model/tracer-base.ts#L118

### Additional context
Do I misunderstand the rules for this? I thought that an incoming request with:
 x-b3-traceid
 x-b3-spanid
would generate a new root element with the same trace id.

mayurkale22:
After consolidation of `Span` and `RootSpan`, we set the RootSpan to CurrentRootSpan only when `parentSpanId` is null.

Which is:
```
if (!this.parentSpanId) {
  this.tracer.setCurrentRootSpan(this);
}
```
This causes an issue when `parentSpanId` is propagating on wire. In that case, although we create a new `RootSpan` it does not get set to CurrentRootSpan, therefore we create a new root span (instead of child span) for subsequent (outgoing) request.

which is:
```
// Checks if this outgoing request is part of an operation by checking
// if there is a current root span, if so, we create a child span. In
// case there is no root span, this means that the outgoing request is
// the first operation, therefore we create a root span.

if (!plugin.tracer.currentRootSpan) {
  // outgoingRequest starting a root span
  return plugin.tracer.startRootSpan(traceOptions);
} else {
  // outgoingRequest starting a child span
  const span = plugin.tracer.startChildSpan(traceOptions);
  ....
}
```
Please use https://github.com/census-instrumentation/opencensus-node/issues/566#issuecomment-497739363 example to reproduce the problem.

This PR will set CurrentRootSpan whenever created/started span is `RootSpan`.

### DIFF
--- a/packages/opencensus-core/src/trace/model/span.ts
+++ b/packages/opencensus-core/src/trace/model/span.ts
@@ -310,10 +310,8 @@ export class Span implements types.Span {
       parentSpanId: this.parentSpanId,
       traceState: this.traceState
     });
-    if (!this.parentSpanId) {
-      this.tracer.setCurrentRootSpan(this);
-    }
 
+    if (this.isRootSpan()) this.tracer.setCurrentRootSpan(this);
     this.tracer.onStartSpan(this);
   }
 
--- a/packages/opencensus-core/test/test-span.ts
+++ b/packages/opencensus-core/test/test-span.ts
@@ -113,6 +113,34 @@ describe('Span', () => {
 
       assert.ok(span.started);
     });
+
+
+    it('should start a RootSpan and set CurrentRootSpan when parentSpanId is empty',
+       () => {
+         const rootSpan = new RootSpan(tracer, name, kind, traceId, '');
+         rootSpan.start();
+         assert.strictEqual(tracer.currentRootSpan, rootSpan);
+
+         const span = new Span(tracer, rootSpan);
+         span.start();
+         assert.strictEqual(tracer.currentRootSpan, rootSpan);
+
+         assert.ok(span.started);
+       });
+
+    it('should start a RootSpan and set CurrentRootSpan when parentSpanId is not empty',
+       () => {
+         const rootSpan =
+             new RootSpan(tracer, name, kind, traceId, 'd5955a12632d46a1');
+         rootSpan.start();
+         assert.strictEqual(tracer.currentRootSpan, rootSpan);
+
+         const span = new Span(tracer, rootSpan);
+         span.start();
+         assert.strictEqual(tracer.currentRootSpan, rootSpan);
+
+         assert.ok(span.started);
+       });
   });
 
   /**


jdbennet2001:
The logic seems reasonable. I wonder if:
```
 if (this.isRootSpan()) this.tracer.setCurrentRootSpan(this);
```

should be:
```
 if (this.isRootSpan() && !this.currentRootSpan ){
   this.tracer.setCurrentRootSpan(this);
}
```

which would leave you less vulnerable to changes in the http code.

packages/opencensus-core/src/trace/model/span.ts
```diff
@@ -310,10 +310,8 @@ export class Span implements types.Span {
       parentSpanId: this.parentSpanId,
       traceState: this.traceState
     });
-    if (!this.parentSpanId) {
-      this.tracer.setCurrentRootSpan(this);
-    }
 
+    if (this.isRootSpan()) this.tracer.setCurrentRootSpan(this);
```
draffensperger: Why is `!this.parentSpanId` insufficient for determining that a span is a root span?

jdbennet2001: A span, with a parent,  can be generated in one of two ways:
1) An outgoing call that's a child of an incoming request.  (Handled correctly)
2) An incoming call that's generated with trace and span id headers. (_Not_ handled correctly).

In case #2, the incoming span id is set to the parent id and outgoing calls are assigned a completely new trace id. It seems like a stop ship defect, tbh.

jdbennet2001:
@draffensperger Any more thoughts on this? It's a blocking defect for us.

packages/opencensus-core/src/trace/model/span.ts
draffensperger: OK, I understand now, in case #2 the span is not really a "root" span from the overall trace's perspective (the true root of the trace lies in an upstream service), but rather is only a "root" from this particular Node.js process's perspective.

In my opinion at some time in the future, it might be nice to just consolidate the distinction between a "root span" and a regular span as in this case it's actually incorrect to call it a root since it's only a root relative to this process. But I can approve this to unblock the issue.

mayurkale22: > In my opinion at some time in the future, it might be nice to just consolidate the distinction between a "root span" and a regular span as in this case it's actually incorrect to call it a root since it's only a root relative to this process. But I can approve this to unblock the issue.

I agree with you and I think we should consider this in OpenTelemetry project.

mayurkale22:
> ```
>  if (this.isRootSpan() && !this.currentRootSpan ){
>        this.tracer.setCurrentRootSpan(this);
> }
> ```
> 
> which would leave you less vulnerable to changes in the http code.

Thanks for the suggestion, but I am not sure about the extra check. AFAIK Root Span should always be set to context manager. For me,  ``` !this.currentRootSpan``` check looks like a superfluous. Let me know your thoughts, I am happy to follow your suggestion.

jdbennet2001:
> > ```
> >  if (this.isRootSpan() && !this.currentRootSpan ){
> >        this.tracer.setCurrentRootSpan(this);
> > }
> > ```
> > 
> > 
> > which would leave you less vulnerable to changes in the http code.
> 
> Thanks for the suggestion, but I am not sure about the extra check. AFAIK Root Span should always be set to context manager. For me, ` !this.currentRootSpan` check looks like a superfluous. Let me know your thoughts, I am happy to follow your suggestion.

Fair enough. Please merge away. I will use our systems to thoroughly test the fix and report whatever I find.

PR #545, opened 2019-05-22
Fix duplicate spans issue and clean up

mayurkale22:
Fixes #542

PR #574, opened 2019-06-04
Fix nested child span support for Zipkin

hekike:
0.14.0 unfortunately, breaks nested child support in Zipkin.

- fix support for nested child spans
- test for published spans

Related to:
https://github.com/census-instrumentation/opencensus-node/pull/545/files#diff-427d5fa67f108e0a06e484bd47da9015R81


mayurkale22:
@hekike Let me know if you are waiting for this release?

hekike:
@mayurkale22 thanks for asking, currently I just use `0.13.0`, so it's not blocking.

PR #629, opened 2019-07-13
feat(span): make tracer available on span

hekike:
Making tracer always available on spans would make it easier to create child spans from existing span instances when you pass your span around.

Further explanation:
https://github.com/open-telemetry/opentelemetry-specification/issues/21

### DIFF
--- a/packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
+++ b/packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
@@ -30,7 +30,7 @@ export class NoRecordRootSpan extends NoRecordSpan {
    */
   private parentSpanIdLocal: string;
   /** A tracer object */
-  tracer: types.TracerBase;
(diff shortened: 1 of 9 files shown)


packages/opencensus-core/src/trace/model/no-record/no-record-root-span.ts
```diff
@@ -31,6 +29,8 @@ export class NoRecordRootSpan extends NoRecordSpan {
    * parent was likely started on another machine.
    */
   private parentSpanIdLocal: string;
+  /** A tracer object */
+  tracer: types.TracerBase;
```
draffensperger: Could this be `readonly`? (Applies to other uses below too)

hekike: I changed it

mayurkale22:
Please update the CHANGELOG with new feat.

hekike:
@mayurkale22 I added.

PR #635, opened 2019-07-31
Use Zipkin quickstart script in download instructions

shakuzen:
The group id of the Zipkin Server maven artifact has changed in recent versions, so the previous instructions would not download the latest version of the server. The quickstart script provided by Zipkin is the recommended way to download the Zipkin Server jar and will download the latest version.


mayurkale22:
@shakuzen thanks for the PR.

PR #638, opened 2019-08-14
Nominating @hekike as a code owner

mayurkale22:
Justin Smith is no longer associated with OpenCensus project due to other assignments. Replacing him with Peter Marton (@hekike).


hekike:
thanks!

PR #644, opened 2019-08-26
Fix: add batch limit for stackdriver exporter

### LINKED ISSUE #634: Support specifying batch limit
On high-velocity metrics, you get:
```UnhandledPromiseRejectionWarning: Error: Field timeSeries had an invalid value: A maximum of 200 TimeSeries can be written in a single request.```

Apparently Stackdriver backend has the limit that each request cannot batch more than 200 time series so exporter needs to do some processing. For example https://github.com/census-instrumentation/opencensus-java/blob/9b2ac2bdc9565517155a9efb1b8581d1a91329dc/exporters/stats/stackdriver/src/main/java/io/opencensus/exporter/stats/stackdriver/CreateTimeSeriesExporter.java#L76-L77.

Using 0.0.16.

### PR DESCRIPTION
**Pull request by mayurkale22**
Fixes #634 

These changes have all been tested locally against a real stackdriver exporter and been confirmed to be working.

### DIFF
--- a/packages/opencensus-exporter-stackdriver/test/test-stackdriver-monitoring-utils.ts
+++ b/packages/opencensus-exporter-stackdriver/test/test-stackdriver-monitoring-utils.ts
@@ -32,12 +32,14 @@ import {
   createTimeSeriesList,
   OPENCENSUS_TASK_VALUE_DEFAULT,
   TEST_ONLY,
+  partitionList,
 } from '../src/stackdriver-monitoring-utils';
 import {
   Distribution,
   MetricDescriptor,
   MetricKind,
   ValueType,
+  TimeSeries,
 } from '../src/types';
 import * as nocks from './nocks';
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by mayurkale22** [comment]
@draffensperger Please review when you have a time.

File: packages/opencensus-exporter-stackdriver/test/test-stackdriver-monitoring-utils.ts [hunk 1]
```diff
@@ -616,4 +618,34 @@ describe('Stackdriver Stats Exporter Utils', () => {
...
+            value: {
+              int64Value: i,
+            },
+          },
+        ],
+      });
+    }
+
+    it('should return a partition lists', () => {
+      const list = partitionList(timeSeriesList, 200);
```
draffensperger: Would it make sense to test the exporter directly rather than the helper function? That would enable check that the helper function is used correctly and would implicitly check the helper (which could then even be made non-exported).

I'm open to leaving as-is for now if you prefer though.

ofpiyush:
Can we merge this and get a release?

philicious:
I'd like to mention another, related problem with exporting Metrics to Stackdriver:

> Rate at which data can be written to a single time series: one point each 10 seconds

(used to be even only 1 per minute in December 2019 when I run into this problem)

its still ridiculously low for per-request metrics and also causes millions of API errors, eating up memory and CPU in your service if you dont implement some throttling yourself

this caused us to not even start using Stackdriver Monitoring for custom metrics but ended up using Prometheus again.

6utt3rfly:
Is there a reason this isn't released yet? I'm running into the same issue as #634

mamidon:
We're seeing this issue ourselves, is there any workaround available?

philicious:
I guess there won't be much love for opencensus anymore as it's deprecated in favor of opentelemetry

mamidon:
Bah, that would probably explain it.  Of course the GCP docs I read must've been out of date.  Thanks for the FYI.

PR #647, opened 2019-08-27
Fix circular import

StoneDot:
This PR fixes the circular import problems occurring when you load `http2` or `https` module before `http` module.

For example, the following code reproduces a problem `@opencensus-instrumentation-http` module doesn't load properly.

```javascript
const tracing = require('@opencensus/nodejs');
const zipkin = require('@opencensus/exporter-zipkin');
const assert = require('assert');

const tracer = tracing.start({samplingRate: 1.0}).tracer;

const https = require('https');
const http = require('http');

tracer.registerSpanEventListener(new zipkin.ZipkinTraceExporter({
    url: 'http://localhost:9411/api/v2/spans',
    serviceName: 'node.js-open-census-bug'
}));

http.get('http://example.com/', (res) => {
  res.setEncoding('utf8');
  let rawData = '';
  res.on('data', (chunk) => { rawData += chunk; });
  res.on('end', () => {
    console.log(rawData);
  });
});
```

The process of problem is as follows schematically:

1. Load `http2` module.
2. `require-in-the-middle` load `@opencensus-instrumentation-http2` module.
3. `@opencensus-instrumentation-http2` load `@opencensus-instrumentation-http` module.
4. `@opencensus-instrumentation-http` load `http`.
5. `require-in-the-middle` try to load `@opencensus-instrumentation-http`, but loading of `@opencensus-instrumentation-http` is already in progress. Therefore require statement return partial `@opencensus-instrumentation-http`.
6. The problem happens.


StoneDot:
@googlebot I signed it!

StoneDot:
Rebased for fixing security/snyk

mayurkale22:
@draffensperger and @hekike what do you think about this?

packages/opencensus-instrumentation-http2/src/http2.ts
```diff
@@ -23,6 +23,7 @@ import {
   SpanKind,
   TraceOptions,
 } from '@opencensus/core';
+import 'http';
```
draffensperger: Could you add a comment for why this line is needed to avoid the circular import? And similar for the other case below?

Issue #641, opened 2019-08-22
Multiple http2.connect call cause disappearing of the root span


StoneDot:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
v0.0.16

### What version of Node are you using?
v11.11.0

### What did you do?
A following code reproduces the error.

```javascript
const tracing = require('@opencensus/nodejs');
const zipkin = require('@opencensus/exporter-zipkin');
const assert = require('assert');

const tracer = tracing.start({samplingRate: 1.0}).tracer;

const http2 = require('http2');

tracer.registerSpanEventListener(new zipkin.ZipkinTraceExporter({
    url: 'http://localhost:9411/api/v2/spans',
    serviceName: 'node.js-open-census-bug'
}));

const url = 'https://example.com/index.html';
const clientDummy = http2.connect(url);
const client = http2.connect(url);

const {
  HTTP2_HEADER_STATUS
} = http2.constants;
const headers = {};
tracer.startRootSpan({}, () => {
  // The root span exists at this point
  assert.ok(tracer.currentRootSpan);

  const req = client.request(headers);

  // At this point the root span still exists
  assert.ok(tracer.currentRootSpan);

  req.on('response', headers => {
    // But at this point, the root span disappears
    assert.ok(tracer.currentRootSpan);
    
    console.log(headers[HTTP2_HEADER_STATUS]);
    req.on('data', chunk => console.log(chunk));
    req.on('end', () => process.exit());
  });
});
```

### What did you expect to see?
```
[error] "2019-08-22T07:05:16.703Z"  "could not load plugin @opencensus/instrumentation-http of module http. Error: Cannot read property 'enable' of undefined"
200
<Buffer 3c 21 64 6f 63 74 79 70 65 20 68 74 6d 6c 3e 0a 3c 68 74 6d 6c 3e 0a 3c 68 65 61 64 3e 0a 20 20 20 20 3c 74 69 74 6c 65 3e 45 78 61 6d 70 6c 65 20 44 ... 1220 more bytes>
```

Also, it contains another bug...

### What did you see instead?

```
[error] "2019-08-22T07:04:05.607Z"  "could not load plugin @opencensus/instrumentation-http of module http. Error: Cannot read property 'enable' of undefined"
assert.js:340
    throw err;
    ^

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert.ok(tracer.currentRootSpan)

[...]

PR #650, opened 2019-09-04
Fix the problem caused by multiple http2.connect calls

StoneDot:
This PR fixes https://github.com/census-instrumentation/opencensus-node/issues/641.


mayurkale22:
@StoneDot Thanks again for your contribution!

Issue #653, opened 2019-09-16
https.get seems to break the b3 propagation and stackdriver propagation libraries


loneparadox:
`https.get` seems to break propagation libraries

### What version of OpenCensus are you using?
"@opencensus/exporter-stackdriver": "0.0.17",
"@opencensus/nodejs": "0.0.17",
"@opencensus/propagation-b3": "0.0.17"
"@opencensus/propagation-stackdriver": "0.0.17"

### What version of Node are you using?
v10.16.1

### What did you do?
The following code should reproduce the error:
```js
const tracing = require('@opencensus/nodejs');
const propagation = require('@opencensus/propagation-b3');
const b3 = new propagation.B3Format();

// const propagation = require('@opencensus/propagation-tracecontext');
// const traceContext = new propagation.TraceContextFormat();
// const propagation = require('@opencensus/propagation-stackdriver');
// const sd = propagation.v1;

tracing.start({propagation: b3});

const http = require('http');

//create a server object:
http.createServer(function (req, res) {
  const https = require('https');
  https.get('https://www.google.com', (resp) => {
    let data = '';

    resp.on('data', (chunk) => {
      data += chunk;
    });

    resp.on('end', () => {
      console.log(data.substr(0,20));

      res.write('Hello World!');
      res.end();
    });

  }).on("error", (err) => {
    console.log("Error: ", err);
    res.write(err.message);
    res.end();
  });

}).listen(3000);
```

### What did you see instead?
The above code will break with Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client

### Additional context
Notice if you change the above code to:
```js
const tracing = require('@opencensus/nodejs');
const propagation = require('@opencensus/propagation-b3');
const b3 = new propagation.B3Format();

// const propagation = require('@opencensus/propagation-tracecontext');
// const traceContext = new propagation.TraceContextFormat();
// const propagation = require('@opencensus/propagation-stackdriver');
// const sd = propagation.v1;

tracing.start({propagation: b3});

const http = require('http');

//create a server object:
http.createServer(function (req, res) {
  const https = require('https');
  let new_req = https.request({hostname: 'www.google.com', port:443}, (resp) => {
    let data = '';

    resp.on('data', (chunk) => {
      data += chunk;
    });

    resp.on('end', () => {
      console.log(data.substr(0,20));

      res.write('Hello World!');
      res.end();
    });

  }).on("error", (err) => {
    console.log("Error: ", err);
[...]

PR #657, opened 2019-10-03
Fix propagation breaking for https.get

loneparadox:
shameful copy from the http plugin code, to fix the
`https.get` same as `http.get`.

https://github.com/census-instrumentation/opencensus-node/issues/653

Copied from https://github.com/census-instrumentation/opencensus-node/pull/324/files

### DIFF
--- a/packages/opencensus-instrumentation-https/src/https.ts
+++ b/packages/opencensus-instrumentation-https/src/https.ts
@@ -59,11 +59,26 @@ export class HttpsPlugin extends HttpPlugin {
       this.getPatchHttpsOutgoingRequest()
     );
     if (semver.satisfies(this.version, '>=8.0.0')) {
-      shimmer.wrap(
-        this.moduleExports,
-        'get',
-        this.getPatchHttpsOutgoingRequest()
-      );
+      shimmer.wrap(this.moduleExports, 'get', () => {
+        // Re-implement http.get. This needs to be done (instead of using
+        // getPatchOutgoingRequestFunction to patch it) because we need to
+        // set the trace context header before the returned ClientRequest is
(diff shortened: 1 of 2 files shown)


loneparadox:
@googlebot I signed it!

packages/opencensus-instrumentation-https/src/https.ts
```diff
@@ -59,11 +59,26 @@ export class HttpsPlugin extends HttpPlugin {
       this.getPatchHttpsOutgoingRequest()
     );
     if (semver.satisfies(this.version, '>=8.0.0')) {
-      shimmer.wrap(
-        this.moduleExports,
-        'get',
-        this.getPatchHttpsOutgoingRequest()
-      );
+      shimmer.wrap(this.moduleExports, 'get', () => {
```
loneparadox: @mayurkale22 I could not figure out how to change the `http` module to put this code into a method there, and then I could have just referenced it from this `https` module...

But everything references a specific version(test and code) so I did not spend more time in trying to figure out how to do this. 
Hence the reason I ended up coping the same code from the `http` module and just dumping it here.

mayurkale22:
> I'm OK with the duplication personally particularly given that we are transitioning efforts to OpenTelemetry JS. 
> @mayurkale22 WDYT?

I agree with @draffensperger. I will fine with merging like this.

mayurkale22:
I'd like to encourage you to consider looking the https://github.com/open-telemetry/opentelemetry-js repo. This would be the next major release of [OpenCensus + OpenTracing merger](https://medium.com/opentracing/a-roadmap-to-convergence-b074e5815289). We are getting ready for Alpha release, within 1-2 days.

Issue #706, opened 2019-11-06
Fails to record metrics when running from GCP Cloud Run


patwhite:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
0.0.18

### What version of Node are you using?
10.15

### What did you do?
Create a simple docker container which exports a metric using the stackdriver exporter.  Deploy to Google Cloud Run.

### What did you expect to see?
Metrics successfully logged to stackdriver

### What did you see instead?
One or more TimeSeries could not be written: The set of resource labels is incomplete. Missing labels: (instance_id).: timeSeries[0]

### Additional context
Appears to happen because it's trying to use the GCP VM defaults, but instance ID is not a cloud run construct. There's probably a way to override this, but I'm not sure how to do it, would love that as a workaround for the time being.

mayurkale22:
Published in `0.0.19`, let us know if you encounter any issues!

PR #707, opened 2019-11-13
Added check for available InstanceID to handle Cloud Run

patwhite:
Currently, if you attempt to run the Stackdriver exporter in a cloud run container, it incorrectly detects an GCP environment and attempts to grab an InstanceID, crashing on an missing time series label of Instance ID when submitting metrics. This attempts to fix that.

The error seen is:
One or more TimeSeries could not be written: The set of resource labels is incomplete. Missing labels: (instance_id).: timeSeries[0]

### DIFF
--- a/packages/opencensus-resource-util/src/resource-utils.ts
+++ b/packages/opencensus-resource-util/src/resource-utils.ts
@@ -63,12 +63,14 @@ export async function getResourceType(): Promise<ResourceType> {
   return resourceType;
 }
 
-/** Determine if the GCP metadata server is currently available, and if an instance ID is available per compute engine */
+/** Determine if the GCP metadata server is currently available,
+ * and if an instance ID is available per compute engine
+ */
 async function isRunningOnComputeEngine() {
   const metadataAvailable = await gcpMetadata.isAvailable();
-  if(metadataAvailable) {
+  if (metadataAvailable) {
     const instance = await getInstanceId();
-    if(instance !== '') {
+    if (instance !== '') {
       return true;
     }
   }
--- a/packages/opencensus-resource-util/src/resource-utils.ts
+++ b/packages/opencensus-resource-util/src/resource-utils.ts
@@ -63,9 +63,17 @@ export async function getResourceType(): Promise<ResourceType> {
   return resourceType;
 }
 
-/** Determine if the GCP metadata server is currently available. */
+/** Determine if the GCP metadata server is currently available, and if an instance ID is available per compute engine */
 async function isRunningOnComputeEngine() {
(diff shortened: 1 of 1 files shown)


patwhite:
@googlebot I signed it!

patwhite:
Tests appear to be failing, I don't think it's particularly worthwhile for me to learn the internals here, if perhaps someone from the team could take a look. Appears like instance ID is actually not supplied in some of the tests, but that's just 5 minutes of reviewing.

patwhite:
And honestly, it would probably be better to explicitly detect cloud run to include tags of some of the interesting stuff, but again, I'm not well versed enough on the internals to do that.

mayurkale22:
@patwhite Thanks for the PR! I will try to do a review today before EOD

patwhite:
Thanks buddy, also this is tied to #706 - like I said, I don't think I did a great job on this, but hopefully it's a start!

patwhite:
Oh, and in case it helps, just a bit more info - cloud run DOES have a metadata service, but it returns a different set of data in cloud run than in the VMs. So, it detects it as GCP Compute, tried to grab an instanceId, and fails reporting that.

mayurkale22:
Did you get a chance to verify this fix on the Cloud Run container? If the fix is confirmed to be working fine, I can help you to fix the tests. Let me know.

patwhite:
I wasn't able to use the unpackaged code in my project - I'm using a typescript project, and couldn't figure out how to structure the reference in such a way it would make it through our CI pipeline. Any advice on how to do that?

patwhite:
I tried npm pack but got errors running that inside the stackdriver export package directory

mayurkale22:
I am trying to run sample app on the Cloud Run to repro'd the issue. I will update you shortly.

mayurkale22:
I also tried your solution and confirmed to be working fine, only thing is metrics are reported under [global](https://cloud.google.com/monitoring/api/resources#tag_global) resource types instead of [gce_instance](https://cloud.google.com/monitoring/api/resources#tag_gce_instance). WDYT?

patwhite:
I think it's ok to send the metrics to global for now, looking through the documentation it wasn't even clear where cloud run metrics should go.

Do you want me to close this per the PR you submitted?

mayurkale22:
> Do you want me to close this per the PR you submitted?

This PR works well in case of missing `instance_id`, But I tried to solve more generic problem with either GCE, GKE, AWS etc. resource labels. I think we should close this one and merge the #711

mayurkale22:
Closed via https://github.com/census-instrumentation/opencensus-node/pull/711

PR #711, opened 2019-11-20
Fix: fallback to global resource type when expected labels are missing

### LINKED ISSUE #706: Fails to record metrics when running from GCP Cloud Run

mayurkale22:
Resolves #706

### DIFF
--- a/packages/opencensus-exporter-stackdriver/src/common-utils.ts
+++ b/packages/opencensus-exporter-stackdriver/src/common-utils.ts
@@ -51,6 +51,9 @@ export async function getDefaultResource(
       }
     }
   });
+  if (Object.keys(labels).length !== Object.keys(mappings).length) {
+    return { type: 'global', labels: { project_id: projectId } };
+  }
   return { type, labels };
 }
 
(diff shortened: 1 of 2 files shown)


mayurkale22:
@patwhite Please review, these changes have all been tested against a Cloud Run and been confirmed to be working.

patwhite:
@mayurkale22 Thanks so much for pulling this together! Seems fundamentally different than my PR in that it's not correcting the actual environment detection logic that's causing the issue, but I'm so unfamiliar with this code, this certainly may be the best option! 

BTW, the suggestion your proposed with the OC_ env variable worked for me, thanks again!

mayurkale22:
@draffensperger @hekike Please review and approve.

draffensperger:
I'm OK with this, but I would like us to better understand the Cloud Run environment and try to actually write against the correct monitored resource. Could you open up an issue to track that?

mayurkale22:
> I'm OK with this, but I would like us to better understand the Cloud Run environment and try to actually write against the correct monitored resource. Could you open up an issue to track that?

Sure I will open an issue.

PR #782, opened 2020-03-19
Fixed getMetadata

gcangussu:
The previous implementation had an issue. It wasn't considering that the user could supply an `undefined` or a `null` in the position of the metadata argument. So it would insert a new argument in the array instead of replacing the nullish value.

For example:
```js
const args = [{}, null, null, () => {}];
getMetadata(unary, args);
// args = [{}, Metadata, null, null, () => {}]
```

This would end up resulting in a call to `makeUnaryRequest` with a null on the place of the callback:
```
//         1      2       3     4    ignored
// args = [{}, Metadata, null, null, () => {}]
makeUnaryRequest(path, serialize, deserialize, {}, Metadata, null, null)
//                                             1       2      3     4
```

The fix was to consider the type of the call to correctly assert if any nullish value is occupying the place of a metadata, in this case we replace the value instead of inserting.

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -83,7 +83,7 @@ export type SendUnaryDataCallback = (
   flags?: grpcTypes.writeFlags
 ) => void;
 
-type GrpcClientFunc = typeof Function & {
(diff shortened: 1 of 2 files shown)


gcangussu:
@mayurkale22 Sorry for the flood, but are the changes on this PR going to be taken into consideration?

PR #783, opened 2020-03-19
Cover makeGenericClientConstructor with tracing

gcangussu:
Node.js `grpc` module exports a `makeGenericClientConstructor` value, which is equal to `client.makeClientConstructor` (where client is the module exported by grpc's `client.js` file). ([source](https://github.com/grpc/grpc-node/blob/f5294f7258a3222d114c7a23d993a84ba2e42d9b/packages/grpc-native-core/index.js#L281-L282)).

We are correctly patching `client.makeClientConstructor`, but since `grpc.makeGenericClientConstructor` has a reference to the original implementation, we must update it to have the reference to the just patched version of `client.makeClientConstructor`. Without this, code that create client constructors with `grpc.makeGenericClientConstructor` will not have tracing.

### DIFF
--- a/packages/opencensus-instrumentation-grpc/src/grpc.ts
+++ b/packages/opencensus-instrumentation-grpc/src/grpc.ts
@@ -141,6 +141,9 @@ export class GrpcPlugin extends BasePlugin {
         'makeClientConstructor',
         this.getPatchClient()
       );
+
+      this.moduleExports.makeGenericClientConstructor =
+        GrpcClientModule.makeClientConstructor;
     }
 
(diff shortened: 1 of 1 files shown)


gcangussu:
@mayurkale22 Is this module still being maintained? Is the changes on this PR going to be taken into consideration?

aabmass:
Hey @gcangussu apologies for not looking at this. Is this PR still relevant?

gcangussu:
I'm not using this library anymore. But for people who may depend on it, yes, definitely this would be a good fix to have. Fell free to reopen.

PR #793, opened 2020-04-11
Use cls-hooked instead of continuation-local-storage

### LINKED ISSUE #792: Dependency on deprecated continuation-local-storage

Latest (https://github.com/census-instrumentation/opencensus-node/blob/master/packages/opencensus-core/package.json has a dependency on CLS)

v12

### What did you do?
Using `pm2` (https://pm2.keymetrics.io/), which depends on `@opencensus/core` internally.

No deprecation warnings.

### What did you see instead?
The following warning:

```
DeprecationWarning: Socket.prototype._handle is deprecated
```

coming from `continuation-local-storage`'s use of `async-listener`, which polyfills a removed API with another deprecated API. This package should be migrated to use [`cls-hooked`](https://www.npmjs.com/package/cls-hooked) instead, which uses supported async hooks APIs.

### Additional context

Migration should be trivial.

vaskevich:
Closes #792.

Would appreciate a maintainer to verify that this PR works, as I'm not familiar with this package: just trying to fix a deprecation warning a few levels deep into dependencies.


packages/opencensus-core/src/internal/cls-ah.ts
```diff
@@ -1,174 +0,0 @@
-/**
```
vaskevich, line 1: `cls-hooked` already handles different versions of Node, so I don't think we need this.

packages/opencensus-core/src/internal/cls.ts
```diff
@@ -17,39 +17,24 @@
...
-export type Func<T> = CLS.Func<T>;
-
-const useAsyncHooks: boolean = semver.satisfies(process.version, '>=8'); //&&
-// !!process.env.GCLOUD_TRACE_NEW_CONTEXT;
-
-const cls: typeof CLS = useAsyncHooks
-  ? require('./cls-ah')
-  : require('continuation-local-storage');
+// eslint-disable-next-line no-any
+export type Func<T> = (...args: any[]) => T;
```
vaskevich, line 24: Not a huge fan of the `no-any` bypass here, but I just took this definition from the original `continuation-local-storage` package. Open to suggestions, though.

vaskevich:
Yikes, tests failed and looks this needs more work. Could be that this project or a dependency is interfering with queuing (https://github.com/nodejs/diagnostics/blob/master/tracing/AsyncHooks/problematic-modules.md), though I'm not at all familiar with this library and probably won't get to debugging in the near future. If anyone else with more familiarity has any ideas and a motivation to tackle this, please do.

PR #818, opened 2020-07-02
StackDriver Propagator: Make the google header configurable

nhumrich:
Other google (stackdriver) propagator libraries in open census (for example, python) allow for the header to be configurable. I already have a bunch of projects using a different header. This change allows the header with the trace-context info to be configurable if needed.


nhumrich:
@googlebot I signed it!

nhumrich:
@googlebot I signed it!

PR #886, opened 2021-02-14
chore: update broken package-lock resolves

aabmass:
Some package-lock.json files had were using an npm mirror from netflix that is no longer working. For example https://artifacts.netflix.com/api/npm/npm-netflix/@babel/code-frame/-/code-frame-7.0.0.tgz is broken and it should use registry.npmjs.org. So I ran:

```
perl -i -p -e 's/https:\/\/artifacts\.netflix\.com\/api\/npm\/npm-netflix/https:\/\/registry.npmjs.org/' `git ls-files **/package-lock.json`
```
and then was able to run `npm install` normally and everything completed alright.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,24 +1,3557 @@
 {
   "name": "@opencensus/opencensus-base",
   "version": "0.0.1",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
   "requires": true,
+  "packages": {
+    "": {
+      "name": "@opencensus/opencensus-base",
+      "version": "0.0.1",
+      "hasInstallScript": true,
+      "license": "Apache-2.0",
+      "devDependencies": {
+        "@types/node": "^10.12.12",
+        "lerna": "^2.11.0",
+        "typedoc": "^0.17.0",
+        "typescript": "^3.0.0"
+      },
+      "engines": {
+        "node": ">=8"
+      }
+    },
+    "node_modules/@types/node": {
+      "version": "10.12.12",
+      "resolved": "https://registry.npmjs.org/@types/node/-/node-10.12.12.tgz",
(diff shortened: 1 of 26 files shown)


dashpole (approved):

jsuereth (approved):
Sorry, This was suppose dto say "how is this the failure mode?  Deleted all package locks?"   Should that be updated?

punya (commented):
Thanks for looking into this. Are we pulling from any mirrors other than registry.npmjs.org elsewhere in our lockfiles? I'm curious if the addition of netflix as a mirror was intentional, was reviewed in the appropriate way, and if we have safeguards to prevent accidental or malicious introduction of mirrors in future.

punya (approved):

aabmass:
> Are we pulling from any mirrors other than registry.npmjs.org elsewhere in our lockfiles?

@punya not anymore:

```sh
$ pcregrep -h -o1 '\"resolved\": \"https:\/\/(.+?)\/' `git ls-files **/package-lock.json` | sort | uniq
registry.npmjs.org
```

> I'm curious if the addition of netflix as a mirror was intentional, was reviewed in the appropriate way, and if we have safeguards to prevent accidental or malicious introduction of mirrors in future

This change was introduced ~2 years ago in https://github.com/census-instrumentation/opencensus-node/pull/564. The author of the PR works at netflix and judging by [this npm package](https://www.npmjs.com/package/nf-iosui-cli-tools/v/1.0.43?activeTab=readme), netflix devs probably are/were using a custom registry, either with `--registry http://artifacts.netflix.com/api/npm/npm-netflix` or probably in their npmrc file. I think this was commited by mistake and the URLs have just since broken. It doesn't look like we have any safeguards to prevent this in place, I can file an issue.

Also want to point out that package-lock.json files are not distributed on npm when publishing so mirrors can't be maliciously introduced to users (could effect local builds of OC though)

@jsuerth I tried just deleting the package locks first and regenerating but this upgraded a bunch of dependencies and broke the build as well :( so I took the route of replacing the mirror URLs instead. The changes to other lockfiles is npm upgrading the to lockfile version 2 since I am using npm v7.

Issue #864, opened 2020-11-06
Nested dependencies have vulnerabilies


DanielP50:
Please answer these questions before submitting a bug report.

### What version of OpenCensus are you using?
@opencensus/exporter-stackdriver@0.0.22

### What version of Node are you using?
12.18.2

### What did you do?
Update opencensus/exporter-stackdriver to latest version

### What did you expect to see?
Nested dependency json-bigint to have version 1.0.0 or greater
Nested dependency node-forge to have version 0.10.0 or greater

### What did you see instead?
Nested dependency json-bigint have version 0.3.1
Nested dependency node-forge have version 0.9.2

### Additional context
There is vulnerabilities related to those versions
https://nvd.nist.gov/vuln/detail/CVE-2020-8237
https://nvd.nist.gov/vuln/detail/CVE-2020-7720

aabmass:
Fixed in #887

PR #887, opened 2021-02-14
chore(stackdriver): upgrade google-auth-library, drop node8

aabmass:
BREAKING CHANGE: major version upgrade google-auth-library
BREAKING CHANGE: dropping support node8 for security fix

---

Partial fix for #864

- To fix a high vulnerability security problem in google-auth-library, had to upgrade to v6 which only supports node10+.
- Clean/refactor the CircleCI config so that I can pass extra arguments to skip this package on the node8 build
- Also had to upgrade typescript to support this, and make some ts annotation changes for this
- Upgraded package-lock (doesn't effect users) with `npm audit fix`, no longer any security issues.
- `npm run fix` to re-format code with upgrade gts

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,51 +1,26 @@
----
+version: 2.1
+
 release_tags: &release_tags
   tags:
     only: /^v\d+(\.\d+){2}(-.*)?$/
 
-test_env: &test_env
-  OPENCENSUS_MONGODB_TESTS: 1
-  OPENCENSUS_REDIS_TESTS: 1
-
-# "Include" for unit tests definition.
-unit_tests: &unit_tests
-  steps:
-    - checkout
-    - run:
-        name: Install modules and dependencies.
-        command: npm install
-    - run:
-        name: Check code style and linting
-        command: npm run check
-    - run:
-        name: Ensure code compiles to JS.
-        command: npm run compile
-    - run:
-        name: Run unit tests.
-        command: npm run test
-    - run:
-        name: Submit coverage data to codecov.
-        command: npm run codecov
-        when: always
-
-mongo_service: &mongo_service
-  image: mongo
-redis_service: &redis_service
-  image: redis
-
-version: 2.0
 workflows:
   version: 2
   tests:
     jobs:
-      - node8:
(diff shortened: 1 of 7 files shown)


packages/opencensus-exporter-stackdriver/test/test-stackdriver-cloudtrace.ts
```diff
@@ -273,16 +266,13 @@ describe('Stackdriver Trace Exporter', function() {
...
-              PROJECT_ID,
-              (body: { spans: Span[] }): boolean => {
-                assert.strictEqual(body.spans.length, 2);
-                const spans = body.spans;
-                assert.strictEqual(spans[0].spanId, rootSpan.id);
-                assert.strictEqual(spans[1].spanId, span.id);
-                return true;
-              }
-            );
+            nocks.batchWrite(PROJECT_ID, (body: { spans: Span[] }): boolean => {
```
jsuereth, line 269: Is this kind of churn due to a style-formatter upgrade?

jsuereth (approved):

.circleci/config.yml
```diff
@@ -1,51 +1,26 @@
...
           filters: *release_tags
-      - node12:
+          node_version: 8
+          npm_script_extra_args: --ignore @opencensus/exporter-stackdriver
+      - unit_tests:
+          name: node<< matrix.node_version >>
           filters: *release_tags
+          matrix:
+            parameters:
+              node_version: [10, 11, 12]
```
punya, line 23: Not related to this PR, but in future we should consider not adding support for new unstable (odd-numbered) versions of NodeJS.

punya (approved):

packages/opencensus-exporter-stackdriver/test/test-stackdriver-cloudtrace.ts
aabmass: Yes, there was a security vulnerability in the `gts` formatter

.circleci/config.yml
aabmass: agreed

aabmass: Should we just leave these versions and not add any?

